PR #79, opened 2020-05-19
Performance documentation

fabianfett:


pokryfka:
@fabianfett please let me know if you'd like any X-Ray graphs to be generated.

If possible, I think it would be fair to mention performance of `Foundation` JSON encoder/decoder per https://github.com/swift-server/swift-aws-lambda-runtime/issues/153

`swift-aws-lambda-runtime` makes it easy to change it **if** ones know about it.

For reference results of a cold run of the same function with and without `Foundation`, note:
- shorter initialization (129 vs 168ms), the packages size is ~3MiB smaller without  `libFoundation.so` which (uncompressed) is 11.8 MiB size; although the heaviest one is `libicudataswift.so.65` with 28 MiB (!)
- much faster `DecodeIn`(11.9 vs 173ms)  and `EncodeOut`(0.1 vs 40.ms)
- the difference is **much** less dramatic for warm runs

Screen Shot 2020-08-14 at 11 08 02

Screen Shot 2020-08-14 at 11 08 20

tomerd:
@fabianfett could you please update this PR to be against the `main` branch so we can delete the old `master` branch

tomerd:
closing in active PRs, feel free to re-open if still relevant

PR #87, opened 2020-05-27
add an option to start the local debugging server based on an env variable

tomerd:
motivation: make using the local debugging server easier to turn off/on without the need to change code when preparing to deploy

this came up when demoing the library to people, the idea of needing to add `#if DEBUG` on their side or repeatedly remove the `withLocalServer` was a point of push back

changes:
* add code to lambda so that in debug mode only, if the LOCAL_SERVER_ENABLED env variable is set the local degging server is started
* update example code

### DIFF
--- a/Examples/LocalDebugging/MyLambda/Sources/MyLambda/main.swift
+++ b/Examples/LocalDebugging/MyLambda/Sources/MyLambda/main.swift
@@ -15,8 +15,8 @@
 import AWSLambdaRuntime
 import Shared
 
-// set LOCAL_SERVER_ENABLED env variable to "true" to start the local server simulator
-// which will allow local debugging
+// set LOCAL_LAMBDA_SERVER_ENABLED env variable to "true" to start
+// a local server simulator which will allow local debugging
 Lambda.run { (_, request: Request, callback: @escaping (Result<Response, Error>) -> Void) in
     // TODO: something useful
     callback(.success(Response(message: "Hello, \(request.name)!")))
--- a/Examples/LocalDebugging/MyLambda/Sources/MyLambda/main.swift
+++ b/Examples/LocalDebugging/MyLambda/Sources/MyLambda/main.swift
@@ -15,9 +15,9 @@
 import AWSLambdaRuntime
 import Shared
 
-try Lambda.withLocalServer {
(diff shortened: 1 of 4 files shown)


eneko:
How about this approach, embedding `#if DEBUG` inside the `withLocalServer` closure?

```swift
extension Lambda {
    /// Execute code in the context of a mock Lambda server.
    ///
    /// - parameters:
    ///     - invocationEndpoint: The endpoint  to post payloads to.
    ///     - body: Code to run within the context of the mock server. Typically this would be a Lambda.run function call.
    ///
    /// - note: This API is designed strictly for local testing and is behind a DEBUG flag
    public static func withLocalServer(invocationEndpoint: String? = nil, _ body: @escaping () -> Void) throws {
        #if DEBUG
        let server = LocalLambda.Server(invocationEndpoint: invocationEndpoint)
        try server.start().wait()
        defer { try! server.stop() } // FIXME:
        #endif
        body()
    }
}
```

#### Usage

This would start a local server in Debug, and run normally in Release mode:
```swift
try Lambda.withLocalServer {
    Lambda.run { (_, request: Request, callback) in
        callback(.success(Response(message: "Hello, \(request.name)!")))
    }
}
```

chrisaljoudi:
So, uh, I might just be missing a ton of context here (probably am)... but wouldn’t doing this with an env-variable (vs. a build-time-deterministic flag) result in larger binaries?

It’s probably only a slight increase (if at all, perhaps I’m misunderstanding whether the compiler would link the entire library either way... though it seems like it wouldn’t with WMO on).

Still, the idea of the resulting binary having _significantly_ different behavior based on the presence of an env variable does make me a little nervous at first glance.

It might very well be the right answer—just sharing some thoughts.

tomerd:
> This would start a local server in Debug, and run normally in Release mode:

@eneko this is def one of the options we considered. the challenge with this approach tho is that the code does not do what it says it does which could lead to further confusion

tomerd:
> I just noticed that the Lambda.env check is inside an if #DEBUG anyway, so Release builds would omit it anyways. 

exactly this entire functionality is stripped away when building a release version, which is what developers should be using to deploy to AWS. the risk if only replying on debug vs. release is that we are all humans so may make a mistake and ship a debug version so the env variable is extra protection

PR #97, opened 2020-05-29
S3 bucket name must be updated before deploying

eneko:
- Indicate bucket name must be updated before running `./scripts/deploy.sh` for example functions

### DIFF
--- a/Examples/LambdaFunctions/README.md
+++ b/Examples/LambdaFunctions/README.md
@@ -22,7 +22,9 @@ Steps to deploy this sample to AWS Lambda using the AWS CLI:
 
   Notes: 
   - This script assumes you have AWS CLI installed and credentials setup in `~/.aws/credentials`.
+  - The default lambda function name is `SwiftSample`. You can specify a different one updating `lambda_name` in `deploy.sh`
   - Update `s3_bucket=swift-lambda-test` in `deploy.sh` before running (AWS S3 buckets require a unique global name)
+  - Both lambda function and S3 bucket must exist before deploying for the first time.
 
 ### Deployment instructions using AWS SAM (Serverless Application Model)
 
--- a/Examples/LambdaFunctions/README.md
+++ b/Examples/LambdaFunctions/README.md
@@ -18,7 +18,9 @@ Steps to deploy this sample to AWS Lambda using the AWS CLI:
   ./scripts/deploy.sh
   ```
 
-  Note: This script assumes you have AWS CLI installed and credentials setup in `~/.aws/credentials`.
+  Notes: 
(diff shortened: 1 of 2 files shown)


tomerd:
@eneko why close?

eneko:
@tomerd thanks, will re-open

Examples/LambdaFunctions/README.md
```diff
@@ -18,7 +18,9 @@ Steps to deploy this sample to AWS Lambda using the AWS CLI:
   ./scripts/deploy.sh
   ```
 
-  Note: This script assumes you have AWS CLI installed and credentials setup in `~/.aws/credentials`.
+  Notes: 
+  - This script assumes you have AWS CLI installed and credentials setup in `~/.aws/credentials`.
+  - Update `s3_bucket=swift-lambda-test` in `deploy.sh` before running (AWS S3 buckets require a unique global name)
```
tomerd, line 23: this is helpful. we may want to point out both  the bucket and function names in the deploy script

Andrea-Scuderi: In my opinion the S3 bucket selection should be in the script, not just in the README.md.
The issue is that the S3 name must be globally unique so a predefined one works only for the owner of it.

eneko: @Andrea-Scuderi, that is a valid point, but I think outside of the scope of this PR.

I think there are two topics to cover:

1. Help new users, unfamiliar with Lambda, get started quickly with these examples. This was the goal for this PR. Maybe documentation within the script would be helpful too.

2. Make deployment script(s) more _robust_, so it's easy to deploy one or more lambdas within a package with one command. This could be a separate PR, maybe even a separate repo (would be nice to have swift cli tool included to do this).

Focusing on the first point, I'll update the PR to indicate how to update both bucket and lambda function name before deploying.

Examples/LambdaFunctions/scripts/deploy.sh
```diff
@@ -15,8 +15,14 @@
...
 set -eu
 
+# Lambda Function name (must exist in AWS Lambda)
 lambda_name=SwiftSample
+
+# S3 bucket name to upload zip file (must exist in AWS S3)
 s3_bucket=swift-lambda-test
+
+
+
```
tomerd, line 25: remove 2 extra lines

tomerd:

tomerd:

Examples/LambdaFunctions/README.md
tomerd: would be good to track any future ideas as illustrated ^^ in GitHub issues so we dont loose track of them

Issue #96, opened 2020-05-29
add user-agent to more identify the runtime


tomerd:
e.g. https://github.com/awslabs/aws-lambda-cpp/blob/fe55d84c808243baf48468369e98e09018e959a3/src/runtime.cpp#L127-L131

PR #98, opened 2020-05-30
Add a User-Agent when using the Lambda Runtime API

bmoffatt:
Addresses #96 - This will help you stand out in a crowd of "Unknown/Unknown" in our logs

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/HTTPClient.swift
+++ b/Sources/AWSLambdaRuntimeCore/HTTPClient.swift
@@ -26,6 +26,7 @@ internal final class HTTPClient {
 
     private var state = State.disconnected
     private var executing = false
+    private let headers = HTTPHeaders([("User-Agent", "Swift-Lambda/Unknown")])
 
     init(eventLoop: EventLoop, configuration: Lambda.Configuration.RuntimeEngine) {
         self.eventLoop = eventLoop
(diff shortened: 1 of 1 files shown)


Sources/AWSLambdaRuntimeCore/HTTPClient.swift
```diff
@@ -26,6 +26,7 @@ internal final class HTTPClient {
 
     private var state = State.disconnected
     private var executing = false
+    private let headers = HTTPHeaders([("User-Agent", "Swift-Lambda/Unknown")])
```
weissi, line 29: I'd do this one `static` so we can save a few allocations (and a little bit of space). Given that we're editing this line anyway, I'd also make the `user-agent` lower-case (just to be more compatible with HTTP/2 if Lambda were to ever switch to H2).

tomerd:

tomerd:
@swift-server-bot test this pease

garnaat:
I assume you are maintaining some sort of version string for this package somewhere in the code.  If so, it might be a good idea to incorporate the version string into the User Agent string.  It would make it possible to track usage of the runtime across versions.

tomerd:
good idea @garnaat https://github.com/swift-server/swift-aws-lambda-runtime/issues/108

weissi:
@tomerd I filed https://bugs.swift.org/browse/SR-12926

PR #99, opened 2020-05-30
Allow specifying port for local server

eneko:
### Enhancements
- Allow specifying TCP port to listen when running lambda in local server. This will allow for running multiple lambdas at the same time in a local set up.
- If no port is specified, default 7000 port (from RuntimeEngine configuration) is used.

#### Usage

```swift
#if DEBUG
try Lambda.withLocalServer(port: 7001) {
    Lambda.run { (_, request: Request, callback) in
        callback(.success(Response(message: "Hello, \(request.name)!")))
    }
}
#else
Lambda.run { (_, request: Request, callback) in
    callback(.success(Response(message: "Hello, \(request.name)!")))
}
#endif
```

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
@@ -32,11 +32,12 @@ extension Lambda {
     ///
     /// - parameters:
     ///     - invocationEndpoint: The endpoint  to post payloads to.
+    ///     - port: Port for local server to listen to.
     ///     - body: Code to run within the context of the mock server. Typically this would be a Lambda.run function call.
     ///
     /// - note: This API is designed stricly for local testing and is behind a DEBUG flag
-    public static func withLocalServer(invocationEndpoint: String? = nil, _ body: @escaping () -> Void) throws {
-        let server = LocalLambda.Server(invocationEndpoint: invocationEndpoint)
+    public static func withLocalServer(invocationEndpoint: String? = nil, port: Int? = nil, _ body: @escaping () -> Void) throws {
+        let server = LocalLambda.Server(invocationEndpoint: invocationEndpoint, port: port)
         try server.start().wait()
         defer { try! server.stop() } // FIXME:
         body()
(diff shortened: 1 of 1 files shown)


tomerd:
thanks @eneko as @fabianfett points out changing the server is not enough, as we need to coordinate the configuration between the lambda (client) and the mock server. we chose to expose this woth an env variable that can be set in Xcode (scheme menu) or  command line

> We could do this by locally setting the AWS_LAMBDA_RUNTIME_API env variable

correct: https://github.com/swift-server/swift-aws-lambda-runtime/blob/master/Sources/AWSLambdaRuntimeCore/LambdaConfiguration.swift#L72

eneko:
Thanks for the feedback, @fabianfett & @tomerd.

While having the environment variable is a great for a single lambda (both release and debug), it is not trivial when running multiple lambdas. The goal for this PR was to make this trivial, very easy todo.

While Xcode schemes allow for defining environment variables for each scheme, this might be harder to achieve without using Xcode.

For example, I'd like to do:
```
$ swift run Lambda1    # on one terminal (port 7000)
$ swift run Lambda2   # on another terminal (port 7001)
etc...
```

How about replacing:
```
let ipPort = env("AWS_LAMBDA_RUNTIME_API")?.split(separator: ":") ?? ["127.0.0.1", "7000"]
```

With something like:
```swift
let ipPort = env("AWS_LAMBDA_RUNTIME_API")?.split(separator: ":") ?? [DefaultIPPort.ip, DefaultIPPort.port]
```

Then, somewhere accessible:

```swift
// Somewhere accessible
struct DefaultIPPort {
  static var ip: String = "127.0.0.1"
  static var port: String = "7000"
}
```

Usage would be:

```swift
DefaultIPPort.port = "7001"
try Lambda.withLocalServer {
    Lambda.run { (_, request: Request, callback) in
        callback(.success(Response(message: "Hello, \(request.name)!")))
    }
}
```

Or like on the PR changes:

```swift
try Lambda.withLocalServer(port: 7001) { // DefaultIPPort.port set to "7001" inside Lambda.withLocalServer
    Lambda.run { (_, request: Request, callback) in
        callback(.success(Response(message: "Hello, \(request.name)!")))
    }
}
```


Thoughts?

fabianfett:
hi @eneko, you can set an env variable on the terminal pretty easy:

```bash
$ AWS_LAMBDA_RUNTIME_API="localhost:7001" swift run lambda1
$ AWS_LAMBDA_RUNTIME_API="localhost:7002" swift run lambda2
```

should do the trick!

eneko:
Yep, I'm aware of that, is just not pretty, or trivial.

If there is no interest on finding a better solution, I'll go ahead and close the PR, no worries.

Thank you!

fabianfett:
> If there is no interest on finding a better solution, I'll go ahead and close the PR, no worries.

Don't get me wrong here. I just wanted to ensure that you have a quick fix for the situation at hand. 😃

As I stated above, I think this is something that we should address. We just need to make sure that we find a good solution... And we should be able to support this going forward. As you may have seen in #83, we have some plans for the local server and we need to make sure that all stars align.

Issue #100, opened 2020-05-30
LocalServer does not expose `/error` endpoint


fabianfett:
The local server currently does not expose an error endpoint. 

That means if a local lambda throws an error the lambda runtime tries to report an error, but get's a 404. This makes the Runtime shutdown.

TBD: What error message do we want to send back to the invoker for an error within the lambda? Probably 500?

PR #104, opened 2020-06-01
handle lambda error corrctly in mock server

tomerd:
motivation: better mock server

changes:
* add handler for :requestID/error
* return .accepted to lambda
* return .internalServerError to client + error json
* small refactoring of response code in mock server to make it DRYer


tomerd:
fixed #100

fabianfett:
Perfect. This will solve a lot of headaches!

Issue #107, opened 2020-06-02
Lambda.run cannot be used without explicit types


Ro-M:
## Steps to reproduce
- setup a basic project that uses `AWSLambdaRuntime`
- in your main file, define an `Input` and `Output` type that should be used for the Lambda
- use the `Lambda.run` function with the non-Void callback

## Observed behavior
I get a compiler error: `Cannot convert value of type 'Output' to expected argument type 'Void'`.
I need to explicitly provide the types for `Input` and for the callback. I'm okay with providing the type of `Input`, but for the callback I have to explicitly write `@escaping (Result<Output, Error>) -> (Void)` to allow the Swift compiler to disambiguate between the overloaded functions. 

## Change Request
I would prefer a solution where I only need to provide the `Output` type that the callback receives  instead of the whole callback type (along with `Input` of course).

## Example that does not currently compile
```swift
struct Input {
  let foo: String
}

struct Output {
  let bar: String
}

Lambda.run { (context, input: Input, callback) in
  callback(.success(Output(bar: input.foo.reversed())))
}
```

## How this currently need to be changed
```swift
struct Input {
  let foo: String
}

struct Output {
  let bar: String
}

Lambda.run { (context, input: Input, callback: @escaping (Result<Output, Error>) -> (Void)) in
  callback(.success(Output(bar: input.foo.reversed())))
}
```

## Solution
There are many ways we can solve this (if this is regarded as an actual issue) so I will just start of the discussion with the following: We can define a generic `typealias` like so:
```swift
typealias Closure<SuccessOutput> = (Result<SuccessOutput, Error>) -> (Void)

Lambda.run { (context, input: Input, callback: @escaping Closure<Output>) in
  // ...
}
```
We would still need to type `@escaping` each time we use the `Lambda.run` function, but we don't need to explicitly write the whole closure type (but we still could if we want to).

tomerd:
cc @jckarter since most of this is compiler related

jckarter:
Overloading is a bad idea to begin with in general, so if there's a way to get the intended API effect with only one overload, I would suggest that no matter what. The compiler isn't going to do type inference across statements, so you're not going to be able to get it to infer the callback argument from the body of a multi-line closure.

fabianfett:
This will be resolved with https://github.com/apple/swift/pull/32223

stevapple:
I believe this should be closed.

[...]

PR #111, opened 2020-06-04
Readme: Making Request and Response types explicit in Lambda.run call…

fabianfett:
…back

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -59,8 +59,8 @@ Next, create a `main.swift` and implement your Lambda.
  import AWSLambdaRuntime
 
  // in this example we are receiving and responding with strings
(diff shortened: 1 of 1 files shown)


tomerd:
thanks @fabianfett this is probably the right thing to do given #107. I think we should also update the examples in the Examples directories as they additional source of copy/paste 

Also, from a formatting pov, why break the signature to a separate line?

fabianfett:
@tomerd I've added the line breaks to ensure the callback is better visible in mobile. 

Screen Shot 2020-06-07 at 13 22 01

Removed them for now though

fabianfett:

Issue #89, opened 2020-05-27
Reorder context, future-proof for a tracing context world


ktoso:
In light of https://github.com/swift-server/async-http-client/pull/227/files#r429299831 and a number of long discussions and pondering about this...

I think we'll get the best APIs and outcome for all server libs if we stick to one style, and the "context last" indeed has some gains in Swift... I can go in depth about what convinced me today after hours of chatting with Johannes and looking at my own APIs.

This proposes to reshape the API to become:

```swift
Lambda.run { (payload: String, context, callback) in 
// this perhaps makes most sense as consistent with "last, 
// but before trailing closures (regardless who calls them)"?

// OR?
Lambda.run { (payload: String, callback, context) in
```

and we'd encourage the same shape in APIs I work on myself, AsyncHTTPClient, and the Baggage / Context work that's ongoing with [GSoC/Moritz](https://github.com/swift-server/swift-aws-lambda-runtime/issues/89). (Swift gRPC already fits).

(Pretty sure people have strong feelings about this so we can talk it over?)

Specific rule wise I think we'd end up with:

- "last"
- but before trailing closures

Believe me, it took a lot of convincing for me to change my mind... As long as we end up on the same style there will be great benefits for consistency of looks of server swift code though.

weissi:
I'm just copy & pasting my arguments from the AsyncHTTPClient PR comments:

---

Swift has many APIs that read like `addObserver<O: Observer>(_ observer: O)`. If we added `context` as the first parameter, that becomes a huge mess:

- `addObserver<O: Observer>(context: Context, _ observer: O)` is just wrong
- `addObserver<O: Observer>(context: Context, observer: O)` is less bad but much longer than necessary
- `addObserver<O: Observer>(_ observer: O, context: Context)` has exactly the right feel in Swift

More importantly, you're only supposed to add the "Observer" bit when you're "compensating for weak type information". So in many APIs where it's much clearer, you're supposed to just name the method `add`.

Example: `add(_ rectangle: Rectangle)`. If we now add `context`, we end up in a very bad spot:

- `add(context: Context, _ rectangle: Rectangle)` is again just wrong because we're adding a rectangle and not a context
- `add(context: Context, rectangle: Rectangle)` is less bad but super confusing
- `add(_ rectangle: Rectangle, context: Context)` is again as good as it gets.

Basically, you cannot just skip over `context` when it's at the front but it works very well if at the back.

[...]

PR #112, opened 2020-06-04
Rename `payload` to `event`

fabianfett:
### Motivation

As @Andrea-Scuderi has correctly [pointed out here](https://github.com/swift-server/swift-aws-lambda-runtime/issues/89#issuecomment-636500412), we should follow with our naming AWS closely. Currently the inputs name on all our handler protocols is `payload`, whereas AWS uses `event` in nearly all its languages:

- [Ruby](https://docs.aws.amazon.com/lambda/latest/dg/ruby-handler.html)
- [Node](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html)
- [Python](https://docs.aws.amazon.com/lambda/latest/dg/python-handler.html)
- [Java](https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html)
- [Go](https://docs.aws.amazon.com/lambda/latest/dg/go-handler.html)

Only exception: 
- [C#](https://docs.aws.amazon.com/lambda/latest/dg/csharp-handler.html)

To ensure general documentation is easier to understand, I would vote that we get in line with AWS, as we [did here](https://github.com/swift-server/swift-aws-lambda-runtime/pull/74).

### Changes

This pr get's our naming in line with AWS' naming.

- Rename `payload` to `event`

I'm well aware that this is a breaking change. But I think we should get those breaking changes out as soon as possible. The barrier for breaking changes will only increase every day by now.

### Open ends

There are some `payload`s left to fix, especially in the Documentation. I'll get those out, if we decide this is something we want to pursue.

### DIFF
--- a/Sources/AWSLambdaRuntime/Lambda+Codable.swift
+++ b/Sources/AWSLambdaRuntime/Lambda+Codable.swift
@@ -57,8 +57,8 @@ internal struct CodableClosureWrapper<In: Decodable, Out: Encodable>: LambdaHand
         self.closure = closure
     }
 
-    func handle(context: Lambda.Context, payload: In, callback: @escaping (Result<Out, Error>) -> Void) {
-        self.closure(context, payload, callback)
+    func handle(context: Lambda.Context, event: In, callback: @escaping (Result<Out, Error>) -> Void) {
+        self.closure(context, event, callback)
     }
 }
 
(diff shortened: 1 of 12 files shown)


tomerd:
@fabianfett @Andrea-Scuderi good catch. lets fix the examples + docs and merge this before we tag 0.1.1?

fabianfett:
Work continues in #115.

PR #122, opened 2020-06-12
Added `shutdown() -> EventLoopFuture<Void>` to the `ByteBufferLambdaHandler`

### LINKED ISSUE #121: Shutdown procedure for local testing
In a non-trivial Lambda we might use an AHC, which we initialize with a factory method like this:

```swift
class MyLambdaHandler: EventLoopLambdaHandler {
    typealias In = SQS.Event
    typealias Out = Void

    let httpClient: HTTPClient

    static func create(_ eventLoop: EventLoop) -> EventLoopFuture<ByteBufferLambdaHandler> {
        let httpClient = HTTPClient(eventLoopGroupProvider: .shared(eventLoop))
        return eventLoop.makeSucceededFuture(MyLambdaHandler(httpClient: httpClient))
    }

    private init(httpClient: HTTPClient) {
        self.httpClient = httpClient
    }

    func handle(context: Lambda.Context, payload: SQS.Event) -> EventLoopFuture<Void> {
        // not interesting
    }
}

Lambda.run(MyLambdaHandler.create)
```

We currently don't have any chance to stop the `httpClient` in a coordinated way, if we stop the `LocalTestServer`, which closes the Lambda.

This is why I propose the following:

We expand the `ByteBufferLambdaHandler` protocol with a function `syncShutdown() throws` which will be invoked, if the Lambda is stopped. We supply a default empty implementation so that developers only need to deal with it if they run into a problem.

@tomerd @weissi @ktoso @adam-fowler wdyt?

fabianfett:
### Motivation

This fixes #121.

Before we also had a small bug in the code:

```swift
public func start() -> EventLoopFuture<Void> {
    assert(self.eventLoop.inEventLoop, "Start must be called on the `EventLoop` the `Lifecycle` has been initialized with.")

    logger.info("lambda lifecycle starting with \(self.configuration)")
    self.state = .initializing
    // triggered when the Lambda has finished its last run
    let finishedPromise = self.eventLoop.makePromise(of: Int.self)
    finishedPromise.futureResult.always { _ in
        self.markShutdown()
    }.cascade(to: self.shutdownPromise)
          // ^--- to cascade to the shutdown promise, the `finishedPromise` has to be fulfilled/failed
    var logger = self.logger
    logger[metadataKey: "lifecycleId"] = .string(self.configuration.lifecycle.id)
    let runner = Runner(eventLoop: self.eventLoop, configuration: self.configuration)
    return runner.initialize(logger: logger, factory: self.factory).map { handler in
        self.state = .active(runner, handler)
        self.run(promise: finishedPromise) 
        // ^-- finishedPromise can only be fulfilled, if startup was successful
    }
}
```

Look out for the `// ^--` comments in the code above. If a factory failed before, the Lambda was not shutdown, which ultimately led to a leaked `EventLoop` and `Lifecycle`. This pr addresses this issue as well and provides a test for it.

### Changes

- added `syncShutdown()` throws to the `ByteBufferLambdaHandler` protocol and provided a default implementation (empty).
- fixed above bug. If the Lambda's factory fails the Lifecycle now will fail the `shutdownFuture` as well.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -170,11 +170,13 @@ public protocol ByteBufferLambdaHandler {
     ///
     /// - Note: In case your Lambda fails while creating your LambdaHandler in the `HandlerFactory`, this method
     ///         **is not invoked**. In this case you must cleanup the created resources immediately in the `HandlerFactory`.
-    func syncShutdown() throws
+    func shutdown(context: Lambda.ShutdownContext) -> EventLoopFuture<Void>
 }
 
 public extension ByteBufferLambdaHandler {
-    func syncShutdown() throws {}
+    func shutdown(context: Lambda.ShutdownContext) -> EventLoopFuture<Void> {
+        context.eventLoop.makeSucceededFuture(Void())
+    }
 }
 
 private enum CodecError: Error {
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -164,6 +164,17 @@ public protocol ByteBufferLambdaHandler {
     /// - Returns: An `EventLoopFuture` to report the result of the Lambda back to the runtime engine.
     ///            The `EventLoopFuture` should be completed with either a response encoded as `ByteBuffer` or an `Error`
     func handle(context: Lambda.Context, event: ByteBuffer) -> EventLoopFuture<ByteBuffer?>
+
+    /// The method to clean up your resources.
+    /// Concrete Lambda handlers implement this method to shutdown their `HTTPClient`s and database connections.
+    ///
+    /// - Note: In case your Lambda fails while creating your LambdaHandler in the `HandlerFactory`, this method
+    ///         **is not invoked**. In this case you must cleanup the created resources immediately in the `HandlerFactory`.
+    func syncShutdown() throws
+}
+
+public extension ByteBufferLambdaHandler {
+    func syncShutdown() throws {}
 }
 
 private enum CodecError: Error {
--- a/Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
@@ -89,11 +89,19 @@ extension Lambda {
                 self.run(promise: finishedPromise)
                 return finishedPromise.futureResult.mapResult { (handler, $0) }
             }
-            .flatMap { (handler, result) -> EventLoopFuture<Int> in
+            .flatMap { (handler, runnerResult) -> EventLoopFuture<Int> in
+                // after the lambda finishPromise has succeeded or failed we need to
+                // shutdown the handler
                 let shutdownContext = ShutdownContext(logger: logger, eventLoop: self.eventLoop)
-                return handler.shutdown(context: shutdownContext).recover { error in
+                return handler.shutdown(context: shutdownContext).flatMapErrorThrowing { error in
+                    // if, we had an error shuting down the lambda, we want to concatenate it with
+                    // the runner result
                     logger.error("Error shutting down handler: \(error)")
-                }.flatMapResult { _ in result }
+                    throw RuntimeError.shutdownError(shutdownError: error, runnerResult: runnerResult)
+                }.flatMapResult { (_) -> Result<Int, Error> in
+                    // we had no error shutting down the lambda. let's return the runner's result
+                    runnerResult
+                }
             }.always { _ in
                 // triggered when the Lambda has finished its last run or has a startup failure.
                 self.markShutdown()
--- a/Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
@@ -29,7 +29,7 @@ extension Lambda {
 
         private var state = State.idle {
             willSet {
-                assert(self.eventLoop.inEventLoop, "State may only be changed on the `Lifecycle`'s `eventLoop`")
+                self.eventLoop.assertInEventLoop()
                 precondition(newValue.order > self.state.order, "invalid state \(newValue) after \(self.state.order)")
             }
         }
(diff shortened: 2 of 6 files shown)


Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -164,6 +164,14 @@ public protocol ByteBufferLambdaHandler {
     /// - Returns: An `EventLoopFuture` to report the result of the Lambda back to the runtime engine.
     ///            The `EventLoopFuture` should be completed with either a response encoded as `ByteBuffer` or an `Error`
     func handle(context: Lambda.Context, event: ByteBuffer) -> EventLoopFuture<ByteBuffer?>
+
+    /// The method to clean up your resources.
+    /// Concrete Lambda handlers implement this method to shutdown their `HTTPClient`s and database connections.
```
ktoso, line 169: Is it worth adding a mention about if your create/init future failed then this would not be invoked so take special case about any resources that need to be torn down there (as we discussed that case on chat)?

Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
```diff
@@ -71,22 +71,37 @@ extension Lambda {
...
                 self.run(promise: finishedPromise)
-            }
+                return finishedPromise.futureResult.always { _ in
+                    // If the lambda is terminated (e.g. LocalServer shutdown), we make sure
+                    // developers have the chance to cleanup their resources.
+                    do {
+                        try handler.syncShutdown()
+                    } catch {
+                        logger.error("Error shutting down handler: \(error)")
+                    }
```
tomerd, line 97: @fabianfett why not surface the handler shutdown error into the finishPromise?

Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
```diff
@@ -71,22 +71,37 @@ extension Lambda {
...
+                // to `run` the lambda.
+                let finishedPromise = self.eventLoop.makePromise(of: Int.self)
                 self.state = .active(runner, handler)
                 self.run(promise: finishedPromise)
-            }
+                return finishedPromise.futureResult.always { _ in
+                    // If the lambda is terminated (e.g. LocalServer shutdown), we make sure
+                    // developers have the chance to cleanup their resources.
+                    do {
+                        try handler.syncShutdown()
```
fabianfett, line 94: @weissi @tomerd We are on the EventLoop at this point. AHCs `httpClient.syncShutdown()` would be a no go at that point. Where do we want to dispatch to?

Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
fabianfett: Good idea, we might overwrite a previous error though? Do we want that? wdyt? @tomerd

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
fabianfett: @ktoso fixed below

Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
tomerd: maybe roll them up to a "ShutdownError" of sorts that contain a collection/array of the underlying errors?

Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
weissi: I don't think the `handler` would expect to be called from outside the `EventLoop` to do the shutdown. I think you need to require the caller to provide an asynchronous shutdown method.

weissi: Well, you could also document that it'll be called on another thread but that's maybe a bit weird. AHC's `syncShutdown` is thread-safe but I don't think we can expect that from everybody.

Also: If shutdown is synchronous, then often the startup is also synchronous, no?

tomerd: > I think you need to require the caller to provide an asynchronous shutdown method.

I tend to agree this fits better with the rest of APIs

fabianfett: @weissi @tomerd @ktoso I addressed that, by making `shutdown` async. Further I went a little ahead and wrapped the `eventLoop` as well as the `logger` into a `ShutdownContext`. To get in line with #126 

lmkwyt

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
ktoso: looks good

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -164,6 +164,19 @@ public protocol ByteBufferLambdaHandler {
...
+    ///
+    /// - Note: In case your Lambda fails while creating your LambdaHandler in the `HandlerFactory`, this method
+    ///         **is not invoked**. In this case you must cleanup the created resources immediately in the `HandlerFactory`.
+    func shutdown(context: Lambda.ShutdownContext) -> EventLoopFuture<Void>
```
weissi, line 173: Is this invoked once per lambda invocation?

fabianfett: no. only in debug mode for shutdown. or if the lambda occurs an recoverable error. and the lambda is therefore closed.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -95,7 +95,7 @@ private extension Lambda.Context {
 }
 
 // TODO: move to nio?
-private extension EventLoopFuture {
+extension EventLoopFuture {
```
tomerd, line 98: is this change leftover from previous iteration?

Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
```diff
@@ -71,22 +71,35 @@ extension Lambda {
...
+                // to `run` the lambda.
+                let finishedPromise = self.eventLoop.makePromise(of: Int.self)
                 self.state = .active(runner, handler)
                 self.run(promise: finishedPromise)
+                return finishedPromise.futureResult.mapResult { (handler, $0) }
             }
+            .flatMap { (handler, result) -> EventLoopFuture<Int> in
+                let shutdownContext = ShutdownContext(logger: logger, eventLoop: self.eventLoop)
+                return handler.shutdown(context: shutdownContext).recover { error in
+                    logger.error("Error shutting down handler: \(error)")
```
tomerd, line 95: not a blocker but ideally the shutdown error would bubble into the `shutdownPromise` instead of being logged

fabianfett: @tomerd I hope I fixed this with my last push.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
weissi: @fabianfett ah right. In that case, couldn't we run it synchronously _after_ `MTELG.withCallingThreadAsEventLoop` returns?

weissi: Ah no, doesn't make sense. Shutdown probably needs the EL :)

weissi: ignore my comments. This looks good.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
fabianfett: no it's needed, since I use `mapResult` in another file.

weissi: Also: Please don't use any access qualifiers (especially not `public`) in front of `extension`. It makes code reviews really difficult because a `func foo() {}` can be public if maybe 100 lines above there's a `public extension XYZ`. I think all qualifiers should be banned in front of `extension`

ktoso: I hope there's some linter for this perhaps? Maybe time to look around for one hm

PR #126, opened 2020-06-12
add initialization context

tomerd:
motivation: in more complext initialization scearios you may want access to a logger or other utilities

changes:
* introduce new InitializationContext type that can be extened in the future without breaking the API in semantic-major way
* instead of passing in EventLoop to the handler factory, pass in a context that includes a Logger and and EventLoop
* fix a bug where we dont hop back to the event loop when coming back from the handler
* adjust tests to the new signature

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -33,9 +33,13 @@ extension Lambda {
         ///         Most importantly the `EventLoop` must never be blocked.
         public let eventLoop: EventLoop
 
-        internal init(logger: Logger, eventLoop: EventLoop) {
+        /// `ByteBufferAllocator` to allocate `ByteBuffer`
+        public let allocator: ByteBufferAllocator
+
+        internal init(logger: Logger, eventLoop: EventLoop, allocator: ByteBufferAllocator) {
             self.eventLoop = eventLoop
             self.logger = logger
+            self.allocator = allocator
         }
     }
 }
@@ -87,7 +91,8 @@ extension Lambda {
                       cognitoIdentity: String? = nil,
                       clientContext: String? = nil,
                       logger: Logger,
-                      eventLoop: EventLoop) {
+                      eventLoop: EventLoop,
+                      allocator: ByteBufferAllocator) {
             self.requestID = requestID
             self.traceID = traceID
             self.invokedFunctionARN = invokedFunctionARN
@@ -96,7 +101,7 @@ extension Lambda {
             self.deadline = deadline
             // utility
             self.eventLoop = eventLoop
-            self.allocator = ByteBufferAllocator()
+            self.allocator = allocator
             // mutate logger with context
             var logger = logger
             logger[metadataKey: "awsRequestID"] = .string(requestID)
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -16,6 +16,33 @@ import Dispatch
 import Logging
 import NIO
 
+// MARK: - InitializationContext
+
+extension Lambda {
+    /// Lambda runtime initialization context.
+    /// The Lambda runtime generates and passes the `InitializationContext` to the Lambda handler as an argument.
+    public final class InitializationContext {
+        /// `Logger` to log with
+        ///
+        /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
+        public let logger: Logger
+
+        /// The `EventLoop` the Lambda is executed on. Use this to schedule work with.
+        /// This is useful when implementing the `EventLoopLambdaHandler` protocol.
+        ///
+        /// - note: The `EventLoop` is shared with the Lambda runtime engine and should be handled with extra care.
+        ///         Most importantly the `EventLoop` must never be blocked.
+        public let eventLoop: EventLoop
+
+        internal init(logger: Logger, eventLoop: EventLoop) {
+            self.eventLoop = eventLoop
+            self.logger = logger
+        }
+    }
+}
+
+// MARK: - Context
+
 extension Lambda {
     /// Lambda runtime context.
     /// The Lambda runtime generates and passes the `Context` to the Lambda handler as an argument.
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -20,7 +20,7 @@ import NIO
 
 extension Lambda {
     /// Lambda runtime initialization context.
-    /// The Lambda runtime generates and passes the `InitializationContext` to the Lambda handler as an argument.
+    /// The Lambda runtime generates and passes the `InitializationContext` to the Lambda factory as an argument.
     public final class InitializationContext {
         /// `Logger` to log with
         ///
@@ -28,7 +28,6 @@ extension Lambda {
         public let logger: Logger
 
         /// The `EventLoop` the Lambda is executed on. Use this to schedule work with.
-        /// This is useful when implementing the `EventLoopLambdaHandler` protocol.
         ///
         /// - note: The `EventLoop` is shared with the Lambda runtime engine and should be handled with extra care.
         ///         Most importantly the `EventLoop` must never be blocked.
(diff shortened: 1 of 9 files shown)


Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -57,6 +58,7 @@ extension Lambda {
                 let context = Context(logger: logger, eventLoop: self.eventLoop, invocation: invocation)
                 logger.debug("sending invocation to lambda handler \(handler)")
                 return handler.handle(context: context, event: event)
+                    .hop(to: self.eventLoop)
```
tomerd, line 61: 👀

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -16,6 +16,33 @@ import Dispatch
 import Logging
 import NIO
 
+// MARK: - InitializationContext
+
+extension Lambda {
+    /// Lambda runtime initialization context.
+    /// The Lambda runtime generates and passes the `InitializationContext` to the Lambda handler as an argument.
+    public final class InitializationContext {
```
tomerd, line 24: also considered `InitContext` but this feels more correct since the function is called `initialize`

tomerd:
cc @adam-fowler

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
fabianfett: @tomerd The only situation in which we would need to `hop` is if developers bring their own `EventLoopGroup`. Is that something that is likely?

tomerd: this can happen is if they use some library (lets say a DB client) that manages it's own ELG for whatever reason and they return a future that originated from that ELG. while this is not a pattern we encourage, we also can't stop anyone from doing it so to be safe we should hop imo

tomerd:
@fabianfett one thing I considered was also exposing an allocator in the new context. do you think it would be a good addition or wait-and-see?

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -36,13 +36,20 @@ extension Lambda {
...
-                self.runtimeClient.reportInitializationError(logger: logger, error: error).peekError { reportingError in
-                    // We're going to bail out because the init failed, so there's not a lot we can do other than log
-                    // that we couldn't report this error back to the runtime.
-                    logger.error("failed reporting initialization error to lambda runtime engine: \(reportingError)")
+            let context = InitializationContext(logger: logger, eventLoop: self.eventLoop)
+            return factory(context)
+                // hopping back to "our" EventLoop is importnant in case the factory returns
+                // a future that originiated from a different EventLoop
+                // this can happen if the factory uses a library (lets say a DB client) that manages it's own EventLoops
+                // for whatever reason and returns a future that originated from that foreign EventLoop.
```
tomerd, line 44: just added a comment, the rest of it is formatting

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -37,13 +37,19 @@ extension Lambda {
...
-            return factory(context).hop(to: self.eventLoop).peekError { error in
-                self.runtimeClient.reportInitializationError(logger: logger, error: error).peekError { reportingError in
-                    // We're going to bail out because the init failed, so there's not a lot we can do other than log
-                    // that we couldn't report this error back to the runtime.
-                    logger.error("failed reporting initialization error to lambda runtime engine: \(reportingError)")
+            return factory(context)
+                // hopping back to "our" EventLoop is importnant in case the factory returns
+                // a future that originiated from a different EventLoop
+                // this can happen if the factory uses a library (lets say a DB client) that manages it's own EventLoops
+                // for whatever reason and returns a future that originated from that foreign EventLoop.
```
fabianfett:
```suggestion
                // Hopping back to "our" `EventLoop` is important in case the `factory` returns
                // a future that originated from a different `EventLoop`.
                // This can happen if the factory uses a library (lets say a DB client) that manages 
                // its own EventLoopGroup with its own loops for whatever reason and 
                // returns a future that originated from a foreign EventLoop.
```

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -36,13 +36,20 @@ extension Lambda {
...
             // 2. report initialization error if one occured
-            return factory(self.eventLoop).hop(to: self.eventLoop).peekError { error in
-                self.runtimeClient.reportInitializationError(logger: logger, error: error).peekError { reportingError in
-                    // We're going to bail out because the init failed, so there's not a lot we can do other than log
-                    // that we couldn't report this error back to the runtime.
-                    logger.error("failed reporting initialization error to lambda runtime engine: \(reportingError)")
+            let context = InitializationContext(logger: logger, eventLoop: self.eventLoop)
+            return factory(context)
+                // Hopping back to "our" EventLoop is importnant in case the factory returns a future
+                // that originiated from a foreign EventLoop/EventLoopGroup.
```
yim-lee, line 42:
```suggestion
                // that originated from a foreign EventLoop/EventLoopGroup.
```

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -36,13 +36,20 @@ extension Lambda {
...
+                // This can happen if the factory uses a library (lets say a DB client) that manages its own threads/loops
```
yim-lee, line 43:
```suggestion
                // This can happen if the factory uses a library (let's say a database client) that manages its own threads/loops
```

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
adam-fowler: Vapor doesn't do a `hop` in its routes and quite a number of people hit that issue at some point. After that they know its an issue and don't do it again. There is an argument for education here instead of protecting against. But at the same time a `hop` onto the same `EventLoop` is hardly a performance hit.

fabianfett:
> @fabianfett one thing I considered was also exposing an allocator in the new context. do you think it would be a good addition or wait-and-see?

Yes. I think we should include it. If people need to make some json calls or whatever a allocator comes in handy.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -16,6 +16,33 @@ import Dispatch
...
+        ///
+        /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
+        public let logger: Logger
+
+        /// The `EventLoop` the Lambda is executed on. Use this to schedule work with.
+        /// This is useful when implementing the `EventLoopLambdaHandler` protocol.
+        ///
+        /// - note: The `EventLoop` is shared with the Lambda runtime engine and should be handled with extra care.
+        ///         Most importantly the `EventLoop` must never be blocked.
+        public let eventLoop: EventLoop
```
fabianfett, line 35:
```suggestion
        /// The `EventLoop` the Lambda is executed on. Use this to schedule work with.
        ///
        /// - note: The `EventLoop` is shared with the Lambda runtime engine and should be handled with extra care.
        ///         Most importantly the `EventLoop` must never be blocked.
        public let eventLoop: EventLoop
```

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -16,6 +16,33 @@ import Dispatch
 import Logging
 import NIO
 
+// MARK: - InitializationContext
+
+extension Lambda {
+    /// Lambda runtime initialization context.
+    /// The Lambda runtime generates and passes the `InitializationContext` to the Lambda handler as an argument.
```
fabianfett, line 23:
```suggestion
    /// The Lambda runtime generates and passes the `InitializationContext` to the Lambda factory as an argument.
```

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
fabianfett: removed: `/// This is useful when implementing the `EventLoopLambdaHandler` protocol.`

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
weissi: Just as a reference: Where ever NIO takes a future (not that many APIs outside of bootstraps), it'll `hop` it to the correct event loop.

tomerd:
> @tomerd Yes. I think we should include it. If people need to make some json calls or whatever a allocator comes in handy.

@fabianfett https://github.com/swift-server/swift-aws-lambda-runtime/pull/126/commits/f21f4a3722f1cf0803f68ba93c0d72d956f90276

PR #128, opened 2020-06-13
Added header when reporting failing invocations or initializations #116

### LINKED ISSUE #116: Lambda-Runtime-Function-Error-Type header missing when reporting errors.
According to the [documentation](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-initerror) we must set a special header (`"Lambda-Runtime-Function-Error-Type: Unhandled"`) when we report an error.

This must be done for both:
- [ ] initialization error
- [ ] invocation error

Ro-M:
### Motivation
This fixes #116 

We need to send a special header when we are reporting a failing invocation or when the runtime encounters an error during initialization.

### Changes
- The `lambda-runtime-function-error-type` header is added by the `RuntimeClient`. As of my understanding this should not be done by the `HTTPClient`
- To allow this I added the option to provide additional headers in the `post` method of `HTTPClient`. Those headers will be added to the default headers of the `HTTPClient`
- For testing purposes I opted into using the `NIOHTTP1TestServer` instead of the `MockLambdaServer`.
- I am using `RuntimeEngine`'s `baseURL` parameter as the primary source to set the `ip` and `port` property, falling back to searching in the environment if it wasn't provided. `baseURL` did exist as a parameter before but was unused. I needed a settable `baseURL` for the `NIOHTTP1TestServer`

### Discussion
- If we are adding an `additionalHeaders` parameter to the `post` method of the `HTTPClient`, should this parameter also be available in the `get` method? (It's not needed needed for this PR)
- If using the `NIOHTTP1TestServer` is ok but evaluating the inbound reads and outbound writes is too much boilerplate code: Should this evaluation be refactored into helper methods? (similar helper methods are internally used in the `swift-nio` package)
- Concerning the `baseURL` parameter of `RuntimeEngine`'s initializer: Directly mapping a url to the `ip` and `port` property does not sound right: I would not expect an `ip` property to contain a url. Maybe rename `ip` to `host`?


fabianfett:

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
```diff
@@ -43,13 +43,16 @@ internal final class HTTPClient {
...
-    func post(url: String, body: ByteBuffer?, timeout: TimeAmount? = nil) -> EventLoopFuture<Response> {
-        self.execute(Request(targetHost: self.targetHost,
-                             url: url,
-                             method: .POST,
-                             headers: HTTPClient.headers,
-                             body: body,
-                             timeout: timeout ?? self.configuration.requestTimeout))
+    func post(url: String, body: ByteBuffer?, timeout: TimeAmount? = nil, additionalHeaders: HTTPHeaders? = nil) -> EventLoopFuture<Response> {
+        var headers = HTTPClient.headers
+        additionalHeaders.flatMap { headers.add(contentsOf: $0) }
```
fabianfett, line 48: @tomerd I would vote to inject all headers on every call and to not have them split over two files. wdyt? In hindsight having static headers directly on the HTTPClient doesn't make much sense IMHO.

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTest.swift
```diff
@@ -209,6 +214,110 @@ class LambdaRuntimeClientTest: XCTestCase {
...
+        XCTAssertEqual(try JSONDecoder().decode(ErrorResponse.self, from: body).errorMessage, "boom")
+
+        XCTAssertEqual(try server.readInbound(), .end(nil))
+
+        XCTAssertNoThrow(try server.writeOutbound(.head(.init(version: .init(major: 1, minor: 1), status: .accepted))))
+        XCTAssertNoThrow(try server.writeOutbound(.end(nil)))
+        XCTAssertNoThrow(try result.wait())
+    }
+
+    func testSuccessHeaders() {
```
fabianfett, line 286: I think I would rename this to `testInvocationResponse` because it isn't as generic as `SuccessHeaders`

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
tomerd: hi @fabianfett do you mean have `RuntimeClient` fully control the headers instead of injecting "additional" ones? id so, that sounds good to me.

tomerd:

tomerd:
> Concerning the baseURL parameter of RuntimeEngine's initializer: Directly mapping a url to the ip and port property does not sound right: I would not expect an ip property to contain a url. Maybe rename ip to host?

happy to explore other names but keep in mind that Lambda will always bind to a localhost IP address

Ro-M:
> Lambda will always bind to a localhost IP address

@tomerd I didn't think about that, thanks! 
So `ip` and `port` seem to definitely be the correct names in that scenario and `baseURL` should be removed I guess?
If we keep `ip` and `port`, do we want to only set them through environment variables or do we want to provide a parameter (or multiple) to optionally set those properties via an initializer instead of the environment?

I would prefer the latter approach: providing the option to set them through an initializer but falling back to the environment and also default the `ip` and `port` parameters to nil. But on the other hand I can see the reasoning behind only using the environment if AWS is doing it that way ...

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTest.swift
Ro-M: Right. I think `Headers` is generally wrong at this point because the method is also testing a lot of other things as well.

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
Ro-M: Would it be ok to include that change in this PR or should I make a separate one for this?

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTest.swift
Ro-M: Is `testInvocationSuccessResponse`ok with you as well? To disambiguate between success and error tests

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
tomerd: I suggest we fix it in this PR

tomerd:
> If we keep ip and port, do we want to only set them through environment variables or do we want to provide a parameter (or multiple) to optionally set those properties via an initializer instead of the environment?

> I would prefer the latter approach: providing the option to set them through an initializer but falling back to the environment and also default the ip and port parameters to nil. But on the other hand I can see the reasoning behind only using the environment if AWS is doing it that way ...

using only env variables makes testing hard, so I also prefer the approach of being able to set it via the config and falling back to the env variable

tomerd:

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTest.swift
fabianfett: @Ro-M yes `testInvocationSuccessResponse` is perfect

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
Ro-M: Ok. I moved the static header declaration into an extension on `RuntimeClient`. The `get` and `post` methods on `HTTPClient` now have a non-optional `headers` parameter (without a default value). I also moved the `headers` parameter to come after the `url`parameter ... that ordering made a more cohesive impression to me ^^

Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift
```diff
@@ -16,6 +16,8 @@ import Logging
 import NIO
 import NIOHTTP1
 
+
+
```
tomerd, line 20: remove empty lines

Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift
```diff
@@ -26,7 +28,7 @@ extension Lambda {
         private let eventLoop: EventLoop
         private let allocator = ByteBufferAllocator()
         private let httpClient: HTTPClient
-
+
```
tomerd, line 31: run ./scripts/sanity

fabianfett:

tomerd:
thanks @Ro-M

PR #130, opened 2020-06-16
Added SES Event type

adam-fowler:
Add SES Lambda Event type 

### Motivation:

Want to support Lambdas triggered by SES

### Modifications:

- Added AWSLambdaEvents/SES.swift which include `SES.Event`
- Added `DateTimeCoding` propertyWrapper to support dates of the format "Wed, 7 Oct 2015 12:34:56 -0700".
- Added `SESTests`

### Result:

You can now decode SES events.

### DIFF
--- a/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
+++ b/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
@@ -83,7 +83,7 @@ public struct DateTimeCoding: Decodable {
         let dateString = try container.decode(String.self)
         guard let date = Self.dateFormatter.date(from: dateString) else {
             throw DecodingError.dataCorruptedError(in: container, debugDescription:
-                "Expected date to be in iso8601 date format with fractional seconds, but `\(dateString)` does not forfill format")
+                "Expected date to be in date-time format with fractional seconds, but `\(dateString)` does not forfill format")
         }
         self.wrappedValue = date
     }
--- a/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
+++ b/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
@@ -71,7 +71,7 @@ public struct ISO8601WithFractionalSecondsCoding: Decodable {
 }
 
 @propertyWrapper
-public struct DateTimeCoding: Decodable {
+public struct RFC5322DateTimeCoding: Decodable {
     public let wrappedValue: Date
 
     public init(wrappedValue: Date) {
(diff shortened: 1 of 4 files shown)


Sources/AWSLambdaEvents/Utils/DateWrappers.swift
```diff
@@ -67,3 +69,30 @@ public struct ISO8601WithFractionalSecondsCoding: Decodable {
...
+    public init(wrappedValue: Date) {
+        self.wrappedValue = wrappedValue
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let dateString = try container.decode(String.self)
+        guard let date = Self.dateFormatter.date(from: dateString) else {
+            throw DecodingError.dataCorruptedError(in: container, debugDescription:
+                "Expected date to be in iso8601 date format with fractional seconds, but `\(dateString)` does not forfill format")
```
fabianfett, line 86: Copy and paste?! `iso8601` 🙃

Sources/AWSLambdaEvents/Utils/DateWrappers.swift
```diff
@@ -67,3 +69,30 @@ public struct ISO8601WithFractionalSecondsCoding: Decodable {
         return formatter
     }
 }
+
+@propertyWrapper
+public struct DateTimeCoding: Decodable {
```
fabianfett, line 74: I'm unsure about the naming since it is very generic? Is there a special name for this format? @tomerd @weissi

tomerd: I agree that we need to find a more specific name. Is this used only in SES?

tomerd: lets also add a unit test for this

tomerd:

tomerd:

Sources/AWSLambdaEvents/Utils/DateWrappers.swift
adam-fowler: > I'm unsure about the naming since it is very generic? Is there a special name for this format? @tomerd @weissi 

RFC is particularly unhelpful https://tools.ietf.org/html/rfc5322#section-3.3

adam-fowler: It’s appears to be a standard internet format. It is the same one used in http headers.

Sources/AWSLambdaEvents/Utils/DateWrappers.swift
adam-fowler: Sorted but message may need to change if we get a more specific date format name.

Sources/AWSLambdaEvents/Utils/DateWrappers.swift
tomerd: AWS docs mention 5322: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/sending-concepts-email-format.html so perhaps we name it `RFC5322DateTimeCoding`?

adam-fowler: renamed

adam-fowler:
> looks great, but please add a quick unit test for the new date formatter

done

PR #132, opened 2020-06-18
add syncShutdown to LambdaHandler

tomerd:
motivation: make shutdown easier to use

changes:
* override shutdown to use the offloadQueue to perform syncShutdown
* add empty syncShutdown that can be implemented by the concrete Lambda function

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -62,6 +62,27 @@ public extension LambdaHandler {
     }
 }
 
+public extension LambdaHandler {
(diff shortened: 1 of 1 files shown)


Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -62,6 +62,27 @@ public extension LambdaHandler {
...
+            } catch {
+                promise.fail(error)
+            }
+        }
+        return promise.futureResult
+    }
+
+    /// Clean up the `LambdaHandler` resources synchronously.
+    /// Concrete Lambda handlers implement this method to shutdown resources like `HTTPClient`s and database connections.
+    func syncShutdown() throws {
```
fabianfett, line 81: do we want to insert the `ShutdownContext` here?

PR #133, opened 2020-06-19
examples deployment scripts refresh

tomerd:
motivation: make sure examples are up tp date

changes:
* use swift:5.2 images instead of nighlies
* compute dependencies to make zipfile as small as possible
* make scripts more consistent

### DIFF
--- a/Examples/LambdaFunctions/scripts/build-and-package.sh
+++ b/Examples/LambdaFunctions/scripts/build-and-package.sh
@@ -19,12 +19,21 @@ executable=$1
 workspace="$(pwd)/../.."
 
 echo "-------------------------------------------------------------------------"
+echo "preparing docker build image"
+echo "-------------------------------------------------------------------------"
+docker build . -t builder
+echo "done"
+
+echo "-------------------------------------------------------------------------"
 echo "building \"$executable\" lambda"
 echo "-------------------------------------------------------------------------"
-docker run --rm -v "$workspace":/workspace -w /workspace/Examples/LambdaFunctions builder bash -cl "swift build --product $executable -c release -Xswiftc -g"
+docker run --rm -v "$workspace":/workspace -w /workspace/Examples/LambdaFunctions builder \
+       bash -cl "swift build --product $executable -c release"
 echo "done"
 
 echo "-------------------------------------------------------------------------"
 echo "packaging \"$executable\" lambda"
 echo "-------------------------------------------------------------------------"
-docker run --rm -v "$workspace":/workspace -w /workspace/Examples/LambdaFunctions builder bash -cl "./scripts/package.sh $executable"
+docker run --rm -v "$workspace":/workspace -w /workspace/Examples/LambdaFunctions builder \
+       bash -cl "./scripts/package.sh $executable"
+echo "done"
--- a/Examples/LambdaFunctions/scripts/deploy.sh
+++ b/Examples/LambdaFunctions/scripts/deploy.sh
@@ -17,42 +17,25 @@ set -eu
 
 # Lambda Function name (must exist in AWS Lambda)
 lambda_name=SwiftSample
-
 # S3 bucket name to upload zip file (must exist in AWS S3)
 s3_bucket=swift-lambda-test
-
-workspace="$(pwd)/../.."
 
 DIR="$(cd "$(dirname "$0")" && pwd)"
 source $DIR/config.sh
 
+workspace="$DIR/../.."
+
 echo -e "\ndeploying $executable"
 
-echo "-------------------------------------------------------------------------"
-echo "preparing docker build image"
-echo "-------------------------------------------------------------------------"
-docker build . -t builder
-
-echo "-------------------------------------------------------------------------"
-echo "building \"$executable\" lambda"
-echo "-------------------------------------------------------------------------"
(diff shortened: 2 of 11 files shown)


Examples/LambdaFunctions/scripts/package.sh
```diff
@@ -17,11 +17,15 @@ set -eu
...
 
-target=.build/lambda/$executable
+target=".build/lambda/$executable"
 rm -rf "$target"
 mkdir -p "$target"
 cp ".build/release/$executable" "$target/"
-cp -Pv /usr/lib/swift/linux/lib*so* "$target"
+#cp -Pv /usr/lib/swift/linux/lib*so* "$target"
+# add the target deps based on ldd
+ldd ".build/release/$executable" | grep swift | awk '{print $3}' | xargs cp -Pv -t "$target"
```
tomerd, line 26: 👀 this is the most interesting part

Examples/LambdaFunctions/scripts/build-and-package.sh
```diff
@@ -18,13 +18,22 @@ set -eu
...
+echo "-------------------------------------------------------------------------"
+docker build . -t builder
+echo "done"
+
 echo "-------------------------------------------------------------------------"
 echo "building \"$executable\" lambda"
 echo "-------------------------------------------------------------------------"
-docker run --rm -v "$workspace":/workspace -w /workspace/Examples/LambdaFunctions builder bash -cl "swift build --product $executable -c release -Xswiftc -g"
+docker run --rm -v "$workspace":/workspace -w /workspace/Examples/LambdaFunctions builder \
+       bash -cl "swift build --product $executable -c release"
```
fabianfett, line 31: Just to fill a knowledge gap on my site: What's the benefit of using `bash -cl "cmd"` vs. just `cmd`

Examples/LambdaFunctions/scripts/deploy.sh
```diff
@@ -17,42 +17,25 @@ set -eu
 
 # Lambda Function name (must exist in AWS Lambda)
 lambda_name=SwiftSample
-
 # S3 bucket name to upload zip file (must exist in AWS S3)
 s3_bucket=swift-lambda-test
```
fabianfett, line 21: As mentioned in some issue, we should make this bucket configurable since bucket names must be unique worldwide.

Examples/LambdaFunctions/scripts/package.sh
```diff
@@ -17,11 +17,15 @@ set -eu
 
 executable=$1
 
-target=.build/lambda/$executable
+target=".build/lambda/$executable"
 rm -rf "$target"
 mkdir -p "$target"
 cp ".build/release/$executable" "$target/"
-cp -Pv /usr/lib/swift/linux/lib*so* "$target"
+#cp -Pv /usr/lib/swift/linux/lib*so* "$target"
```
fabianfett, line 24: leftover?

Examples/LambdaFunctions/scripts/package.sh
fabianfett: I'm not sure, but could we copy the libicu libs directly (and not just the symlinks), if we specify `xargs cp -Lv -t "$target"`

From the `man cp`
```
      -L    If the -R option is specified, all symbolic links are followed.
      -P    If the -R option is specified, no symbolic links are followed.  This is the default.
```

cc @drexin

Examples/LocalDebugging/MyLambda/scripts/package.sh
```diff
@@ -17,11 +17,15 @@ set -eu
...
+target=".build/lambda/$executable"
 rm -rf "$target"
 mkdir -p "$target"
 cp ".build/release/$executable" "$target/"
-cp -Pv /usr/lib/swift/linux/lib*so* "$target"
+#cp -Pv /usr/lib/swift/linux/lib*so* "$target"
+# add the target deps based on ldd
+ldd ".build/release/$executable" | grep swift | awk '{print $3}' | xargs cp -Pv -t "$target"
+# add icu (ldd just gets the symlink)
+cp -Pv /usr/lib/swift/linux/libicu*so* "$target"
```
fabianfett, line 28: same as above ^^

Examples/LambdaFunctions/scripts/build-and-package.sh
tomerd: old habit to make sure we get login shell: https://www.gnu.org/software/bash/manual/html_node/Invoking-Bash.html

Examples/LambdaFunctions/scripts/deploy.sh
tomerd: we should prompt the user for the name of the bucket and lambda. is this what you are suggesting?

Examples/LambdaFunctions/scripts/package.sh
tomerd: yes, intentional for reference while we discuss the "new" way

Examples/LambdaFunctions/scripts/deploy.sh
fabianfett: yes. maybe store it between runs?

tomerd: added a prompt. wdyt?

tomerd:
@fabianfett updated

PR #135, opened 2020-06-26
SES time zone

adam-fowler:
Support decoding time zones in the following format `Fri, 26 Jun 2020 03:04:03 -0500 (CDT)`

### Motivation:

Currently if an email is received with the common header date which includes the timezone in alphabetical form in brackets after the numeric timezone the date does not decode, and SES.Event does not decode.

### Modifications:

Code changes include cropping text after any open bracket found in the date string before passing it through the dateFormatter. I don't know of a dataFormat string that will support the above format. If there is one obviously it would be better to use that instead. 

### Result:

SES.Event decodes even when it holds a date in the above format

### DIFF
--- a/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
+++ b/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
@@ -82,7 +82,7 @@ public struct RFC5322DateTimeCoding: Decodable {
         let container = try decoder.singleValueContainer()
         let string = try container.decode(String.self)
         let bracket = string.firstIndex(of: "(") ?? string.endIndex
-        let dateString = String(string[string.startIndex..<bracket])
+        let dateString = String(string[string.startIndex ..< bracket])
         guard let date = Self.dateFormatter.date(from: dateString) else {
             throw DecodingError.dataCorruptedError(in: container, debugDescription:
                 "Expected date to be in RFC5322 date-time format with fractional seconds, but `\(string)` does not forfill format")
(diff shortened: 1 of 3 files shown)


Sources/AWSLambdaEvents/Utils/DateWrappers.swift
```diff
@@ -80,10 +80,12 @@ public struct RFC5322DateTimeCoding: Decodable {
 
     public init(from decoder: Decoder) throws {
         let container = try decoder.singleValueContainer()
-        let dateString = try container.decode(String.self)
+        let string = try container.decode(String.self)
+        let bracket = string.firstIndex(of: "(") ?? string.endIndex
+        let dateString = String(string[string.startIndex ..< bracket])
```
fabianfett, line 85: I would love to see a comment here why we are doing this.

tomerd: could we potentially save an allocation by making the string a var and rewriting it only if we find a bracket?

adam-fowler: both of them are done

adam-fowler:
> orthogonal to these changes but we should add a link to SES event from https://github.com/swift-server/swift-aws-lambda-runtime#integration-with-aws-platform-events

Added the link in

tomerd:
thanks @adam-fowler

Sources/AWSLambdaEvents/Utils/DateWrappers.swift
rickky008: Like

PR #136, opened 2020-06-29
!context reordering context towards common guideline / style for context passing

### LINKED ISSUE #89: Reorder context, future-proof for a tracing context world

ktoso:
PR for discussing and potentially merging the context reordering, in accordance with some guidelines around context passing style we're trying to align SSWG projects on.

This is not yet set in stone but since I had a moment between bigger tasks wanted to get this PR out of the way so we can sit on it for a bit.

### Motivation:

This PR contains the full writeup of the guidelines we are considering for contexts in general: https://github.com/slashmo/gsoc-swift-baggage-context/pull/2/files

It's about the baggage context but it applies to contexts in general, and I think we're pretty close to aligning many libs to this style; including gRPC (already is), HTTPClient (kind of is, with Logger today and baggage context in the future) other Lambda SDKs as well. 

### Modifications:

Reorder context parameter.

### Result:

- Resolves #89

### DIFF
--- a/Examples/LambdaFunctions/Sources/APIGateway/main.swift
+++ b/Examples/LambdaFunctions/Sources/APIGateway/main.swift
@@ -27,7 +27,7 @@ struct APIGatewayProxyLambda: EventLoopLambdaHandler {
     public typealias In = APIGateway.V2.Request
     public typealias Out = APIGateway.V2.Response
 
-    public func handle(context: Lambda.Context, event: APIGateway.V2.Request) -> EventLoopFuture<APIGateway.V2.Response> {
+    public func handle(event: APIGateway.V2.Request, context: Lambda.Context) -> EventLoopFuture<APIGateway.V2.Response> {
         context.logger.debug("hello, api gateway!")
         return context.eventLoop.makeSucceededFuture(APIGateway.V2.Response(statusCode: .ok, body: "hello, world!"))
     }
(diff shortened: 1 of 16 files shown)


tomerd:
hi @ktoso if you want to keep this around, could you please update this PR to be against the `main` branch so we can delete the old `master` branch

PR #138, opened 2020-07-02
WIP: Mock API Gateway response

pokryfka:
### Motivation:

Allow user to configure `LocalLambda.Server` to create invocations:
- mocking different AWS services (for example `API Gateway`)
- with custom events (for example `AppSync` events) or
- implementing specific use cases (for example injecting `Authorization` header/logic).

### Modifications:

This picks up and updates https://github.com/swift-server/swift-aws-lambda-runtime/pull/83

- The `LocalLambda.Server` opens two ports:
  - one to mock HTTP API for lambda custom runtimes - `ControlPlaneHandler`
  - one to create new invocations - `InvokeHandler`
- The ServerState (TODO: rename) encapsulates all the "business" logic
  - queueing incoming requests
  - forwarding them to lambda handler
  - forwarding response from lambda handler back to the `InvokeHandler` which created them
- user can provide custom implementation of `LocalLambdaInvocationProxy`, defaults to `InvokeProxy` (current behavior)

the main difference from https://github.com/swift-server/swift-aws-lambda-runtime/pull/83 is `bootstrap` and separation of `ProxyType` from `ServerState`;
`ServerState` provides API to queue new invocations and promises to fulfill them but it has no interest how they were created and what is going to happen with the result

### Result:

Allows to setup `LocalLambda` providing custom implementation of `LocalLambdaInvocationProxy`, `InvokeProxy` by default.

```swift
#if DEBUG
LocalLambda.bootstrap(InvokeProxy.self)
#endif

```

If the change is approved I can create `APIGatewayV2Proxy` in `AWSLambdaEvents` which could be used to `bootstrap` `LocalLambda`:

```swift
#if DEBUG
LocalLambda.bootstrap(APIGatewayV2Proxy.self)
#endif

```


pokryfka:
I missed https://github.com/swift-server/swift-aws-lambda-runtime/pull/83 which defines `LocalLambdaInvocationProxy` and implements `InvokeProxy`.

fabianfett:
@pokryfka the work on this pr has stopped, since we chose to go a different route with configuring the `LocalServer`. Since the `LocalServer` is now automatically started if the ENV variable is set, we need to find a different way. One idea is to provide a `bootstrap` method like in `swift-logging` to configure the LocalServer before actual startup. Does this make sense?

tomerd:
@pokryfka should we pursue this further in the context of #83?

Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
```diff
@@ -46,62 +45,269 @@ extension Lambda {
...
+public protocol LocalLambdaInvocationProxy {
+    init(eventLoop: EventLoop)
+
+    // TODO: in most (all?) cases we do not need async interface
+
+    /// throws `LocalLambda.InvocationError` if request is invalid
+    func invocation(from request: LocalLambda.HTTPRequest) -> EventLoopFuture<ByteBuffer>
+    func processResult(_ result: ByteBuffer?) -> EventLoopFuture<LocalLambda.HTTPResponse>
+    func processError(_ error: ByteBuffer?) -> EventLoopFuture<LocalLambda.HTTPResponse>
+}
```
pokryfka, line 57: I think in most cases there is no need for async calls.

In a typical scenario a HTTP request would be used to create new HTTP request with different payload.

Async interface would be useful in a fairly exotic scenario (as far as local testing/development is concerned); for example, if the `LocalLambdaInvocationProxy` was connecting with external server to set valid Authorization header.

Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
```diff
@@ -46,62 +45,269 @@ extension Lambda {
...
+    }
+
+    /// Mocks an HTTP API for custom runtimes to receive invocation events from Lambda and send response data back, see
+    /// [AWS Lambda runtime interface](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html).
+    private final class ControlPlaneHandler: ChannelInboundHandler {
+        typealias InboundIn = HTTPServerRequestPart
+        typealias OutboundOut = HTTPServerResponsePart
+
+        // TODO: do we need a circular buffer? there is only (up to) one request at a time
+        private var pending = CircularBuffer<(head: HTTPRequestHead, body: ByteBuffer?)>()
```
pokryfka, line 302: do we need a circular buffer? there is only (up to) one request at a time

pokryfka:
@tomerd @fabianfett please let me know your thoughts on this, a few comments/questions inline (above)

tomerd:
hi @parkera if you want to keep this around, could you please update this PR to be against the `main` branch so we can delete the old `master` branch

PR #139, opened 2020-07-02
Generate trace ID in correct format (LocalLambda Server)

pokryfka:
Generate (X-Ray) Trace ID in correct format.

### Motivation:

Generate (X-Ray) Trace ID in correct format. Helpful when testing X-Ray tracing.

#### References

- [Generating trace IDs](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids)
- [Tracing header](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader)

### Modifications:

- different value of [Tracing header](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader) generated by `LocalLambda Server` with:
  - correctly formatted trace ID (root)
  - removed parent segment ID (which is optional)

### Result:

Example value of  `Lambda-Runtime-Trace-Id` header:

```
Root=1-5efd7f0b-da552a1eb226e4718fff8790;Sampled=1
```

before (root and parent values used to be signed integers):

```
"Root=25457;Parent=14820;Sampled=1"
```

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Utils.swift
+++ b/Sources/AWSLambdaRuntimeCore/Utils.swift
@@ -104,3 +104,27 @@ extension String {
         bytes.append(UInt8(ascii: "\""))
     }
 }
+
+/// Generates (X-Ray) trace ID.
+/// # Trace ID Format
+/// A `trace_id` consists of three numbers separated by hyphens.
+/// For example, `1-58406520-a006649127e371903a2de979`. This includes:
+/// - The version number, that is, 1.
+/// - The time of the original request, in Unix epoch time, in **8 hexadecimal digits**.
+/// For example, 10:00AM December 1st, 2016 PST in epoch time is `1480615200` seconds, or `58406520` in hexadecimal digits.
+/// - A 96-bit identifier for the trace, globally unique, in **24 hexadecimal digits**.
+/// # References
+/// - [Generating trace IDs](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids)
+/// - [Tracing header](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader)
+func generateTraceID() -> String {
+    // The version number, that is, 1.
+    let version: UInt = 1
+    // The time of the original request, in Unix epoch time, in 8 hexadecimal digits.
+    let now = UInt32(DispatchWallTime.now().millisSinceEpoch / 1000)
+    let dateValue = String(now, radix: 16, uppercase: false)
+    let datePadding = String(repeating: "0", count: max(0, 8 - dateValue.count))
+    // A 96-bit identifier for the trace, globally unique, in 24 hexadecimal digits.
+    let identifier = String(UInt64.random(in: UInt64.min ... UInt64.max) | 1 << 63, radix: 16, uppercase: false)
+        + String(UInt32.random(in: UInt32.min ... UInt32.max) | 1 << 31, radix: 16, uppercase: false)
+    return "\(version)-\(datePadding)\(dateValue)-\(identifier)"
+}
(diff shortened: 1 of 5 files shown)


pokryfka:
Should not the value of `AmazonHeaders.traceID` be `X-Amzn-Trace-Id` instead of `Lambda-Runtime-Trace-Id`?

see [Tracing header](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader)

pokryfka:
> @pokryfka This is awesome! I'm fine with this change but let's wait for @tomerd's opinion. Further, would you mind creating a test that verifies the format?

Sure, I can create a test.

In which case the function will need to be internal (its defined within private `LocalLambda` enum at the moment),
in which case.. should I move it to `Utils` or keep it in `Lambda+LocalServer`?
The function is generic though probably will only be used by the `LocalLambda:Server`.

@fabianfett

tomerd:

tomerd:
thanks @pokryfka this is great. I think there are also places in the unit tests where we fake the traceID and would be nice to use a more correct one. 

wondering if we should rename the function to be generateXRayTraceID to clarify the intent. I also always like to add the "internal" modifier to make it super clear

pokryfka:
@tomerd  Thank you for the feedback. Please see 2 questions below.

I could rename the function or/and create `XRay` namespace?

```swift
internal enum XRay {
    static func generateTraceID() -> String {
        // ...
    }
}
```

XRay related functions or types would end up there **if** needed or practical, for example we could also have:

```swift
internal enum XRay {
    static func generateTracingHeader(parent: String? = nil, sampled: Bool? = nil) -> String {
        // ...
    }
}
```

---

At the moment `LocalLambda:Server` is the only place where tracing header value is generated.

`MockLambdaServer` in `AWSLambdaRuntimeCoreTests` creates the tracing header using (correctly formatted) string literal; I dont think the actual values are used during tests; should they be generated?

```
                responseHeaders = [
                    (AmazonHeaders.requestID, requestId),
                    (AmazonHeaders.invokedFunctionARN, "arn:aws:lambda:us-east-1:123456789012:function:custom-runtime"),
                    (AmazonHeaders.traceID, "Root=1-5bef4de7-ad49b0e87f6ef6c87fc2e700;Parent=9a9197af755a6419;Sampled=1"),
                    (AmazonHeaders.deadline, String(deadline)),
                ]
```

tomerd:
@parkera wdyt about adding it to `AmazonHeaders`?

pokryfka:
@tomerd made both changes, please let me know if you have more comments

Sources/AWSLambdaRuntimeCore/Utils.swift
```diff
@@ -104,3 +104,30 @@ extension String {
...
+    /// A `trace_id` consists of three numbers separated by hyphens.
+    /// For example, `1-58406520-a006649127e371903a2de979`. This includes:
+    /// - The version number, that is, 1.
+    /// - The time of the original request, in Unix epoch time, in **8 hexadecimal digits**.
+    /// For example, 10:00AM December 1st, 2016 PST in epoch time is `1480615200` seconds, or `58406520` in hexadecimal digits.
+    /// - A 96-bit identifier for the trace, globally unique, in **24 hexadecimal digits**.
+    /// # References
+    /// - [Generating trace IDs](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids)
+    /// - [Tracing header](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader)
+    @inlinable
```
tomerd, line 120: is @inlinable needed here?

tomerd:

tomerd:
@pokryfka looks great, one question

Sources/AWSLambdaRuntimeCore/Utils.swift
pokryfka: removed

tomerd:

tomerd:
thank you @pokryfka <3

PR #140, opened 2020-07-02
Use URLSession dataTaskPublisher in iOS demo app

pokryfka:
Use URLSession dataTaskPublisher in iOS demo app

### Motivation:

`URLSession`'s `dataTaskPublisher` makes the API code more concise and the logic more clear.

Arguably `Combine` is a better match for a SwiftUI app.

On the downside, some validation like HTTP status code, has been removed.
(Could be restored with `tryMap`).

### Modifications:

- using `URLSession` `dataTaskPublisher` to make the HTTP query
- small layout changes so that the message (or error) is not truncated

### Result:

```swift
URLSession.shared.dataTaskPublisher(for: request)
    .map(\.data)
    .decode(type: Response.self, decoder: JSONDecoder())
    .map(\.message)
    .catch { Just("Error: \($0.localizedDescription)") }
    .receive(on: DispatchQueue.main)
    .assign(to: \.message, on: self)
```

### DIFF
--- a/Examples/LocalDebugging/MyApp/MyApp/ContentView.swift
+++ b/Examples/LocalDebugging/MyApp/MyApp/ContentView.swift
@@ -12,20 +12,51 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Combine
 import Shared
 import SwiftUI
 
 struct ContentView: View {
+    class API: ObservableObject {
+        let url = URL(string: "http://localhost:7000/invoke")!
+
+        @Published var message: String = ""
+
+        private var task: AnyCancellable?
+
+        func register(name: String, password: String) {
+            self.task?.cancel()
+
+            var request = URLRequest(url: url)
+            request.httpMethod = "POST"
+
+            guard let jsonRequest = try? JSONEncoder().encode(Request(name: name, password: password)) else {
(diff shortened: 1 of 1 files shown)


tomerd:
> this is super nice! my only concern it may be less-friendly to beginners. iow, given that the goal of the examples is to make things as approachable as possible, the additional cognitive complexity in combine’s functional composition may be too much?

@pokryfka @fabianfett wdyt about

pokryfka:
> this is super nice! my only concern it may be less-friendly to beginners. iow, given that the goal of the examples is to make things as approachable as possible, the additional cognitive complexity in combine’s functional composition may be too much?

@tomerd thank you for your feedback,

this argument makes complete sense to me although, following the same reasoning, `UIKit` could be favored over `SwiftUI` which is new to everyone.

this is a cosmetic change of an example, if there are any concerns I think let's just close it

ktoso:
I'm in favor of this change -- swiftui and publishers is "the" way nowadays after all IMO. 👍

tomerd:
hi @parkera if you want to keep this around, could you please update this PR to be against the `main` branch so we can delete the old `master` branch

PR #141, opened 2020-07-17
RFC: Added ServiceLifecycle dependency and started integration

fabianfett:
The current start and shutdown services is okay but not great. Thankfully `ServiceLifecycle` has been released two days ago. Let's see if we can use it for AWS Lambda Runtime.

This is a draft pr, exploring possibilities and starting a discussion.

### Open ends/questions

- In the first approach we only expose the `ServiceLifecycle` to the user but don't use it ourselves, even though we could. But for this we will need to think about state handling (current state of mind):
  - Currently lambda can be shutdown because of two reasons: 
    - The user cancels the process (only possible for DEBUG builds)
    - The Lambda can not communicate correctly with the Control Plane API
  - We could handle the first shutdown case with the new ServiceLifecycle fairly easy.
- Do we want to deprecate the shutdown methods? So that those can be removed before `1.0`?

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -20,6 +20,7 @@ import Darwin.C
 
 import Backtrace
 import Logging
+import Lifecycle
 import NIO
 
 public enum Lambda {
@@ -101,13 +102,15 @@ public enum Lambda {
     // for testing and internal use
     internal static func run(configuration: Configuration = .init(), factory: @escaping HandlerFactory) -> Result<Int, Error> {
         let _run = { (configuration: Configuration, factory: @escaping HandlerFactory) -> Result<Int, Error> in
-            Backtrace.install()
             var logger = Logger(label: "Lambda")
             logger.logLevel = configuration.general.logLevel
+            
+            // we don't intercept the shutdown signal here yet.
+            let serviceLifecycle = ServiceLifecycle(configuration: .init(logger: logger, shutdownSignal: [], installBacktrace: true))
 
             var result: Result<Int, Error>!
             MultiThreadedEventLoopGroup.withCurrentThreadAsEventLoop { eventLoop in
-                let lifecycle = Lifecycle(eventLoop: eventLoop, logger: logger, configuration: configuration, factory: factory)
+                let lifecycle = Lifecycle(eventLoop: eventLoop, serviceLifecycle: serviceLifecycle, logger: logger, configuration: configuration, factory: factory)
                 #if DEBUG
                 let signalSource = trap(signal: configuration.lifecycle.stopSignal) { signal in
                     logger.info("intercepted signal: \(signal)")
@@ -121,11 +124,19 @@ public enum Lambda {
                     #if DEBUG
                     signalSource.cancel()
                     #endif
-                    eventLoop.shutdownGracefully { error in
+                    
+                    serviceLifecycle.shutdown { (error) in
                         if let error = error {
-                            preconditionFailure("Failed to shutdown eventloop: \(error)")
+                            preconditionFailure("Failed to shutdown service: \(error)")
+                        }
+                        
+                        eventLoop.shutdownGracefully { error in
+                            if let error = error {
+                                preconditionFailure("Failed to shutdown eventloop: \(error)")
+                            }
                         }
                     }
+                    
                     result = lifecycleResult
                 }
             }
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import Dispatch
+import Lifecycle
 import Logging
 import NIO
 
@@ -32,13 +33,17 @@ extension Lambda {
         /// - note: The `EventLoop` is shared with the Lambda runtime engine and should be handled with extra care.
         ///         Most importantly the `EventLoop` must never be blocked.
         public let eventLoop: EventLoop
+        
+        /// `ServiceLifecycle` to register services with
+        public let serviceLifecycle: ServiceLifecycle
 
         /// `ByteBufferAllocator` to allocate `ByteBuffer`
         public let allocator: ByteBufferAllocator
 
-        internal init(logger: Logger, eventLoop: EventLoop, allocator: ByteBufferAllocator) {
+        internal init(logger: Logger, eventLoop: EventLoop, serviceLifecycle: ServiceLifecycle, allocator: ByteBufferAllocator) {
             self.eventLoop = eventLoop
             self.logger = logger
+            self.serviceLifecycle = serviceLifecycle
             self.allocator = allocator
         }
     }
--- a/Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import Logging
+import Lifecycle
 import NIO
 import NIOConcurrencyHelpers
 
@@ -22,6 +23,7 @@ extension Lambda {
     /// - note: It is intended to be used within a single `EventLoop`. For this reason this class is not thread safe.
     public final class Lifecycle {
         private let eventLoop: EventLoop
+        private let serviceLifecycle: ServiceLifecycle
         private let shutdownPromise: EventLoopPromise<Int>
         private let logger: Logger
         private let configuration: Configuration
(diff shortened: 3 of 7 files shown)


Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -118,11 +121,19 @@ public enum Lambda {
...
                     #endif
-                    eventLoop.shutdownGracefully { error in
+                    
+                    serviceLifecycle.shutdown { (error) in
                         if let error = error {
-                            preconditionFailure("Failed to shutdown eventloop: \(error)")
+                            preconditionFailure("Failed to shutdown service: \(error)")
+                        }
+                        
+                        eventLoop.shutdownGracefully { error in
```
ktoso, line 130: shouldn't that also be triggered by the lifecycle shutdown as a lifecycle task?

Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -118,11 +121,19 @@ public enum Lambda {
                     #if DEBUG
                     signalSource.cancel()
                     #endif
-                    eventLoop.shutdownGracefully { error in
+                    
+                    serviceLifecycle.shutdown { (error) in
```
ktoso, line 125: why `(error)`?

Sources/AWSLambdaRuntimeCore/Lambda.swift
tomerd: I think thats a good idea: register the event loop in L110 then just call  serviceLifecycle.shutdown

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -32,13 +33,17 @@ extension Lambda {
         /// - note: The `EventLoop` is shared with the Lambda runtime engine and should be handled with extra care.
         ///         Most importantly the `EventLoop` must never be blocked.
         public let eventLoop: EventLoop
+        
+        /// `ServiceLifecycle` to register services with
+        public let serviceLifecycle: ServiceLifecycle
```
tomerd, line 38: maybe simplify name to "lifecycle"?

Sources/AWSLambdaRuntimeCore/LambdaLifecycle.swift
```diff
@@ -22,6 +23,7 @@ extension Lambda {
     /// - note: It is intended to be used within a single `EventLoop`. For this reason this class is not thread safe.
     public final class Lifecycle {
         private let eventLoop: EventLoop
+        private let serviceLifecycle: ServiceLifecycle
```
tomerd, line 26: ditto on name

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -34,19 +35,34 @@ extension Lambda {
...
+                // Hopping back to "our" EventLoop is important in case the factory returns a future
                 // that originated from a foreign EventLoop/EventLoopGroup.
                 // This can happen if the factory uses a library (let's say a database client) that manages its own threads/loops
                 // for whatever reason and returns a future that originated from that foreign EventLoop.
                 .hop(to: self.eventLoop)
+                .flatMap { (handler) in
+                    let promise = self.eventLoop.makePromise(of: ByteBufferLambdaHandler.self)
+                    // after we have created the LambdaHandler we must now start the services.
+                    // in order to not have to map once our success case returns the handler.
+                    serviceLifecycle.start { (error) in
```
tomerd, line 56: I wonder if we should add sugar in the lifecycle NIO compact module to make this kind of integration easier. wdyt?

Tests/AWSLambdaRuntimeCoreTests/LambdaLifecycleTest.swift
```diff
@@ -68,6 +74,12 @@ class LambdaLifecycleTest: XCTestCase {
         defer { XCTAssertNoThrow(try server.stop().wait()) }
         let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
         defer { XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully()) }
+        
+        let serviceLifecycle = ServiceLifecycle(configuration: .init(shutdownSignal: [], installBacktrace: false))
+        defer {
+            serviceLifecycle.shutdown()
+            serviceLifecycle.wait()
+        }
```
tomerd, line 82: I wonder if we can now simplify the shutdown logic a bit by relying more on the users using lifecycle

Tests/AWSLambdaRuntimeCoreTests/Utils.swift
```diff
@@ -29,7 +30,14 @@ func runLambda(behavior: LambdaServerBehavior, factory: @escaping Lambda.Handler
     let runner = Lambda.Runner(eventLoop: eventLoopGroup.next(), configuration: configuration)
     let server = try MockLambdaServer(behavior: behavior).start().wait()
     defer { XCTAssertNoThrow(try server.stop().wait()) }
-    try runner.initialize(logger: logger, factory: factory).flatMap { handler in
+    
+    let serviceLifecycle = ServiceLifecycle(configuration: .init(shutdownSignal: [], installBacktrace: false))
+    defer {
+        serviceLifecycle.shutdown()
+        serviceLifecycle.wait()
+    }
```
tomerd, line 38: I guess we can register the server with lifecycle but maybe will make the test harder to reason about

tomerd:
this is cool. added some ideas to discuss

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
rickky008: Thank you

fabianfett: Not sure if just `lifecycle` might add confusion because of the `Lambda.Lifecycle`

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
fabianfett: https://github.com/swift-server/swift-service-lifecycle/pull/72

fabianfett:
Hi @tomerd,

I've taken some time to think about the `ServiceLifecycle` some more. Now, I'm not sure if there is way to bring the two mental models of the lifecycle together:

- Lambda Runtime follows a synchronous pattern of Service creation and start
- ServiceLifecycle follows an async pattern of service registration, start and stop. Services can only be registered, before the ServiceLifecycle is started. Only those services will be shutdown that have been started.

Let's look at a more complex `LambdaHandler`. (Sorry that the example has gotten so big, but we need something complex here...). The `static func create` is our `LambdaFactory` that is called during startup. In this method we get some credentials from the AWS Parameter Store, that we use later to establish a connection to a database. This is an AWS recommended pattern and we use libraries common in the swift server realm. (AsyncHTTPClient, Soto, MySQLKit)

If we the credentials can't be obtained from the parameter store, we must fail the Lambda in startup and report an error. We do this by using `EventLoopFuture`s. In the case of an error, while getting the credentials, the `httpClient` and the `awsClient` must be shutdown, since they already have internal state (an open http connection). 

Right now I can't see how we can achieve this using `ServiceLifecycle` – I don't know when we should call `start` on the Lifecycle... There are three options:

1. Call `ServiceLifecycle.start` before the `create` function. In this case we can't register any service to the Lifecycle.
2. Call `ServiceLifecycle.start` after the successful `create` function. In this case we can't use the Lifecycle to shutdown services that have already been started (in this case `httpClient` and `awsClient`), since only `Tasks` that have been started will be stopped. If we can't get the parameters in this example neither `httpClient` nor `awsClient` could be stopped.
3. **Dirty hack, but only option that i can see right now to make it work:** We call `ServiceLifecycle.start` after `create` – no matter the outcome of the `LambdaFactory`. If the factory has failed, we assume that developers have only used `registerShutdown` for their services and registered only services that are already started. Hence we could assume that all registered services have been started and have therefore an empty start handler in their Task. We could then shut them down immediately after their empty start. But in this case we would use the `ServiceLifecycle` in a very particular way (which probably wasn't intended at all).

```swift
import AsyncHTTPClient
import AWSLambdaEvents
import AWSLambdaRuntimeCore
import Dispatch
import Logging
import MySQLKit
import NIO
import SotoSSM

public class MyLambdaHandler: EventLoopLambdaHandler {
    public typealias In = SNS.Event
    public typealias Out = Void

    let httpClient: HTTPClient
    let database: EventLoopGroupConnectionPool<MySQLConnectionSource>

    public static func create(_ context: Lambda.InitializationContext) -> EventLoopFuture<ByteBufferLambdaHandler> {
        let eventLoop = context.eventLoop
        let httpClient = HTTPClient(eventLoopGroupProvider: .shared(eventLoop))
        // serviceLifecycle.registerShutdown(label: "httpClient") { (completion) in
        //   httpClient.shutdown(completion)
        // }
        
        let awsClient = AWSClient(httpClientProvider: .shared(httpClient))
        // serviceLifecycle.registerShutdown(label: "awsClient") { (completion) in
        //   awsClient.shutdown(completion)
        // }
        let client = SSM(client: awsClient, region: .useast1)

        return eventLoop.makeSucceededFuture(()).flatMapThrowing { (_) -> String in
            let stageName = try Lambda.env(for: "STAGE")

            return stageName
        }.flatMap { (stageName) -> EventLoopFuture<(SSM.GetParametersResult, String)> in
            let parameters = [
                "/\(stageName)/database/username",
                "/\(stageName)/database/password",
            ]

            return client.getParameters(.init(names: parameters)).map { ($0, stageName) }
        }.flatMapThrowing { parameters, stageName in
            let config = MySQLConfiguration(
                hostname: try Lambda.env(for: "DATABASE_HOST"),
                username: try parameters.value(for: "/\(stageName)/database/username"),
                password: try parameters.value(for: "/\(stageName)/database/password"),
                database: try Lambda.env(for: "DATABASE_NAME"),
                tlsConfiguration: nil
            )

            let pool = EventLoopGroupConnectionPool(
                source: MySQLConnectionSource(configuration: config),
                on: eventLoop
            )
            // serviceLifecycle.registerShutdown(label: "database-pool") { (completion) in
            //   pool.shutdown(completion)
            // }
            return pool
        }.flatMapThrowing { (pool) -> ByteBufferLambdaHandler in
            MyLambdaHandler(database: pool, httpClient: httpClient)
        }
    }

    init(database: EventLoopGroupConnectionPool<MySQLConnectionSource>, httpClient: HTTPClient) {
        self.database = database
        self.httpClient = httpClient
    }

    public func handle(context: Lambda.Context, event: SNS.Event) -> EventLoopFuture<Void> {
        // ... do something using the httpClient and the database
        preconditionFailure()
    }
}

extension SSM.GetParametersResult {
    func value(for name: String) throws -> String {
        guard let value = parameters?.first(where: { $0.name == name })?.value else {
            throw ReviewImporterLambdaHandler.Error.missingParameterStoreValue(name: name, invalidParameters: invalidParameters)
        }

        return value
    }
}

extension Lambda {
    static func env(for name: String) throws -> String {
        guard let env = Lambda.env(name) else {
            throw ReviewImporterLambdaHandler.Error.missingEnvironmentVariable(name)
        }
        return env
    }
}
```

Excited about your thoughts. CC: @ktoso @yim-lee

tomerd:
hi @fabianfett if you want to keep this around, could you please update this PR to be against the `main` branch so we can delete the old `master` branch

PR #143, opened 2020-07-21
Remove section about static linking

jcavar:
Static linking doesn't work on Linux

### Motivation:

It is currently not possible to statically link binaries on Linux due to this bug:

https://bugs.swift.org/browse/SR-648

### Modifications:

README only.

### Result:

README only.
I am not sure if it is worth adding note about that to README?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -135,7 +135,7 @@ To deploy Lambda functions to AWS Lambda, you need to compile the code for Amazo
 
 AWS offers several tools to interact and deploy Lambda functions to AWS Lambda including [SAM](https://aws.amazon.com/serverless/sam/) and the [AWS CLI](https://aws.amazon.com/cli/). The [Examples Directory](/Examples) includes complete sample build and deployment scripts that utilize these tools.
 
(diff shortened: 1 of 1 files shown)


drexin:
@jcavar As noted in the README, static linking requires some additional linker flags to work, but it is not fundamentally broken. You can statically link an executable with SPM by telling the compiler directly to do it with `-Xswiftc -static-executable` instead of `--static-swift-stdlib`. In addition to that you may have to pass additional library dependencies to the linker with `-Xlinker`, e.g. when using Dispatch, you have to pass `-Xlinker -lDispatchStubs`. This is of course not optimal and we hope to have a better solution in the future, but it allows executables to be statically linked.

jcavar:
Oh, I see! Thank you very much.

pokryfka:
Static linking does work and improves performance, however the doc should probably be updated to include information how to do it:

- use `-static-executable` instead of `--static-swift-stdlib`
- in my setup I had also add `-lDispatchStubs` and `-latomic`

pokryfka:
@drexin 

what is `DispatchStubs`?
what does it "stub"?

PR #144, opened 2020-07-23
JSONEncoder/JSONDecoder and String convenience methods

fabianfett:
### Motivation:

Little common denominator after a long discussion in #129: `JSONEncoder` and `JSONDecoder` should have convenience methods to encode into `String` and decode from `String`.

### Result:

- It's a little easier to deal with json payloads that are delivered in a String. (APIGateway, SNS, SQS)
- Public API is increasing

### DIFF
--- a/Sources/AWSLambdaRuntime/Lambda+Codable.swift
+++ b/Sources/AWSLambdaRuntime/Lambda+Codable.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 @_exported import AWSLambdaRuntimeCore
+import struct Foundation.Data
 import class Foundation.JSONDecoder
 import class Foundation.JSONEncoder
 import NIO
@@ -128,5 +129,19 @@ extension JSONEncoder: LambdaCodableEncoder {
         var buffer = allocator.buffer(capacity: 1024)
         try self.encode(value, into: &buffer)
         return buffer
+    }
+}
+
+extension JSONEncoder {
+    /// Convenience method to allow encoding json directly into a String. It can be used to encode a payload into an APIGateway.V2.Response's body.
+    public func encodeAsString<T: Encodable>(_ value: T) throws -> String {
+        try String(decoding: self.encode(value), as: Unicode.UTF8.self)
+    }
+}
+
+extension JSONDecoder {
+    /// Convenience method to allow decoding json directly from a String. It can be used to decode a payload from an APIGateway.V2.Request's body.
+    public func decode<T: Decodable>(_ type: T.Type, from string: String) throws -> T {
+        try self.decode(type, from: Data(string.utf8))
     }
 }
--- a/Sources/AWSLambdaRuntime/Lambda+Codable.swift
+++ b/Sources/AWSLambdaRuntime/Lambda+Codable.swift
@@ -133,14 +133,14 @@ extension JSONEncoder: LambdaCodableEncoder {
 }
 
 extension JSONEncoder {
-    /// Convenience method to allow encoding json directly into a String. It can be used to encode a payload into an APIGateway.V2.Response's body.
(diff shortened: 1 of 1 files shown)


Sources/AWSLambdaRuntime/Lambda+Codable.swift
```diff
@@ -130,3 +131,17 @@ extension JSONEncoder: LambdaCodableEncoder {
         return buffer
     }
 }
+
+extension JSONEncoder {
+    /// Convenience method to allow encoding json directly into a String. It can be used to encode a payload into an APIGateway.V2.Response's body.
+    public func encodeAsString<T: Encodable>(_ value: T) throws -> String {
```
Roshanboy4u, line 137: can be refactored as
public func encodeAsString<Element: Encodable>(_ value: Element) throws -> String

Sources/AWSLambdaRuntime/Lambda+Codable.swift
```diff
@@ -130,3 +131,17 @@ extension JSONEncoder: LambdaCodableEncoder {
...
+        try String(decoding: self.encode(value), as: Unicode.UTF8.self)
+    }
+}
+
+extension JSONDecoder {
+    /// Convenience method to allow decoding json directly from a String. It can be used to decode a payload from an APIGateway.V2.Request's body.
+    public func decode<T: Decodable>(_ type: T.Type, from string: String) throws -> T {
```
Roshanboy4u, line 144: can be refactored as
public func decode<Element: Decodable>(_ type: Element.Type, from string: String) throws -> Element

Sources/AWSLambdaRuntime/Lambda+Codable.swift
```diff
@@ -130,3 +131,17 @@ extension JSONEncoder: LambdaCodableEncoder {
         return buffer
     }
 }
+
+extension JSONEncoder {
+    /// Convenience method to allow encoding json directly into a String. It can be used to encode a payload into an APIGateway.V2.Response's body.
```
ktoso, line 136:
```suggestion
    /// Convenience method to allow encoding json directly into a `String`. It can be used to encode a payload into an `APIGateway.V2.Response`'s body.
```

since it's an API type being called out?

Sources/AWSLambdaRuntime/Lambda+Codable.swift
```diff
@@ -130,3 +131,17 @@ extension JSONEncoder: LambdaCodableEncoder {
...
+    public func encodeAsString<T: Encodable>(_ value: T) throws -> String {
+        try String(decoding: self.encode(value), as: Unicode.UTF8.self)
+    }
+}
+
+extension JSONDecoder {
+    /// Convenience method to allow decoding json directly from a String. It can be used to decode a payload from an APIGateway.V2.Request's body.
```
ktoso, line 143:
```suggestion
    /// Convenience method to allow decoding json directly from a `String`. It can be used to decode a payload from an `APIGateway.V2.Request`'s body.
```

since it's names of types

ktoso:
I somewhat fail to see how this specific extensions are addressing the wider goal that was discussed in the linked issue #129 @fabianfett ? Tomer's comment https://github.com/swift-server/swift-aws-lambda-runtime/pull/129#issuecomment-652613419 outlines a somewhat wider effort than just the string coding functions?

ktoso:
Chatted some more; as I mentioned, not really strong opinion here so going to leave to @tomerd to decide if we're ok with these 👍

tomerd:
my only concern here is with extending public types from the SDK from which may conflict with other libraries if they do the same - we typically try to avoid doing so

@weissi wdyt?

fabianfett:
> my only concern here is with extending public types from the SDK from which may conflict with other libraries if they do the same - we typically try to avoid doing so

I don't want to steer the discussion in one direction and I'm sure you're aware, but let me mention for completeness, that we also extend the `JSONEncoder` and `JSONDecoder` for `ByteBuffer`.

PR #145, opened 2020-07-23
Refactoring

Roshanboy4u:
Improved design and structure.

### Motivation:

Request and Response are different concern which need to be addressed separately

### Modifications:

Refactored Request and Response

### Result:

It will help to reduce complexity and maintain good code structure

### DIFF
--- a/Examples/LocalDebugging/Shared/Sources/Shared/Request.swift
+++ b/Examples/LocalDebugging/Shared/Sources/Shared/Request.swift
@@ -1,0 +1,22 @@
+//
+//  Request.swift
+//  
+//
+//  Created by Roshan sah on 24/07/20.
+//
+
+import Foundation
+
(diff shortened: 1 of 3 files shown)


ktoso:
Hi @Roshanboy4u, thanks for the PR.

I'm not really convienced by the added value from the moving of the type here to be honest. Not sure if we'll want to merge this unless there's some other reasons and gains that I'm missing.

Roshanboy4u:
Hi @ktoso 
Correct me if I'm wrong. 
I tried to address separation of concerns. So in near future when it will grow, it can be addressed with ease.

ktoso:
I know what you mean but I'm not sure we want to accept this PR.
Only shuffling code around is very prone to one or the other author's opinion and thus could result in back and forth PRs etc.

Thanks for the effort but let's close this one. Thanks

PR #150, opened 2020-07-30
Make `run` return `Never`

helje5:
If an API method is designed to never return, it should be marked as returning `Never`. This way Swift will warn you if you add code after a call to such a method.


fabianfett:

helje5:
Note: Other `run` style methods might need this treatment (if deemed appropriate).

helje5:
Note that this is API semver major. Which probably triggers the errors. I'll turn the PR into an Issue.

tomerd:
thanks @helje5 we are not at 1.0 yet, so this should be fine imo 

@fabianfett wdyt?

helje5:
I've created https://github.com/swift-server/swift-aws-lambda-runtime/issues/151 to track this.

fabianfett:
@tomerd I think this would be a great addition and something we should get in before `1.0`

Issue #148, opened 2020-07-29
Local Server Immediate Exit code: 0 (Xcode 12 b3)


IrvTheSwirv:
Wondering if this is an Xcode 12 beta 3 issue:

Having had this issue with my own first attempt at a simple swift lambda I tried building and running the MyLambda example project here and had the same result. If I run the project without the LOCAL_LAMBDA_SERVER_ENABLED flag I get the expected:

`2020-07-29T12:22:40+0100 info Lambda : lambda lifecycle starting with Configuration
  General(logLevel: info))
  Lifecycle(id: 148308112161193, maxTimes: 0, stopSignal: TERM)
  RuntimeEngine(ip: 127.0.0.1, port: 7000, keepAlive: true, requestTimeout: nil
2020-07-29T12:22:40+0100 error Lambda : lifecycleIteration=0 could not fetch work from lambda runtime engine: badStatusCode(NIOHTTP1.HTTPResponseStatus.unprocessableEntity)
2020-07-29T12:22:40+0100 error Lambda : lifecycleIteration=0 lambda invocation sequence completed with error: badStatusCode(NIOHTTP1.HTTPResponseStatus.unprocessableEntity)
2020-07-29T12:22:40+0100 info Lambda : shutdown completed
Program ended with exit code: 0`

However, if I add the LOCAL_LAMBDA_SERVER_ENABLED flag and run, the process starts and finishes immediately and the only output is the following:

`Program ended with exit code: 0`

Some pointers on where to look to resolve this would be great.

IrvTheSwirv:
I've managed to step through the LocalServer code and found that the error is occurring in EvenLoopFuture in the wait() function (Line 928). The error is:

'(Result<(), Error>?) $R0 = failure {
  failure = {
    failureDescription = "bind(descriptor:ptr:bytes:)"
    error = errno (errno = 48)
  }
}'

fabianfett:
@IrvTheSwirv I hope this helps: https://fabianfett.de/getting-started-with-swift-aws-lambda-runtime#step-4-test-your-lambda-locally

IrvTheSwirv:
> @IrvTheSwirv I hope this helps: https://fabianfett.de/getting-started-with-swift-aws-lambda-runtime#step-4-test-your-lambda-locally

Hi. @fabianfett  Thanks but the problem happens when I add that flag. I managed to trace it to an error occurring inside the code that sets up the local server (see above) but I'm a little stuck beyond that.

[...]

PR #155, opened 2020-08-05
give better error when lambda fails to start

tomerd:
motivation: in error cases where lamnda fails to start the process fails silently and exists with code 0

changes: change the blocking API to check on the result and fatalError if lambda cannot be started / run


tomerd:
I am not 100% about this fix but probably better than what we have now. @fabianfett wdyt?

tomerd:
fixes #148

Issue #152, opened 2020-07-31
Remove requirement for macOS 10.13


helje5:
The runtime requires macOS 10.13, which is quite inconvenient because all consuming packages now _also_ need to explicitly require 10.13 in the package manifest. It is especially annoying if one tries to conditionally implement Lambda support in a 10.10 package via `#canImport`, e.g.: https://github.com/Macro-swift/Macro/blob/feature/lambda-1/Sources/http/Lambda/lambda.swift#L9

Slack says this requirement only exists for the ISO date formatter, not sure whether you use it for rendering only or also for parsing. In any case, it should be quite easy to replace w/ a custom parser/renderer. Should be reasonable because the format is fixed and never changes.

It could be built on top of `timegm` and `strptime`, though it needs an extra processing step to capture the milliseconds:
```swift
import func   Foundation.strptime
import func   Foundation.timegm
import struct Foundation.tm

let s          = "20180905T140903.591680Z"
var parsedTime = tm()

s.withCString { cstr in
  strptime(cstr, "%Y%m%dT%H%M%S.%fZ", &parsedTime) // this needs to extract the ms
}
let time = timegm(&parsedTime)
let ti   = TimeInterval(time) // + milliseconds
```

You could also provide this, and still use the ISO formatter if available (via `if #available(macOS 10.13, ...)`), which is what I do in Macro.

fabianfett:
I'm a fan of this and think that we should address it. @tomerd wdyt?

tomerd:
+1 conceptually - I dont like the 10.13 constraint either.

do we know  if this can be done in all formats coming form AWS? I think foundation used ICU for some timezone localization aspects which may be more complicated than the one demonstrated here, but it may not be used here

helje5:
? It’s ISO, there are only really two variants, with ms and without

helje5:
Oh, timezones, this format doesn’t support timezones (anything locale bound requiring tzdata) , max gmt offsets, though it probably is Z always

tomerd:
right, but IIRC there are other date formats we deal with in the library (unfortunately different AWS events have different date formats) so trying to understand if we can get away with this in all cases

helje5:
neither dateformatter nor locale require 10.13, it is just the specific isodateformatter, and iso is fixed

tomerd:
works for me!

fabianfett:
[...]

PR #156, opened 2020-08-05
remove requirment on macOS 10.13

tomerd:
motivation: simplify downstream libraries

changes: use custom ISO8601 decoders on macOS < 10.13

### DIFF
--- a/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
+++ b/Sources/AWSLambdaEvents/Utils/DateWrappers.swift
@@ -48,32 +48,14 @@ public struct ISO8601Coding: Decodable {
         if #available(macOS 10.12, *) {
             return Self.dateFormatter.date(from: string)
         } else {
-            return self.decodeISO8601Date(from: string).flatMap(Date.init(timeIntervalSince1970:))
+            // unlikely *debugging* use case of swift 5.2+ on older macOS
+            preconditionFailure("Unsporrted macOS version")
         }
         #endif
     }
 
     @available(macOS 10.12, *)
     private static let dateFormatter = ISO8601DateFormatter()
-
-    // strptime not avail on through Glibc
-    #if os(macOS)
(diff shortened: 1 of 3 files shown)


tomerd:
closes #152

tomerd:
@helje5

tomerd:
note `strptime` is not available via Libc so this is a bit uglier with #if os blocks

tomerd:
that sounds okay to me too. @helje5 wdyt?

helje5:
Either is fine for me. There are some Mac users which tend to stick to older versions, but they are probably not Swift target audience in the first place ;-)

helje5:
The really interesting question here is why spm decided to stick to 10.10 still, that feels a little weird, but the same reasoning should be applied i guess

tomerd:
@fabianfett what is your concern about taking this as-is? code bloat/complexity?

fabianfett:
> @fabianfett what is your concern about taking this as-is? code bloat/complexity?

@tomerd I came to the conclusion, when I thought about, how we could test this... Date conversions should be tested by a unit test imho. When thinking about it, I wasn’t sure if there is actually a machine that could fulfill the requirements to run this branch. If there is no machine that the code could be ever run on, we shouldn’t add it because it increases complexity.

tomerd:
@fabianfett addressed

Sources/AWSLambdaEvents/Utils/DateWrappers.swift
```diff
@@ -12,6 +12,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if os(Linux)
+import Glibc
+#else
+import Darwin.C
+#endif
+
```
fabianfett, line 20: @tomerd leftover?

tomerd: good catch, removed

PR #157, opened 2020-08-06
better error message when invoking local lambda server with incorrect http method

tomerd:
motivation: nicer error messages

changes: return 405 when using wrong http method on the lambda invocatoin endpoint (local debug server)


tomerd:
closes #146 

@fabianfett @helje5 this is a *minimal* fix to the most common issue. I agree with @fabianfett idea of addressing more broadly by separating the external facing server and the internal one

tomerd:
5.3 failure is due to swift-format, can ignore for now

PR #161, opened 2020-08-13
update swift-format to 0.45.6

tomerd:
motivation: swift-format is broken on 5.3

changes: update to 0.45.6 that is compatible with 5.3


fabianfett:
@tomerd Why has this been closed?

tomerd:
@fabianfett its not really helping with the 5.3 issue. I am waiting for 5.3 nighties to be available to see if this is an issue on master - which is what we are using on 5.2 CI rn

cc @mishalsha

PR #125, opened 2020-06-12
Add LambdaAPI Example of a Serverless REST API

Andrea-Scuderi:
# Motivation

Add an Example to show how to build a Serverless REST API using:
- APIGateway
- Lambda
- DynamoDB

PR #162, opened 2020-08-14
RFC: instrument lambda handler

pokryfka:
PoC instrumentation of the lambda handler.

### Motivation:

Add support for instrumentation.

### Modifications:

- creates an instance of a tracer (currently `AWSXRayRecorder`) - needs to be configurable
- pass tracer `TracingInstrument` and (context) baggage `BaggageContext` in `Lambda.Context`
- instrument invocation handling
- flushes the tracer after each invocation

### TODO

- [ ] import `swift-tracing` instead of `aws-xray-sdk-swift` to limit dependencies
  - [ ] stable `TracingInstrument` API
- [ ] extend `Lambda` bootstrap to be able to use the same `EventLoop` in an instance of `XRayUDPEmitter`
- [ ] discuss what exactly should be recorder by the handler

### Result:

Example in X-Ray console of handling a simple HelloWorld, cold start, Foundation JSON encoder/decoder.

Screen Shot 2020-08-14 at 11 08 02

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -21,6 +21,7 @@ import Darwin.C
 import Backtrace
 import Logging
 import NIO
+import NIOConcurrencyHelpers
 
 public enum Lambda {
     public typealias Handler = ByteBufferLambdaHandler
@@ -29,6 +30,23 @@ public enum Lambda {
     ///
     /// A function that takes a `InitializationContext` and returns an `EventLoopFuture` of a `ByteBufferLambdaHandler`
     public typealias HandlerFactory = (InitializationContext) -> EventLoopFuture<Handler>
+
+    internal typealias TracerFactory = (EventLoop) -> TracingInstrument
+
+    private static let lock = Lock()
+    private static var tracerFactory: TracerFactory = { _ in NoOpTracingInstrument() }
+
+    /// Select the desired `TracingInstrument` implementation.
+    ///
+    /// The caller is responsible for the tracer lifecycle.
+    ///
+    /// - parameters:
+    ///     -  tracerFactory: creates `TracingInstrument` implementation
+    public static func bootstrap(_ tracerFactory: @escaping (EventLoop) -> TracingInstrument) {
+        self.lock.withLockVoid {
+            self.tracerFactory = tracerFactory
+        }
+    }
 
     /// Run a Lambda defined by implementing the `LambdaHandler` protocol.
     ///
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -18,6 +18,7 @@ import Glibc
 import Darwin.C
 #endif
 
+import AWSXRaySDK
 import Backtrace
 import Logging
 import NIO
@@ -106,8 +107,12 @@ public enum Lambda {
             logger.logLevel = configuration.general.logLevel
 
             var result: Result<Int, Error>!
+            // TODO: bootstrap/configure the trace lifecycle
+            let tracer: XRayRecorder = XRayRecorder(eventLoopGroupProvider: .createNew)
             MultiThreadedEventLoopGroup.withCurrentThreadAsEventLoop { eventLoop in
-                let lifecycle = Lifecycle(eventLoop: eventLoop, logger: logger, configuration: configuration, factory: factory)
+//                let tracer = XRayRecorder(eventLoopGroupProvider: .shared(eventLoop))
+                let lifecycle = Lifecycle(eventLoop: eventLoop, logger: logger, tracer: tracer,
+                                          configuration: configuration, factory: factory)
                 #if DEBUG
                 let signalSource = trap(signal: configuration.lifecycle.stopSignal) { signal in
                     logger.info("intercepted signal: \(signal)")
@@ -127,6 +132,14 @@ public enum Lambda {
                         }
                     }
                     result = lifecycleResult
+                }
+            }
+
+            // TODO: this is broken, the tracer tries to eat cake and have cake: flash on eventLoop then syncShutdown
+            // fix in XRayRecorder
+            tracer.shutdown { error in
+                if let error = error {
+                    preconditionFailure("Failed to shutdown tracer: \(error)")
                 }
             }
 
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -21,7 +21,6 @@ import Darwin.C
 import Backtrace
 import Logging
 import NIO
-import NIOConcurrencyHelpers
 
 public enum Lambda {
     public typealias Handler = ByteBufferLambdaHandler
@@ -30,23 +29,6 @@ public enum Lambda {
     ///
     /// A function that takes a `InitializationContext` and returns an `EventLoopFuture` of a `ByteBufferLambdaHandler`
     public typealias HandlerFactory = (InitializationContext) -> EventLoopFuture<Handler>
-
-    internal typealias TracerFactory = (EventLoop) -> TracingInstrument
-
-    private static let lock = Lock()
-    private static var tracerFactory: TracerFactory = { _ in NoOpTracingInstrument() }
-
-    /// Select the desired `TracingInstrument` implementation.
-    ///
-    /// The caller is responsible for the tracer lifecycle.
-    ///
-    /// - parameters:
-    ///     -  tracerFactory: creates `TracingInstrument` implementation
-    public static func bootstrap(_ tracerFactory: @escaping (EventLoop) -> TracingInstrument) {
-        self.lock.withLockVoid {
-            self.tracerFactory = tracerFactory
-        }
-    }
 
     /// Run a Lambda defined by implementing the `LambdaHandler` protocol.
     ///
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import AWSXRayRecorder
 import Dispatch
 import Logging
 import NIO
@@ -68,10 +69,15 @@ extension Lambda {
         /// For invocations from the AWS Mobile SDK, data about the client application and device.
         public let clientContext: String?
 
+        public let baggage: XRayContext // TODO: use BaggageContext when swift-tracign is ready
+
         /// `Logger` to log with
         ///
         /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
         public let logger: Logger
+
+        /// Tracing instrument.
+        public let tracer: TracingInstrument
 
         /// The `EventLoop` the Lambda is executed on. Use this to schedule work with.
         /// This is useful when implementing the `EventLoopLambdaHandler` protocol.
@@ -91,8 +97,10 @@ extension Lambda {
                       cognitoIdentity: String? = nil,
                       clientContext: String? = nil,
                       logger: Logger,
+                      tracer: TracingInstrument,
                       eventLoop: EventLoop,
-                      allocator: ByteBufferAllocator) {
+                      allocator: ByteBufferAllocator)
+        {
             self.requestID = requestID
             self.traceID = traceID
             self.invokedFunctionARN = invokedFunctionARN
@@ -106,7 +114,10 @@ extension Lambda {
             var logger = logger
             logger[metadataKey: "awsRequestID"] = .string(requestID)
             logger[metadataKey: "awsTraceID"] = .string(traceID)
+            // TODO: handle error in better way
+            self.baggage = try! XRayContext(tracingHeader: traceID)
             self.logger = logger
+            self.tracer = tracer
         }
 
         public func getRemainingTime() -> TimeAmount {
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import AWSXRayRecorder
+import Baggage
 import Dispatch
 import Logging
 import NIO
@@ -69,7 +70,8 @@ extension Lambda {
         /// For invocations from the AWS Mobile SDK, data about the client application and device.
         public let clientContext: String?
 
-        public let baggage: XRayContext // TODO: use BaggageContext when swift-tracign is ready
+        /// Context baggage.
+        public var baggage: BaggageContext
 
         /// `Logger` to log with
         ///
@@ -114,8 +116,10 @@ extension Lambda {
             var logger = logger
             logger[metadataKey: "awsRequestID"] = .string(requestID)
             logger[metadataKey: "awsTraceID"] = .string(traceID)
-            // TODO: handle error in better way
-            self.baggage = try! XRayContext(tracingHeader: traceID)
+            var baggage = BaggageContext()
+            // TODO: handle error
+            baggage.xRayContext = try? XRayContext(tracingHeader: traceID)
+            self.baggage = baggage
             self.logger = logger
             self.tracer = tracer
         }
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -71,7 +71,7 @@ extension Lambda {
         public let clientContext: String?
 
         /// Context baggage.
-        public var baggage: BaggageContext
+        public let baggage: BaggageContext
 
         /// `Logger` to log with
         ///
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -117,7 +117,7 @@ extension Lambda {
             logger[metadataKey: "awsRequestID"] = .string(requestID)
             logger[metadataKey: "awsTraceID"] = .string(traceID)
             var baggage = BaggageContext()
-            // TODO: handle error
+            // TODO: use `swift-tracing` API, note that we can ONLY extract X-Ray Context from the invocation data
             baggage.xRayContext = try? XRayContext(tracingHeader: traceID)
             self.baggage = baggage
             self.logger = logger
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -17,6 +17,7 @@ import Baggage
 import Dispatch
 import Logging
 import NIO
+import NIOConcurrencyHelpers
 
 // MARK: - InitializationContext
 
@@ -52,6 +53,9 @@ extension Lambda {
     /// Lambda runtime context.
     /// The Lambda runtime generates and passes the `Context` to the Lambda handler as an argument.
     public final class Context: CustomDebugStringConvertible {
+        // TODO: use RWLock (separate PR)
+        private let lock = Lock()
+
         /// The request ID, which identifies the request that triggered the function invocation.
         public let requestID: String
 
@@ -70,8 +74,14 @@ extension Lambda {
         /// For invocations from the AWS Mobile SDK, data about the client application and device.
         public let clientContext: String?
 
-        /// Context baggage.
-        public let baggage: BaggageContext
+        // TODO: or should the Lambda "runtime" context and the Baggage context be separate?
+        private var _baggage: BaggageContext
+
+        /// Baggage context.
+        public var baggage: BaggageContext {
+            get { self.lock.withLock { _baggage } }
+            set { self.lock.withLockVoid { _baggage = newValue } }
+        }
 
         /// `Logger` to log with
         ///
@@ -117,9 +127,9 @@ extension Lambda {
             logger[metadataKey: "awsRequestID"] = .string(requestID)
             logger[metadataKey: "awsTraceID"] = .string(traceID)
             var baggage = BaggageContext()
-            // TODO: use `swift-tracing` API, note that we can ONLY extract X-Ray Context from the invocation data
+            // TODO: use `swift-tracing` API, note that, regardless, we can ONLY extract X-Ray Context
             baggage.xRayContext = try? XRayContext(tracingHeader: traceID)
-            self.baggage = baggage
+            self._baggage = baggage
             self.logger = logger
             self.tracer = tracer
         }
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -129,17 +129,17 @@ public protocol EventLoopLambdaHandler: ByteBufferLambdaHandler {
 public extension EventLoopLambdaHandler {
     /// Driver for `ByteBuffer` -> `In` decoding and `Out` -> `ByteBuffer` encoding
     func handle(context: Lambda.Context, event: ByteBuffer) -> EventLoopFuture<ByteBuffer?> {
-        // TODO: creating subsegments with NIO which will be easier if the baggage is passes in channel
-        // see https://github.com/slashmo/gsoc-swift-tracing/issues/48
         context.tracer.segment(name: "HandleEvent", context: context.baggage) {
-            // TODO: create helper to record errors in result types
-            let decodedEvent = context.tracer.segment(name: "DecodeIn", context: context.baggage) { _ in self.decodeIn(buffer: event) }
+            // TODO: create helper to record errors in passed in result types
+            let decodedEvent = context.tracer.segment(name: "DecodeIn", context: context.baggage) { _ in
+                self.decodeIn(buffer: event)
+            }
             switch decodedEvent {
             case .failure(let error):
                 return context.eventLoop.makeFailedFuture(CodecError.requestDecoding(error))
             case .success(let `in`):
                 return self.handle(context: context, event: `in`).flatMapThrowing { out in
-                    try context.tracer.segment(name: "encodeOut", context: context.baggage) { _ in
+                    try context.tracer.segment(name: "EncodeOut", context: context.baggage) { _ in
                         switch self.encodeOut(allocator: context.allocator, value: out) {
                         case .failure(let error):
                             throw CodecError.responseEncoding(error)
@@ -149,6 +149,9 @@ public extension EventLoopLambdaHandler {
                     }
                 }
             }
+        }.flatMap { result in
+            // flush the tracer after each invocation and return the invocation result
+            context.tracer.flush(on: context.eventLoop).map { result }
         }
     }
 
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -140,7 +140,8 @@ public extension EventLoopLambdaHandler {
             return context.eventLoop.makeFailedFuture(CodecError.requestDecoding(error))
         case .success(let `in`):
             let subsegment = segment.beginSubsegment(name: "HandleIn")
-            context.baggage = subsegment.baggage
+            // context is not thread safe, do not change it
+            // context.baggage = subsegment.baggage
             return self.handle(context: context, event: `in`)
                 .endSegment(subsegment)
                 .flatMapThrowing { out in
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -129,17 +129,29 @@ public protocol EventLoopLambdaHandler: ByteBufferLambdaHandler {
 public extension EventLoopLambdaHandler {
     /// Driver for `ByteBuffer` -> `In` decoding and `Out` -> `ByteBuffer` encoding
     func handle(context: Lambda.Context, event: ByteBuffer) -> EventLoopFuture<ByteBuffer?> {
-        context.tracer.segment(name: "HandleEvent", context: context.baggage) {
-            // TODO: create helper to record errors in passed in result types
-            let decodedEvent = context.tracer.segment(name: "DecodeIn", context: context.baggage) { _ in
-                self.decodeIn(buffer: event)
-            }
-            switch decodedEvent {
-            case .failure(let error):
-                return context.eventLoop.makeFailedFuture(CodecError.requestDecoding(error))
-            case .success(let `in`):
-                return self.handle(context: context, event: `in`).flatMapThrowing { out in
-                    try context.tracer.segment(name: "EncodeOut", context: context.baggage) { _ in
+        let segment = context.tracer.beginSegment(name: "HandleEvent", baggage: context.baggage)
+        // TODO: record errors propagated in result types?
+        let decodedEvent = segment.subsegment(name: "DecodeIn") { _ in
+            self.decodeIn(buffer: event)
+        }
+        switch decodedEvent {
+        case .failure(let error):
+            segment.addError(error)
+            segment.end()
+            return context.eventLoop.makeFailedFuture(CodecError.requestDecoding(error))
+        case .success(let `in`):
+            // TODO: use NIO helpers?
+            let subsegment = segment.beginSubsegment(name: "HandleIn")
+            context.baggage = subsegment.baggage
+            return self.handle(context: context, event: `in`)
+                .always { result in
+                    if case .failure(let error) = result {
+                        subsegment.addError(error)
+                    }
+                    subsegment.end()
+                }
+                .flatMapThrowing { out in
+                    try context.tracer.segment(name: "EncodeOut", baggage: segment.baggage) { _ in
                         switch self.encodeOut(allocator: context.allocator, value: out) {
                         case .failure(let error):
                             throw CodecError.responseEncoding(error)
@@ -147,11 +159,12 @@ public extension EventLoopLambdaHandler {
                             return buffer
                         }
                     }
+                }.always { result in
+                    if case .failure(let error) = result {
+                        segment.addError(error)
+                    }
+                    segment.end()
                 }
-            }
-        }.flatMap { result in
-            // flush the tracer after each invocation and return the invocation result
-            context.tracer.flush(on: context.eventLoop).map { result }
         }
     }
 
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -140,8 +140,7 @@ public extension EventLoopLambdaHandler {
             return context.eventLoop.makeFailedFuture(CodecError.requestDecoding(error))
         case .success(let `in`):
             let subsegment = segment.beginSubsegment(name: "HandleIn")
-            // TODO: context is not thread safe, do not change it
-            // context.baggage = subsegment.baggage
+            context.baggage = subsegment.baggage
             return self.handle(context: context, event: `in`)
                 .endSegment(subsegment)
                 .flatMapThrowing { out in
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -129,16 +129,24 @@ public protocol EventLoopLambdaHandler: ByteBufferLambdaHandler {
 public extension EventLoopLambdaHandler {
     /// Driver for `ByteBuffer` -> `In` decoding and `Out` -> `ByteBuffer` encoding
     func handle(context: Lambda.Context, event: ByteBuffer) -> EventLoopFuture<ByteBuffer?> {
-        switch self.decodeIn(buffer: event) {
-        case .failure(let error):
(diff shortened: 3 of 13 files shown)


pokryfka:
CC @ktoso @slashmo

ktoso:

ktoso:
Awesome 👏 😎 Will give it a look :-)

pokryfka:
@ktoso 

> Awesome 👏 😎 Will give it a look :-)

I have been using it for a week, obviously its PoC state, but I think PR may be a good way to discuss how it could be integrated including context propagation (note that Lambda defines its own `Context` which I used to pass the "context baggage" `BaggageContext` hmm...)

Attributes aside, I think it would be great to be able to share `EventLoop` to send UDP;
something `XRayRecorder` does support but is currently not exposed by `Lambda`, `InstrumentationSystem` (`bootstrap`) and `TracingInstrument` (`forceFlush`)

pokryfka:
> 14:23:44 Precondition failed: client not stopped before the deinit.: file /code/.build/checkouts/aws-xray-sdk-swift/Sources/AWSXRayUDPEmitter/UDPClient.swift, line 71

this breaks the tests at the moment, I was hoping `ServiceLifecycle` support will get merged https://github.com/swift-server/swift-aws-lambda-runtime/pull/141

slashmo:
Nice @pokryfka, thanks for taking the initiative! 🎉

tomerd:
thanks @pokryfka tagged this as RFC/draft for discussion

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -91,8 +99,10 @@ extension Lambda {
                       cognitoIdentity: String? = nil,
                       clientContext: String? = nil,
                       logger: Logger,
+                      tracer: TracingInstrument,
                       eventLoop: EventLoop,
-                      allocator: ByteBufferAllocator) {
+                      allocator: ByteBufferAllocator)
```
tomerd, line 104: @ktoso I think we have an abstraction that deals with combining the baggage and logger. could be appropriate here so that the logs include the tracing meta?

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -106,7 +116,12 @@ extension Lambda {
             var logger = logger
             logger[metadataKey: "awsRequestID"] = .string(requestID)
             logger[metadataKey: "awsTraceID"] = .string(traceID)
+            var baggage = BaggageContext()
+            // TODO: handle error
+            baggage.xRayContext = try? XRayContext(tracingHeader: traceID)
```
tomerd, line 121: would we want to let the users decide what tracing to use, or since this is AWS oriented anyways just pin to x-ray?

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -80,13 +89,18 @@ extension Lambda {
...
-                self.runtimeClient.reportResults(logger: logger, invocation: invocation, result: result).peekError { error in
-                    logger.error("could not report results to lambda runtime engine: \(error)")
+                self.tracer.segment(name: "ReportResults", baggage: baggage) { _ in
+                    self.runtimeClient.reportResults(logger: logger, invocation: invocation, result: result).peekError { error in
+                        logger.error("could not report results to lambda runtime engine: \(error)")
+                    }
                 }
+            }.flatMap {
+                // flush the tracer after each invocation
+                self.tracer.flush(on: self.eventLoop)
```
tomerd, line 103: should trace deeper in this method, e.g. the getNextInvocation call?

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -80,13 +89,18 @@ extension Lambda {
                         if case .failure(let error) = result {
                             logger.warning("lambda handler returned an error: \(error)")
                         }
-                        return (invocation, result)
+                        return (invocation, result, baggage)
                     }
-            }.flatMap { invocation, result in
+            }.flatMap { (invocation, result, baggage: BaggageContext) in
```
tomerd, line 94: I'd pass the context not just baggage. seems cleaner

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
tomerd: would this be `always` instead of `flatMap`?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -129,18 +129,42 @@ public protocol EventLoopLambdaHandler: ByteBufferLambdaHandler {
...
+                            throw CodecError.responseEncoding(error)
+                        case .success(let buffer):
+                            return buffer
+                        }
+                    }
+                }.always { result in
+                    if case .failure(let error) = result {
+                        segment.addError(error)
+                    }
+                    segment.end()
```
tomerd, line 166: I wonder if this pattern can be made generic with some extensions to SwiftNIO futures

ktoso:
>  (note that Lambda defines its own Context which I used to pass the "context baggage" BaggageContext hmm...)

Yeah, that's as-intended 👍 

The lambda context should conform to [LoggingBaggageContextCarrier](https://github.com/slashmo/gsoc-swift-baggage-context/blob/main/Sources/BaggageLogging/LoggingBaggageContextCarrier.swift) as well.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -68,11 +70,17 @@ extension Lambda {
         /// For invocations from the AWS Mobile SDK, data about the client application and device.
         public let clientContext: String?
 
+        /// Context baggage.
+        public var baggage: BaggageContext
```
ktoso, line 74: 👍

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -68,11 +70,17 @@ extension Lambda {
...
+
         /// `Logger` to log with
         ///
         /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
         public let logger: Logger
 
+        /// Tracing instrument.
+        public let tracer: TracingInstrument
```
ktoso, line 82: Not sure about the general pattern of storing a tracer, but also not really against it, let's see how it goes -- in lambda specifically I agree it can be nice to have the tracer just available like this, makes adding spans by users much easier 👍

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: Yes we have an adapter that handles the conversions etc, it's used like this:

```
import BaggageLogging

// let logger has to become:
public var logger: Logger {
  self._logger.with(context: self.baggage) // because we always want to use the "latest" value, if someone added some values there etc
}
```

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: I believe that's the main TODO in this PR - not using the xray explicitly.

Even if it is bound to xRay in reality, we should make sure to only use the abstract API, maybe maybe some day there would be some other tracer or maybe amazon decide to make their own or something, no idea, but let's keep the door open for future evolution.

Using the tracing API also means that while developing locally you could plug in the Instruments(.app) (naming gets confusing...) tracer: https://github.com/slashmo/gsoc-swift-tracing/pull/97 and see spans in Instruments on the mac. Instruments does not really understand / visualize "traces" with parents etc well today... but it's something we can keep in mind, maybe it'll get better at displaying those and then when developing locally you get the same user experience with tracing as on prod :-)

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: Oh actually, `Lambda.Context` is a class... so this may need lock protection

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -129,18 +129,42 @@ public protocol EventLoopLambdaHandler: ByteBufferLambdaHandler {
 public extension EventLoopLambdaHandler {
     /// Driver for `ByteBuffer` -> `In` decoding and `Out` -> `ByteBuffer` encoding
     func handle(context: Lambda.Context, event: ByteBuffer) -> EventLoopFuture<ByteBuffer?> {
-        switch self.decodeIn(buffer: event) {
+        let segment = context.tracer.beginSegment(name: "HandleEvent", baggage: context.baggage)
+        // TODO: record errors propagated in result types?
+        let decodedEvent = segment.subsegment(name: "DecodeIn") { _ in
```
ktoso, line 134: btw, I've been wondering if we should offer `tracer.withSpan() { ... }` as built-in for wrapping synchronous blocks with a span in the swift-tracing API right away, or if we should stay away from adding any kind of sugar in the API package.

This is quite common I think so I think we could add it...

We could also do the same with a NIO extensions package then to handle Future returning blocks 🤔 

WDYT @pokryfka @tomerd ?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -129,18 +129,42 @@ public protocol EventLoopLambdaHandler: ByteBufferLambdaHandler {
...
+            segment.end()
             return context.eventLoop.makeFailedFuture(CodecError.requestDecoding(error))
         case .success(let `in`):
-            return self.handle(context: context, event: `in`).flatMapThrowing { out in
-                switch self.encodeOut(allocator: context.allocator, value: out) {
-                case .failure(let error):
-                    throw CodecError.responseEncoding(error)
-                case .success(let buffer):
-                    return buffer
+            // TODO: use NIO helpers?
```
ktoso, line 143: Yeah, let's kick off designing those soon, ticketified: [Provide NIO Futures TracingSupport extensions package (not sure how to call the module yet) #125](https://github.com/slashmo/gsoc-swift-tracing/issues/125)

ktoso: @tomerd packaging wise, I'm thinking we should separate out all such support packages from the swift-tracing repo... like `swift-tracing-support` and there we'd add all such wrappers, also for URLSession and similar...? WDYT about separating it entirely on a repo level (not just separate module), so those can be versioned separately -- I suspect there can be much more "churn" on the "sugar" APIs than on the core stable boring API after all.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -129,18 +129,42 @@ public protocol EventLoopLambdaHandler: ByteBufferLambdaHandler {
...
+            context.baggage = subsegment.baggage
+            return self.handle(context: context, event: `in`)
+                .always { result in
+                    if case .failure(let error) = result {
+                        subsegment.addError(error)
+                    }
+                    subsegment.end()
+                }
+                .flatMapThrowing { out in
+                    try context.tracer.segment(name: "EncodeOut", baggage: segment.baggage) { _ in
```
ktoso, line 154: Naming nitpick: I really would like to get all libs to be consistent with the use of the `context` parameter name, note that it can then accept a carrier and it becomes easier to just pass the `context: segment` / `context: span` / `context: context (the lambda context); the only exception is `context.baggage` (like in `Lambda.Context`)

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
ktoso: Yup, let's kick off those "support" extension API discussions soon:  [Provide NIO Futures TracingSupport extensions package (not sure how to call the module yet) #125](https://github.com/slashmo/gsoc-swift-tracing/issues/125)

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -12,14 +12,19 @@
 //
 //===----------------------------------------------------------------------===//
 
+import AWSXRaySDK // TODO: use swift-tracing when available
+import Baggage
 import Dispatch
 import Logging
 import NIO
 
+public typealias TracingInstrument = XRayRecorder
```
ktoso, line 21: Ah I see, that confused me for a moment why it's an instrument but used segment() ;) 

Okey for the PoC, agreed on migrating once "released" though it would be good to use this PR to already migrate and see that the API supports everything one needs here.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -28,6 +33,7 @@ extension Lambda {
         init(eventLoop: EventLoop, configuration: Configuration) {
             self.eventLoop = eventLoop
             self.runtimeClient = RuntimeClient(eventLoop: self.eventLoop, configuration: configuration.runtimeEngine)
+            self.tracer = XRayRecorder(eventLoopGroupProvider: .shared(eventLoop))
```
ktoso, line 36: I see... wanting to share the EL can be good... We'll need to figure out how to allow this when using the abstract API. @tomerd I guess users don't really create the EL, the framework creates it right?

Since the tracing API can't depend on NIO, we can't solve it by initializers in there; we could solve it perhaps by lambda offering some `bootstrapTracer: EL -> TracingInstrument`? Or users have to create the event loop (group), then bootstrap their tracer, then run the lambda... a bit more ceremony I guess hm. Need to look into this more

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
fabianfett: @ktoso I generally agree with you. I'm however a little concerned, that tracing will require manual adjustment (incl. adding another dependency) if we don't include the XRay tracing by default. Lambda tracing wouldn't work out of the box in this case.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -68,11 +70,17 @@ extension Lambda {
         /// For invocations from the AWS Mobile SDK, data about the client application and device.
         public let clientContext: String?
 
+        /// Context baggage.
```
slashmo, line 73: I think calling it "Context baggage" could lead to confusion.

```suggestion
        /// Baggage context.
```

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
fabianfett: @ktoso +1

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
fabianfett: > I see... wanting to share the EL can be good... We'll need to figure out how to allow this when using the abstract API. @tomerd I guess users don't really create the EL, the framework creates it right?

YES

> Since the tracing API can't depend on NIO, we can't solve it by initializers in there; we could solve it perhaps by lambda offering some bootstrapTracer: EL -> TracingInstrument?

+1

> Or users have to create the event loop (group), then bootstrap their tracer, then run the lambda... a bit more ceremony I guess hm.

-1

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: Yeah I get that -- though the lambda runtime "core" should not be instrumented using any specific tracer, regardless of "yes it'll be xray" but maybe some day down the road there's other impls, and you'd want to swap it.

We could absolutely though make some "batteries included" package, we should think how to pull that off, wdyt?

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -117,9 +127,9 @@ extension Lambda {
             logger[metadataKey: "awsRequestID"] = .string(requestID)
             logger[metadataKey: "awsTraceID"] = .string(traceID)
             var baggage = BaggageContext()
-            // TODO: handle error
+            // TODO: use `swift-tracing` API, note that, regardless, we can ONLY extract X-Ray Context
```
ktoso, line 130: not really; "we" don't decide what we extract, the configuration of instruments decides that, and yes since xray would be configured it'd extract it's own context here.

Specifically: 

`tracer.extract(<from where to extract, could be a dictionary> / http headers etc, into: &baggage, using: extractor apropriate to the first parameter, so http headers or similar)`

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
pokryfka: I do use "helpers" both with closures and NIO futures as referenced here https://github.com/slashmo/gsoc-swift-tracing/issues/125#issuecomment-674356002 
when API is changed to `TracingInstrument` the amount of sugar will depend on what `swift-tracing` provides

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -52,6 +53,9 @@ extension Lambda {
     /// Lambda runtime context.
     /// The Lambda runtime generates and passes the `Context` to the Lambda handler as an argument.
     public final class Context: CustomDebugStringConvertible {
+        // TODO: use RWLock (separate PR)
+        private let lock = Lock()
```
ktoso, line 57: tbh we can start out with just a lock and change only if proven to matter a lot; 

The https://blog.nelhage.com/post/rwlock-contention/ keeps being brought up when we reach for RWLocks recently

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
pokryfka: this will be "context" when using `TracingInstrument` API:

```swift
var span = tracer.startSpan(named: "EncodeOut", context: segment.baggage) 
```

I dont have strong opinion on that, but to avoid confusion in XRaySDK I use "context" for `XRayContext` type (which does have the X-Ray trace context, strongly typed) and "baggage" for `BaggageContext` which *may* have the X-Ray trace context

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
pokryfka: this should make the API transition easier, temporary ...

ktoso: Would it be possible to PoC this out using the real swift-tracing API right away though?
We _need_ real feedback on the API by using it in _real_ use-cases, such as instrumenting lambda, http client and grpc; If there's things that show up as missing or sucky during that work now is the time to keep fixing the API. We should not wait for the API to move to its final destination to PoC things out and provide feedback. WDYT?

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
pokryfka: > should trace deeper in this method, e.g. the getNextInvocation call?

added recording of `getNextInvocation`, this is one is a bit tricky though:

1. we will get the X-Ray trace context in the `getNextInvocation` result
2. GET is blocking (one event at a time), without getting deeper inside of the HTTP client we dont know when data started (or ended) to be received
3. added simple instrumentation of the HTTPClient used by lambda runtime, however - because of the above - its only used when sending Result, at the moment (I am not sure if it adds much value that way)

> would this be always instead of flatMap?

correct, it should always flush (also when there were errors);
fixed and created a NIO convenience method `flush` to make it easier in other places

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
pokryfka: `BaggageContext` is value type, `Lambda.Context` is reference type

here we want to pass the "original" baggage value, while the value of `context.baggage` may have been changed by the time we get the handler result

maybe `Lambda.Context` should also be a struct or perhaps `BaggageContext` should be propagated separately?

@ktoso WDYT?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
pokryfka: updated the sugar in the XRaySDK API to make it cleaner and reusable

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
pokryfka: made it thread safe but I think it needs to be revisited https://github.com/swift-server/swift-aws-lambda-runtime/pull/162#discussion_r471457430

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
pokryfka: the problem here is that only the X-Ray trace context is provided by Lambda Runtime API (in header), see https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next

the context of other instruments may be provided in the payload BUT
the payload is not handled by the Lambda runtime but by (user provided) lambda event handler.

example:

the source event may be API Gateway which will contain HTTP headers.

However the `AWSLambdaRuntimeCore` requires only the event to be `Decodable` but the type itself is defined outside.
So lambda event handler (implemented by user) may parse the HTTP headers in the payload of API Gateway event
and extract context from them, but this will have to happen outside (done by user).

pokryfka: so to make it sorter:

`XRayInstrument` can use the `Instrument.extract` API (its implemented and tested)
but lambda runtime has only information about the X-Ray trace context.

Other context can be provided by payload and needs to be handled by user)

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
pokryfka: thank you, changed

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
pokryfka: @fabianfett I would really appreciate any suggestions how/where to configure/bootstrap tracer.

The challenge here is to pass the event loop which is currently not exposed by `InstrumentationSystem` not Lambda runtime.

`XRayRecorder` can create its own but its not optimal (low tier lambda configuration has just a fraction of vCPU, starting from 1/8 as far as I remember).

```swift
let instrument = XRayRecorder() // same as XRayRecorder(eventLoopGroupProvider: .createNew)
defer { instrument.shutdown() }
InstrumentationSystem.bootstrap(instrument)
```

this will most likely be affected (get easier?) by `ServiceLifecycle` integration https://github.com/swift-server/swift-aws-lambda-runtime/pull/141

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
ktoso: What would you say about the following spelling though:

```
var context = BaggageContext()
context.xRay =  XRayContext()
// since also:
// context.zipkin = ZipkinState() // I call it state in some things I worked on,
//
// because it aligns with  https://www.w3.org/TR/trace-context/#combined-header-value
// "tracestate" where each tracer may carry their own state by a vendor identified key
//
// Example: vendorname1=opaqueValue1,vendorname2=opaqueValue2
```

also because one can use the BaggageContextCarrier to assign _through_ into the underlying baggage context;
This way all use sites, regardless if a carrier, raw baggage context, or any "my specific framework type" can use the same call site style.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
pokryfka: This should be pretty much straight forward as far API is concerned.

I think the missing part is to freeze and release `TracingInstrument` API.
it will not be the final release but, if you consider existing API complete or a candidate for that, lets make a release and try to use it.

i am trying to keep `XRayInstrument` in sync but theres been many breaking changes recently in baggage and swift-tracing; I pin versions using git hashes, quite randomly.

Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -130,6 +135,14 @@ public enum Lambda {
                 }
             }
 
+            // TODO: this is broken, the tracer tries to eat cake and have cake: flash on eventLoop then syncShutdown
+            // fix in XRayRecorder
+            tracer.shutdown { error in
```
tomerd, line 140: maybe move this to L129

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
tomerd: RWlock are great when you do single (or very very very little) write and all-reads. in mixed mode it can get tricky tp get good performance

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
tomerd: so should we use it here?

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -12,22 +12,31 @@
...
     internal final class Runner {
         private let runtimeClient: RuntimeClient
+        private let tracer: TracingInstrument
         private let eventLoop: EventLoop
         private let allocator: ByteBufferAllocator
 
         private var isGettingNextInvocation = false
 
-        init(eventLoop: EventLoop, configuration: Configuration) {
+        init(eventLoop: EventLoop, configuration: Configuration, tracer: TracingInstrument = NoOpTracingInstrument()) {
```
tomerd, line 35: since this internal I prefer no default and just pipe the right things explicitly

Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift
```diff
@@ -58,7 +62,7 @@ extension Lambda {
         }
 
         /// Reports a result to the Runtime Engine.
-        func reportResults(logger: Logger, invocation: Invocation, result: Result<ByteBuffer?, Error>) -> EventLoopFuture<Void> {
+        func reportResults(logger: Logger, invocation: Invocation, result: Result<ByteBuffer?, Error>, context: BaggageContext) -> EventLoopFuture<Void> {
```
tomerd, line 65: in some places we refer to `BaggageContext` as `context` argument and in other places its referred to as `baggage`. I assume we all agree we need to be consistent. Maybe if not too late (and if makes sense) we should just rename the type to the short form `Baggage` and refer to it `baggage` since in most frameworks there will be some kind of `Context` that is the container/carrier. WDYT?

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -68,11 +74,23 @@ extension Lambda {
         /// For invocations from the AWS Mobile SDK, data about the client application and device.
         public let clientContext: String?
 
+        // TODO: or should the Lambda "runtime" context and the Baggage context be separate?
+        private var _baggage: BaggageContext
+
+        /// Baggage context.
+        public var baggage: BaggageContext {
```
tomerd, line 81: does this need to a `var` vs `let`? is it immutable or would the user ever write on this field?

Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift
tomerd: ah saw more discussion to this end on the above. looks like this is still a bit of area of confusion so maybe good feedback for the API and naming choices. Lets see if we can come with something that is intuitive otherwise we will end up having this discussion every time

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: I see, I missed that bit of the Lambda design. So in this integration style, even if one triggered the lambda via an http request, it would not get “the http request” but just the body, and the headers are the ones as listed on there, including the XRay trace header etc.

There AFAIR exists an integration mode though to get the entire request, right?

> Pass through the entire request – A Lambda function can receive the entire HTTP request (instead of just the request body) and set the HTTP response (instead of just the response body) using the AWS_PROXY integration type.

Is this something that the runtime currently is able to handle? Seems more to be about how the API Gateway is configured right? Though I’ve not had the time to dig deeper into this yet.

You’d probably know more about this @fabianfett, we can catch up about this today maybe?

pokryfka: I guess you are referring to [API Gateway Lambda proxy integration](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html)

This does affect Lambda custom runtime API, it affects API Gateway v1 "REST API" routing which, if configured that way, does not try to route events based on a RESTful model, instead it forwards all events to lambda which needs to resolve HTTP method, path and arguments itself:

> ANY /{proxy+}: The client must choose a particular HTTP method, must set a particular resource path hierarchy, and can set any headers, query string parameters, and applicable payload to pass the data as input to the integrated Lambda function.

Note that:
- events with HTTP requests created by API Gateway may have different syntax, specifically:
  - API Gateway v2 "HTTP API", see [AWSLambdaEvents/APIGateway+V2](https://github.com/swift-server/swift-aws-lambda-runtime/blob/main/Sources/AWSLambdaEvents/APIGateway%2BV2.swift)
  - API Gateway v1 "REST API", see [AWSLambdaEvents/APIGateway](https://github.com/swift-server/swift-aws-lambda-runtime/blob/main/Sources/AWSLambdaEvents/APIGateway.swift)
- event types are not defined in `AWSLambdaCore` but in `AWSLambdaEvents ` which, unlike `AWSLambdaCore`, does have dependency on `Foundation` at the moment.
- AWS does not limit context propagation to API Gateway (but it will not propagate context for other instruments unless it copies headers from the original requests and includes them in event payload, which is true for API Gateway)

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
pokryfka: @ktoso response from another thread

> The lambda context should conform to [LoggingBaggageContextCarrier](https://github.com/slashmo/gsoc-swift-baggage-context/blob/main/Sources/BaggageLogging/LoggingBaggageContextCarrier.swift) as well 

note that `LoggingBaggageContextCarrier` requires logger to be mutable, not just return a logger with the "latest" (baggage) context like in the snippet above

pokryfka: Also I think the logger should not use the "latest" but the "current" context.

This gets tricky if the (baggage) context is propagated in a reference type.

I tried to summarize it here https://github.com/slashmo/gsoc-swift-tracing/issues/126#issuecomment-675294268

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: It's currently forced to this via the Carrier's requirement, but it may well be that requirement is quite wrong and should just be `get` rather than `get/set` tbh. We'll look into that. It loops into a few pieces fo feedback @pokryfka had here 👀

Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift
ktoso: Yeah very much agreed... We wrote thought this through and there are documented guidelines here: https://github.com/slashmo/gsoc-swift-baggage-context#argument-namingpositioning

Specifically, it (and any other context) basically always should be context; a parameter should never be called baggage; the ONLY place there can be `baggage` is in a `BaggageContextCarrier` (which are always a "framework context") in order to avoid `context.context` it becomes `context.baggage` but since libraries should accept carriers rather than baggage itself most of the time you can always pass it like this `let context: Lambda.Context (is ...Carrier)` into `get(..., context: context)` since http client would accept `...Carrier` (the protocol, rather than the specific context type).

// I'm very open to finding better names. Will try to brainstorm this some more with feedback from this PR.

ktoso: Using _just_ the name Baggage we should think some more about if that's an option... That's going to be a standard name: https://w3c.github.io/baggage/ and it matches pretty much what we are TBH, a bag of pretty structured data.

BUT there is also https://www.w3.org/TR/trace-context/ and their relationship still remains messy (in the standards to be honest).

BaggageContext is a term known from PivotTracing and TracingPlane and somewhat "known" in the tracing space...

Open question really, but we need to be careful here :~

tomerd: personally, naming it just `Baggage` and the a framework context is a `BaggageCarrier` (cute) or `BaggageContainer` (technical) seems compelling but I may be missing use cases where this does not work?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
ktoso: I made a ticket to discuss naming once more: https://github.com/slashmo/gsoc-swift-baggage-context/issues/23

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: Thanks for explaining this in more depth @pokryfka, I need to read up some more here about aws/lambda in general it seems.

Sources/AWSLambdaRuntimeCore/Lambda.swift
pokryfka: @tomerd 

the problem is that the current implementation in `tracer.shutdown`:
- tries to flush first (on created or shared `EventLoop`),
- waits for it to finish,
- then shutdowns UDP client (which shares or owns the `EventLoop`).

this does not work well if lifecycle and tracer share the `EventLoop`:

```
Precondition failed: BUG DETECTED: wait() must not be called when on an EventLoop.
```

Not sure what the best solution is, I am thinking to remove flushing operation from shutdown,
that way client could always flush on the loop and then shutdown would just close all resources without trying to make any operation on the `EventLoop`.

@ktoso [`TracingInstrument`](https://github.com/slashmo/gsoc-swift-tracing/blob/fe80d764ad225b1dfd06dcb57d08b5e3485662f9/Sources/TracingInstrumentation/TracingInstrument.swift#L19) defined in `swift-tracing` only defines interface to sync flush.

```swift
public protocol TracingInstrument: Instrument {
  func forceFlush()
  // ...
}
```

for reference types in [`AWSXRayRecorder`](https://github.com/pokryfka/aws-xray-sdk-swift/blob/c505ac2d75ce838311365a40a7725cdfdfdcd2ec/Sources/AWSXRayRecorder/Recorder%2BNIO.swift#L17)

```swift
class XRayRecorder {
  // tries to flush firs
  public func shutdown(_ callback: ((Error?) -> Void)? = nil) { }
 // sync flush 
  public func wait(_ callback: ((Error?) -> Void)? = nil)
}

extension XRayRecorder {
    public func flush(on eventLoop: EventLoop) -> EventLoopFuture<Void> {}
}
```

ktoso: This is somewhat of a layering question.

swift-tracing cannot know about nio, so nothing about ELs or similar. 

We _could_ expose a `shutdown(callback)` if needed; question being, should it also flush, does flush need a callback as well then? Flushing is best-effort today and simply a signal to the tracer to try to flush.

The tracing instruments do not define shutdown because, similar like metrics, it is kind of assumed you're managing its lifecycle. We thought that normally you'd likely hook into `swift-lifecycle` with your tracer, and that'd be a specifc type there so you can do whatever you want.

Open to ideas here, what should we co in the API layer to help?
Would it help if we offered hooks that can interop with swift-lifecycle style callbacks nicely?

```swift
// pseudo code
let migrator = DatabaseMigrator()
lifecycle.register(
    label: "tracing",
    start: .sync { bootstrap tracer},
    shutdown: .async(shutdown tracer)
)
```

```swift
// this is again one of those examples which highlight the need for:
https://github.com/swift-server/swift-service-lifecycle/issues/11
// because we could express it as:
lifecycle.register(
    label: "tracing",
    start: .sync { 
         let tracer = MyTracer()
         Instrumentation.bootstrap(...)
         return tracer
    },
    shutdown: .async { tracer in tracer.shutdown(callback: $)) }
)
```

?

pokryfka: @ktoso 

> OR, do we need a shutdown(callback) on all instruments? 

this is related but could be handled outside of `TracingInstrument` API IMO

my problem is more with flushing:
- in my case (but I expect this is going to be common) I emit segments (spans) data using UDP client which uses SwiftNIO `EventLoop`
- I would like to share the `EventLoop` if possible, this is already implemented and works
- in case of lambda I need to flush after each invvocation
- `XRayRecoder` provides a method to flush on provided loop which allows to "async flush" but them hop in within invocation
- now, the problem is, `TracingInstrument` only defined sync `forceFlush` - which I provided as ~`flush(eventLoop).wait()`; this will work but not if EventLoop is shared

TL;DR I want to change API to `TracingInstrument` as soon as its release.
I cannot use its `forceFlush` if XRaySDK shares event loop with lambda runtime (which it should...)

ktoso: I'd argue forceFlush() is a _signal_ not a blocking function. don't wait on it, do your best, and your shutdown _can_ wait until flushes are complete. Wouldn't this solve the issue?

tomerd:
hi @pokryfka if you want to keep this around, could you please update this PR to be against the `main` branch so we can delete the old `master` branch

PR #167, opened 2020-09-16
[RFC] Adopts *Writable* BaggageContext (Better)

ktoso:
This adopts the baggage API in the shape of https://github.com/slashmo/gsoc-swift-baggage-context/pull/34

I have discussed with @fabianfett and @tomerd a little bit about this.

---

This adoption is _preferable_ I believe, because of comments that @adam-fowler raised in https://github.com/slashmo/gsoc-swift-baggage-context/pull/34#discussion_r488416689 In general, this means is also ready for the `Context` requiring a `set/get` on the baggage, making lifes easier for other folks.

Specifically, we talked with @fabianfett why the `Lambda.Context` context type was a class -- since it was immutable, it avoided copying around things a lot. We can achieve the same niceness by making it a CoW type, rather than just a struct. It's a bit of work but worth it in the long run as we get the best of both worlds.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -63,9 +63,7 @@ extension Lambda {
             let clientContext: String?
 
             // Implementation note: This logger is the "user provided logger" that we will log to when `context.logger` is used.
-            // However, we don't use it directly but through the `_logger.with(baggage)` pattern, in order to fulfil the `Context`'s
-            // contract about how the `logger` property must be implemented -- i.e. by always containing the latest baggage.
-            // This makes all log statements automatically include any baggage that is meaningful to include in logging.
+            // It must be updated with the latest metadata whenever the `baggage` changes.
             var _logger: Logger
 
             let eventLoop: EventLoop
@@ -77,7 +75,7 @@ extension Lambda {
                 deadline: DispatchWallTime,
                 cognitoIdentity: String?,
                 clientContext: String?,
-                _logger: Logger,
+                logger: Logger,
                 eventLoop: EventLoop,
                 allocator: ByteBufferAllocator
             ) {
@@ -86,7 +84,7 @@ extension Lambda {
                 self.deadline = deadline
                 self.cognitoIdentity = cognitoIdentity
                 self.clientContext = clientContext
-                self._logger = _logger
+                self._logger = logger
                 self.eventLoop = eventLoop
                 self.allocator = allocator
             }
@@ -100,15 +98,18 @@ extension Lambda {
             }
             set {
                 if isKnownUniquelyReferenced(&self.storage) {
+                    self.storage._logger.updateMetadata(previous: self.storage.baggage, latest: newValue)
                     self.storage.baggage = newValue
                 } else {
+                    var logger = self.storage._logger
+                    logger.updateMetadata(previous: self.storage.baggage, latest: newValue)
                     self.storage = Storage(
                         baggage: newValue,
                         invokedFunctionARN: self.storage.invokedFunctionARN,
                         deadline: self.storage.deadline,
                         cognitoIdentity: self.storage.cognitoIdentity,
                         clientContext: self.storage.clientContext,
-                        _logger: self.storage._logger,
+                        logger: self.storage._logger,
                         eventLoop: self.storage.eventLoop,
                         allocator: self.storage.allocator
                     )
@@ -151,10 +152,13 @@ extension Lambda {
         /// - note: The default `Logger.LogLevel` can be configured using the `LOG_LEVEL` environment variable.
         public var logger: Logger {
             get {
-                self.storage._logger.with(self.baggage)
+                self.storage._logger
             }
             set {
-                self.storage._logger = newValue
+                if isKnownUniquelyReferenced(&self.storage) {
+                    self.storage._logger = newValue
+                    self.storage._logger.updateMetadata(previous: .topLevel, latest: self.storage.baggage)
+                }
             }
         }
 
@@ -182,7 +186,7 @@ extension Lambda {
                       logger: Logger,
                       eventLoop: EventLoop,
                       allocator: ByteBufferAllocator) {
-            var baggage = Baggage.background
+            var baggage = Baggage.topLevel
             baggage.lambdaRequestID = requestID
             baggage.lambdaTraceID = traceID
             self.storage = Storage(
@@ -192,7 +196,7 @@ extension Lambda {
                 cognitoIdentity: cognitoIdentity,
                 clientContext: clientContext,
                 // utility
-                _logger: logger,
+                logger: logger,
                 eventLoop: eventLoop,
                 allocator: allocator
             )
@@ -239,6 +243,7 @@ extension Lambda {
 // MARK: - Baggage Items
 
 extension Baggage {
+
     // MARK: - Baggage: RequestID
 
     enum LambdaRequestIDKey: Key {
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -51,9 +51,10 @@ extension Lambda {
     /// Lambda runtime context.
     /// The Lambda runtime generates and passes the `Context` to the Lambda handler as an argument.
     public struct Context: BaggageContext.Context, CustomDebugStringConvertible {
-        private var storage: _Storage
+        /// Used to store all contents of the context and implement CoW semantics for it.
+        private var storage: Storage
 
-        final class _Storage {
+        final class Storage {
             var baggage: Baggage
 
             let invokedFunctionARN: String
@@ -61,6 +62,10 @@ extension Lambda {
             let cognitoIdentity: String?
             let clientContext: String?
 
+            // Implementation note: This logger is the "user provided logger" that we will log to when `context.logger` is used.
+            // However, we don't use it directly but through the `_logger.with(baggage)` pattern, in order to fulfil the `Context`'s
+            // contract about how the `logger` property must be implemented -- i.e. by always containing the latest baggage.
+            // This makes all log statements automatically include any baggage that is meaningful to include in logging.
             var _logger: Logger
 
             let eventLoop: EventLoop
@@ -97,7 +102,7 @@ extension Lambda {
                 if isKnownUniquelyReferenced(&self.storage) {
                     self.storage.baggage = newValue
                 } else {
-                    self.storage = _Storage(
+                    self.storage = Storage(
                         baggage: newValue,
                         invokedFunctionARN: self.storage.invokedFunctionARN,
                         deadline: self.storage.deadline,
@@ -143,9 +148,14 @@ extension Lambda {
 
         /// `Logger` to log with, it is automatically populated with `baggage` information (such as `traceID` and `requestID`).
         ///
-        /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
+        /// - note: The default `Logger.LogLevel` can be configured using the `LOG_LEVEL` environment variable.
         public var logger: Logger {
-            self.storage._logger.with(self.baggage)
+            get {
+                self.storage._logger.with(self.baggage)
+            }
+            set {
+                self.storage._logger = newValue
+            }
         }
 
         /// The `EventLoop` the Lambda is executed on. Use this to schedule work with.
@@ -175,7 +185,7 @@ extension Lambda {
             var baggage = Baggage.background
             baggage.lambdaRequestID = requestID
             baggage.lambdaTraceID = traceID
-            self.storage = _Storage(
+            self.storage = Storage(
                 baggage: baggage,
                 invokedFunctionARN: invokedFunctionARN,
                 deadline: deadline,
--- a/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaContext.swift
@@ -15,6 +15,7 @@
 import Dispatch
 import Logging
 import NIO
+import BaggageContext
 
 // MARK: - InitializationContext
 
@@ -49,12 +50,21 @@ extension Lambda {
 extension Lambda {
     /// Lambda runtime context.
     /// The Lambda runtime generates and passes the `Context` to the Lambda handler as an argument.
-    public final class Context: CustomDebugStringConvertible {
+    public final class Context: BaggageContext.Context, CustomDebugStringConvertible {
+
+        /// Contains contextual metadata such as request and trace identifiers, along with other information which may
+        /// be carried throughout asynchronous and cross-node boundaries (e.g. through HTTPClient calls).
+        public let baggage: Baggage
+
         /// The request ID, which identifies the request that triggered the function invocation.
-        public let requestID: String
+        public var requestID: String {
+            self.baggage.lambdaRequestID
+        }
 
         /// The AWS X-Ray tracing header.
-        public let traceID: String
+        public var traceID: String {
+            self.baggage.lambdaTraceID
+        }
 
         /// The ARN of the Lambda function, version, or alias that's specified in the invocation.
         public let invokedFunctionARN: String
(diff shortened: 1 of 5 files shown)


ktoso:
Whoop I PRed to master... fixing

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -49,40 +50,118 @@ extension Lambda {
...
+
+        final class _Storage {
+            var baggage: Baggage
+
+            let invokedFunctionARN: String
+            let deadline: DispatchWallTime
+            let cognitoIdentity: String?
+            let clientContext: String?
+
+            var _logger: Logger
```
fabianfett, line 64: Why has logger the underscore in its front?

fabianfett: Could this also be a `let`? I can't see anything that modifies it...

fabianfett (commented):
I really like this change – and it is totally in line with what I expected the Baggage change to look like! There are two questions about the use of `var _logger`?

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: Oh good point, yeah it should be modifiable if we go with the get+set Context style (which I do think is superior) along with the CoW contexts. 

Will fix thx!

ktoso: > Why has logger the underscore in its front?

Because the actual logger always has to be created on demand via:

```
var logger: Logger {
             self.storage._logger.with(self.baggage)
```

(that's the contract of `Context`), because then it always has the latest baggage.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -49,40 +50,118 @@ extension Lambda {
...
+    public struct Context: BaggageContext.Context, CustomDebugStringConvertible {
+        private var storage: _Storage
+
+        final class _Storage {
+            var baggage: Baggage
+
+            let invokedFunctionARN: String
+            let deadline: DispatchWallTime
+            let cognitoIdentity: String?
+            let clientContext: String?
```
ktoso, line 62: technically those could also become baggage items, your call really folks.

Only the trace stuff matters for instruments/tracers/logging, which is why I put it there.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
fabianfett: Okay, if you want to make this point explicit, I'd be in favor of a code comment on `Storage._logger`, explaining your thinking.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
fabianfett: If we have those values are present, we should add them to the baggage. Will associate the trace with a user. Though this could raise problems with privacy. Make this configurable?

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: Okey will do.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -49,40 +50,118 @@ extension Lambda {
...
     /// The Lambda runtime generates and passes the `Context` to the Lambda handler as an argument.
-    public final class Context: CustomDebugStringConvertible {
+    public struct Context: BaggageContext.Context, CustomDebugStringConvertible {
+        private var storage: _Storage
+
+        final class _Storage {
+            var baggage: Baggage
+
+            let invokedFunctionARN: String
+            let deadline: DispatchWallTime
```
ktoso, line 60: Deadlines we'll eventually be able to handle in a distributed way -- so this will eventually move into the baggage.

See: Future: Deadlines and Cancelation patterns https://github.com/slashmo/gsoc-swift-baggage-context/issues/24

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: I mean... client context is pretty special, it's actually JSON:

Screen Shot 2020-09-16 at 19 23 14

with some known magic values or "custom".

I don't think it's necessary in the baggage tbh...? Those values dont seem like they are important for tracing but just for this specific node / lambda?

https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html

Should the ClientContext be modeled using proper types?

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -93,20 +172,20 @@ extension Lambda {
...
+            self.storage = _Storage(
+                baggage: baggage,
+                invokedFunctionARN: invokedFunctionARN,
+                deadline: deadline,
+                cognitoIdentity: cognitoIdentity,
+                clientContext: clientContext,
+                // utility
+                _logger: logger,
+                eventLoop: eventLoop,
+                allocator: allocator
```
tomerd, line 187: iirc evenloop and allocator are immutable and dont need to be at the storage level imo

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -93,20 +172,20 @@ extension Lambda {
...
+            baggage.lambdaRequestID = requestID
+            baggage.lambdaTraceID = traceID
+            self.storage = _Storage(
+                baggage: baggage,
+                invokedFunctionARN: invokedFunctionARN,
+                deadline: deadline,
+                cognitoIdentity: cognitoIdentity,
+                clientContext: clientContext,
+                // utility
+                _logger: logger,
```
tomerd, line 185: not sure why we need underscore for logger or Storage

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -49,40 +50,118 @@ extension Lambda {
...
+            self.storage.clientContext
+        }
 
-        /// `Logger` to log with
+        /// `Logger` to log with, it is automatically populated with `baggage` information (such as `traceID` and `requestID`).
         ///
         /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
-        public let logger: Logger
+        public var logger: Logger {
+            self.storage._logger.with(self.baggage)
```
tomerd, line 148: I can't remember what with() does... does it mean every logger access is a new instance?

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
tomerd: or is this or allow them to be as part of the by-ref trick?

ktoso: Yeah this is to keep the outer Context type small -- just a single pointer, otherwise we'd grow the struct and each time it gets `let` or passed somewhere those references would be copied -- this way, everything in storage, nothing gets copied other than the single pointer just when passing the context around.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: Yes, every `context.logger` is a "new logger" but a logger is a cheap value type so that's not terrible.

This has to be like that because we need the logger to point to the "latest" baggage values. As everything here has value semantics, we can't just pass a reference to the context or baggage into the handler, and have to pass it when we're about to use it.

The `with` created LogHandler is lazy though -- it does not create Logger.Metadata _yet_ unless a log statement is actually triggered.

Alternatively, this could be done on a `baggage { didSet { ...` if you'd prefer that but I think that's actually hard to implement correctly without growing memory forever... we'd need to notice "aha, this logger already has a baggage logger, so we need to unwrap it and then wrap again with a baggage aware one" to prevent it growing infinitely with `Logger(BaggageLogHandler(BaggageLogHandler(BaggageLogHandler(BaggageLogHandler(BaggageLogHandler(BaggageLogHandler(...` if you know what I mean... This is hard to implement because `handler` is private in logging, so we can't check for it... Implementing it that way may require making the handler public in `SwiftLog`.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: For logger because it's "underlying logger, don't use this, use `logger` on context" for Storage happy to remove the underscore (done)

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
fabianfett: Oh boy, reading into the docu we should definitely fix that. But I think we should target another PR. I'm afraid though that it will be really hard to wrap our head around the `Custom` value here. I guess we will need something like `AnyCodable` for this.

fabianfett (approved):

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
ktoso: I mean, custom is simply a String... :)

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
tomerd: I can relate to doing so for the member, but not for the ctor argument

ktoso: ah i see, got it will do

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
tomerd: hmm I think this could actually add overhead for cases that log allot. I think keeping state on when baggage changed otherwise returning the existing logger could be a good optimization. maybe even coded into BaggageContext so users dont have to reinvent

ktoso: I revisited this and will try to implement via an "materialize once" into logging metadata. You're right that this is painful if there's a lot of logging, and we should not make that painful :\ 

It won't be doable as a log handler I guess, so that goes out the window; really I guess what would solve the needless copying is to make Logging depend on Baggage, and then we can remove its storage for metadata and replace it with baggage, avoiding the copying alltogether.

Not sure when to pitch this, it'd be a logging 2.0 I guess?

ktoso: Addressed this -- metadata is now set eagerly and we won't get into crazy copying anymore.

Thanks for reminding me about this, it's kind of obvious in hindsight but I missed it.

Some numbers how the copying "eagerly" amortizes (that's for 10k logs); it's just bad for "never logs" cases but one can assume those don't really happen in a large "actual" system (vs a library) tbh.

The `logger.with` lazy-pattern remains available when one wants to use it instead.

Screen Shot 2020-09-18 at 18 17 27

ktoso: (orange/red are the `logger.with(baggage)`)

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
pokryfka: please correct me if i am wrong, as far as I understand it:

1. everything (new storage object) is copied every time baggage is updated, which will happen every time new span is created = often

2. this does implement copy on write however changing of the storage is not thread safe which I think is not a nice API for user who would need to sync the changes himself

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -49,40 +50,132 @@ extension Lambda {
...
+        }
+
+        /// Contains contextual metadata such as request and trace identifiers, along with other information which may
+        /// be carried throughout asynchronous and cross-node boundaries (e.g. through HTTPClient calls).
+        public var baggage: Baggage {
+            get {
+                self.storage.baggage
+            }
+            set {
+                if isKnownUniquelyReferenced(&self.storage) {
```
pokryfka, line 100: please correct me if i am wrong, as far as I understand it:

1. everything (new storage object) is copied every time baggage is updated, which will happen every time new span is created = often

2. this does implement copy on write however changing of the storage is not thread safe which I think is not a nice API for user who would need to sync the changes himself

see discussion here https://gist.github.com/drewmccormack/b1c4487935cf3c3e0a5feaf488a95ebd#gistcomment-2826755

ktoso: Eh seems I have gotten it wrong here, could be.

With the latest changes in context perhaps actually we can get away without the Storage completely after all actually. I'll revisit shortly

ktoso: @pokryfka, so no. This is implementation correct. (You had me worried there for a second! 😉)

This is as thread safe as an `Int` or any other value semantics type is.

`isKnownUniquelyReferenced` is commonly misunderstood, so I understand where your confusion is coming from. However such implementation is safe by itself. Now how _someone_ uses this type may or may not be thread-safe -- and that is what the example you're linking is doing wrong. The same kind of race can be exhibited on even a plain integer value. This type provides value semantics guarantees, and no more.

--- 

> everything (new storage object) is copied every time baggage is updated, which will happen every time new span is created = often

I disagree about the "often" quantification used here. It is _by far_ not as often as a raw struct would be copying -- each passing around and reassigning to a new variable even would technically then be copying.

We're getting the best of both worlds; and simply following the known CoW semantics that are frequent in Swift.

> this does implement copy on write however changing of the storage is not thread safe which I think is not a nice API for user who would need to sync the changes himself

The operations in this piece of code _are_ safe. If they weren't then Array and all of Swift's collections are screwed as well 😉 We are providing the same semantics as those types.

pokryfka: > > everything (new storage object) is copied every time baggage is updated, which will happen every time new span is created = often
> 
> 
> 
> I disagree about the "often" quantification used here. It is _by far_ not as often as a raw struct would be copying -- each passing around and reassigning to a new variable even would technically then be copying.
> 
> 
> 
> We're getting the best of both worlds; and simply following the known CoW semantics that are frequent in Swift.
> 
> 

it used to be a "const reference" and so only the reference (pointer) was copied.

I dont know how expensive is copying of a small struct, and how much more expensive is allocating the space for it on heap rather than on stack; perhaps its not a real issue at all.

if the goal was to avoid copying of the lambda context, perhaps `Context` should have a reference to "static storage" and a reference to CoW "mutable storage" (the baggage). That way 2 references would always be copied but the content of the "static storage" would not need to be copied every time the baggage changes.


> 
> > this does implement copy on write however changing of the storage is not thread safe which I think is not a nice API for user who would need to sync the changes himself
> 
> 
>  
> The operations in this piece of code _are_ safe. If they weren't then Array and all of Swift's collections are screwed as well 😉 We are providing the same semantics as those types.

Thank you for rechecking it.
I do find it a bit confusing...

if I understand that it will be safe because `set` is `mutating`.

```swift
struct Context {

func notSafeUpdateBaggage(_ baggage: Baggage) {
 // update storage
}

// need to make a copy which will increase the storage reference counter
mutating func updateBaggage(_ baggage: Baggage) {
 // update storage
}

}
```

ktoso: > if the goal was to avoid copying of the lambda context, perhaps Context should have a reference to "static storage" and a reference to CoW "mutable storage" (the baggage). That way 2 references would always be copied but the content of the "static storage" would not need to be copied every time the baggage changes.

Yeah if it becomes a problem I think we could do that 🤔 👍 

> I do find it a bit confusing...
> 
> if I understand that it will be safe because set is mutating.

Maybe this example can illustrate: Because this is a struct, in order to change it the function must be mutating (or `set {}` which implicitly is).

Then if anyone has: 

```
let context = ...
```

they can't mutate it unless a second reference is created:

```
let context = ...
var changeMe = context
```

which ensures that we'll have the +1 count there and will copy on mutation.

It *still* is possible to make bugs like:

```
class Me { 
var onlyReference: Lambda.Context = ... 
for t in threads { 
  t.run { onlyReference.mutate() }
}
```

that remains illegal as usual, but not because of the CoW or Context type being wrong but how the user _uses_ it being unsafe. The same example is true for e.g. `var int: Int` which you'd try to mutate `+=1` from many threads. There's nothing wrong about the Int type, but about the unsafe modifications to where we store it.

tomerd:
closing in inactive PRs, feel free to re-open if still relevant

PR #168, opened 2020-09-16
[RFC] Adopts *readonly* BaggageContext (Meh for composition)

ktoso:
This is a "minimal changes" adoption of `Context`.

It is sub optimal I believe, and we should rather adopt: https://github.com/swift-server/swift-aws-lambda-runtime/pull/167

I post both so they can be compared. The CoW style implementation in #167 is much better and I believe we should adopt that, AND the `Context` library _should_ adopt get/set on the Context.baggage. See that issue for more details.


ktoso:
We decided readonly sucks and https://github.com/swift-server/swift-aws-lambda-runtime/pull/167 rocks.

PR #176, opened 2020-10-13
Properties of HTTPResponses are now var

fabianfett:
Properties of HTTPResponse types should be `var` not `let`

### Motivation:

When creating responses in Lambda developers might want to create the response in multiple steps changing and adding properties. As of now this is not possible since all HTTPResponse types only have `let` properties. If a developer wants to mark a response as immutable they have still the option to mark the complete response as `let`.

### Modifications:

- All properties of HTTPResponse types (`APIGateway.Response`, `APIGateway.V2.Response`, `ALB.TargetGroupResponse`) are now `var`iable
- An unsupported property has been removed from `APIGateway.V2.Response`. See [example/documentation at the very end](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)

### Result:

Developers have an easier job creating responses. 🥳

### DIFF
--- a/Sources/AWSLambdaEvents/APIGateway+V2.swift
+++ b/Sources/AWSLambdaEvents/APIGateway+V2.swift
@@ -93,24 +93,21 @@ extension APIGateway.V2 {
 
 extension APIGateway.V2 {
     public struct Response: Codable {
-        public let statusCode: HTTPResponseStatus
-        public let headers: HTTPHeaders?
-        public let multiValueHeaders: HTTPMultiValueHeaders?
-        public let body: String?
-        public let isBase64Encoded: Bool?
-        public let cookies: [String]?
+        public var statusCode: HTTPResponseStatus
+        public var headers: HTTPHeaders?
+        public var body: String?
+        public var isBase64Encoded: Bool?
+        public var cookies: [String]?
 
         public init(
             statusCode: HTTPResponseStatus,
             headers: HTTPHeaders? = nil,
-            multiValueHeaders: HTTPMultiValueHeaders? = nil,
             body: String? = nil,
             isBase64Encoded: Bool? = nil,
             cookies: [String]? = nil
         ) {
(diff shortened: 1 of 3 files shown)


ktoso (approved):
Looks good to me, chatted about this before a bit

Sources/AWSLambdaEvents/APIGateway+V2.swift
```diff
@@ -93,24 +93,21 @@ extension APIGateway.V2 {
...
+        public var statusCode: HTTPResponseStatus
+        public var headers: HTTPHeaders?
+        public var body: String?
+        public var isBase64Encoded: Bool?
+        public var cookies: [String]?
 
         public init(
             statusCode: HTTPResponseStatus,
             headers: HTTPHeaders? = nil,
-            multiValueHeaders: HTTPMultiValueHeaders? = nil,
```
tomerd, line 106: not needed any longer?

tomerd (commented):
looks good. one question re. change in API

Sources/AWSLambdaEvents/APIGateway+V2.swift
fabianfett: It never was supported in the first place. My mistake. Quote from [docu](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html):

> To customize the response, your Lambda function should return a response with the following format.hv
> 
>  ```json
> {
>     "cookies" : ["cookie1", "cookie2"],
>     "isBase64Encoded": true|false,
>     "statusCode": httpStatusCode,
>     "headers": { "headerName": "headerValue", ... },
>     "body": "Hello from Lambda!"
> }     
>  ```

tomerd: sounds good, need to point out in the release notes

Issue #166, opened 2020-09-13
APIGateway.V2.Request Authorizer jwt visibility


MoridinBG:
Currently the JWT struct on the Context's Authorizer is public, but the jwt instance is not. Is this a design decision?
Is there a better approach? Currently I am getting the token from the request `authorization` header and decoding manually to get claims.

I have several lambdas on an HTTP APIGateway API.

### SwiftAWSLambdaRuntime version/commit hash
0.2.0

### Swift & OS version (output of `swift --version && uname -a`)
Swift 5.2.4@MacOS 10.5/Swift 5.2.5@Amazonlinux2

fabianfett:
Thanks @MoridinBG, if you mean this line: 

https://github.com/swift-server/swift-aws-lambda-runtime/blob/f175b5452f8313791fded51c2ea9edf3740ea9f4/Sources/AWSLambdaEvents/APIGateway%2BV2.swift#L40

This has been a terrible oversight from my side. Would you be willing to provide a PR?

MoridinBG:
Gladly

PR #178, opened 2020-12-06
Fixed the visibility of the JWT in an API Gateway V2 Request.

JBosecker:
Made JWT in API Gateway V2 Request public.

### Motivation:

As noted in this issue
https://github.com/swift-server/swift-aws-lambda-runtime/issues/166
the JWT instance should be public.

### Modifications:

Added the public modifier.

### Result:

The JWT instance will be available for consuming modules.


ktoso:

ktoso (approved):
Looks good, thanks a lot for the fix!

PR #182, opened 2021-01-05
Fixing the queryStringParameters variable for ALB

proggeramlug:
Adjusted the `queryStringParameters` to be actually working with ALB.

### Motivation:

I want my Vapor apps to work through Lambda. However, this structure does not properly represent what AWS ALB is sending to the handler. There might be use-cases in which the original implementation is working but it looks fairly untested to me and I could not find the structure, as it is now, in any examples provided by AWS.

### Modifications:

Change `queryStringParameters` from `[String:[String]]` to `[String:String]`.

### Result:

It works with the actual ALB (live example can be provided if desired) and I did not come across any need for the original implementation.

### DIFF
--- a/Sources/AWSLambdaEvents/ALB.swift
+++ b/Sources/AWSLambdaEvents/ALB.swift
@@ -25,7 +25,7 @@ public enum ALB {
 
         public let httpMethod: HTTPMethod
         public let path: String
-        public let queryStringParameters: [String: [String]]
+        public let queryStringParameters: [String: String]
 
         /// Depending on your configuration of your target group either `headers` or `multiValueHeaders`
         /// are set.


fabianfett:

fabianfett (approved):
@proggeramlug thanks! I guess we should add the `multiValueQueryStringParameters` in a separate pr.

proggeramlug:
@fabianfett Do you have a specific use case for `multiValueQueryStringParameters`? I did not dig any deeper than my use-case required but I did not come across that specific parameter being passed.

fabianfett:
You can enable `multiValueHeaders` as [documented by AWS](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#enable-multi-value-headers). AFAIK if you enable `multiValueHeaders` the property be `multiValueQueryStringParameters` will be used instead of `queryStringParameters`. There is more documentation around [`multiValueHeaders`](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#multi-value-headers).

Of course the current implementation doesn't work with either of those. So it's definitely a fix that should go in!

PR #196, opened 2021-03-14
Update swiftformat

fabianfett:
### Motivation:

Our `swiftformat` should be more or less up to date. We should not require contributors to get a special version.

### Modifications:

- Update swiftformat from 0.44 to 0.47
- Bump swiftformat version to `5.2` (Since this is the swift version we require anyway)
- Add format option `--extensionacl on-declarations` to ensure ACL is never set on extensions
- Add format option `--disable typeSugar` to ensure `&&` is not replaced with `,`

### Result:

- Reduce the pain in contributing to `swift-aws-lambda-runtime`

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,6 +1,6 @@
 # file options
 
---swiftversion 5.1
+--swiftversion 5.2
 --exclude .build
 
 # format options
@@ -9,5 +9,10 @@
 --patternlet inline
 --stripunusedargs unnamed-only
 --ifdef no-indent
+--extensionacl on-declarations
+--disable typeSugar
+--disable andOperator
+--disable wrapMultilineStatementBraces
+--disable enumNamespaces
 
 # rules
--- a/Examples/LambdaFunctions/Sources/CurrencyExchange/main.swift
+++ b/Examples/LambdaFunctions/Sources/CurrencyExchange/main.swift
@@ -104,7 +104,8 @@ struct ExchangeRatesCalculator {
                           monthIndex: Array<Date>.Index,
                           currencies: [String],
                           state: [Date: ExchangeRates],
-                          callback: @escaping ((Result<[Date: ExchangeRates], Swift.Error>) -> Void)) {
+                          callback: @escaping ((Result<[Date: ExchangeRates], Swift.Error>) -> Void))
+    {
         if monthIndex == months.count {
             return callback(.success(state))
         }
@@ -145,16 +146,16 @@ struct ExchangeRatesCalculator {
         dateFormatter.dateFormat = "dd/MMM/yy"
         let interval: DateInterval?
         if let period = try document.nodes(forXPath: "/exchangeRateMonthList/@Period").first?.stringValue,
-            period.count == 26 {
+           period.count == 26 {
             // "01/Sep/2018 to 30/Sep/2018"
             let startString = period[period.startIndex ..< period.index(period.startIndex, offsetBy: 11)]
             let to = period[startString.endIndex ..< period.index(startString.endIndex, offsetBy: 4)]
             let endString = period[to.endIndex ..< period.index(to.endIndex, offsetBy: 11)]
             if let startDate = dateFormatter.date(from: String(startString)),
-                let startDay = calendar.dateInterval(of: .day, for: startDate),
-                to == " to ",
-                let endDate = dateFormatter.date(from: String(endString)),
-                let endDay = calendar.dateInterval(of: .day, for: endDate) {
+               let startDay = calendar.dateInterval(of: .day, for: startDate),
+               to == " to ",
+               let endDate = dateFormatter.date(from: String(endString)),
+               let endDay = calendar.dateInterval(of: .day, for: endDate) {
                 interval = DateInterval(start: startDay.start, end: endDay.end)
             } else {
                 interval = nil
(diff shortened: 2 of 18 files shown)


fabianfett:

Examples/LambdaFunctions/Sources/CurrencyExchange/main.swift
```diff
@@ -145,16 +146,18 @@ struct ExchangeRatesCalculator {
         dateFormatter.dateFormat = "dd/MMM/yy"
         let interval: DateInterval?
         if let period = try document.nodes(forXPath: "/exchangeRateMonthList/@Period").first?.stringValue,
-            period.count == 26 {
+           period.count == 26
+        {
```
ktoso, line 150: this is a bit weird?

Examples/LambdaFunctions/Sources/CurrencyExchange/main.swift
```diff
@@ -166,8 +169,9 @@ struct ExchangeRatesCalculator {
         let ratesByCurrencyCode: [String: Decimal?] = Dictionary(uniqueKeysWithValues: try currencyCodes.map {
             let xpathCurrency = $0.replacingOccurrences(of: "'", with: "&apos;")
             if let rateString = try document.nodes(forXPath: "/exchangeRateMonthList/exchangeRate/currencyCode[text()='\(xpathCurrency)']/../rateNew/text()").first?.stringValue,
-                // We must parse the decimal data using the UK locale, not the system one.
-                let rate = Decimal(string: rateString, locale: self.locale) {
+               // We must parse the decimal data using the UK locale, not the system one.
+               let rate = Decimal(string: rateString, locale: self.locale)
+            {
```
ktoso, line 174: yeah these are pretty weird... why does it cut down the {

Sources/AWSLambdaEvents/AppSync.swift
```diff
@@ -146,8 +146,8 @@ public enum AppSync {
     }
 }
 
-public extension AppSync {
-    enum Response<ResultType: Encodable>: Encodable {
+extension AppSync {
+    public enum Response<ResultType: Encodable>: Encodable {
```
ktoso, line 150: these are good actually 👍

Sources/AWSLambdaEvents/DynamoDB.swift
```diff
@@ -15,7 +15,7 @@
 import struct Foundation.Date
 
 // https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html
-public struct DynamoDB {
+public enum DynamoDB {
```
ktoso, line 18: that is honestly crazy... a formatter changing types just because it feels like it?

ktoso (commented):
Looks mostly fine, a few changes are a bit worrying

Sources/AWSLambdaEvents/DynamoDB.swift
tomerd: yea this is nasty imo - is there an option to turn this off?

Examples/LambdaFunctions/Sources/CurrencyExchange/main.swift
tomerd: yea this is odd, is there an option to turn this off?

Sources/AWSLambdaEvents/DynamoDB.swift
fabianfett: Actually I think this pretty clever... It detect's that this is used as a namespace and makes it explicit:

`enumNamespaces` -> Converts types used for hosting only static members into enums.
https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#enumnamespaces

tomerd (approved):

Sources/AWSLambdaEvents/DynamoDB.swift
ktoso: It's NOT up to a formatting tool to decide how I'm designing and using types.

We really should not allow this.

tomerd: I agree, its clever, but it should a human decision to change a type not a formatting tool

.swiftformat
```diff
@@ -9,5 +9,9 @@
 --patternlet inline
 --stripunusedargs unnamed-only
 --ifdef no-indent
+--extensionacl on-declarations
+--disable typeSugar
+--disable andOperator
+--disable wrapMultilineStatementBraces
```
ktoso, line 16: please disable the enum namespaces

ktoso (changes requested):

ktoso:
Ping @fabianfett

Sources/AWSLambdaEvents/DynamoDB.swift
fabianfett: I disabled the matching `swiftformat` rule. Since this change is actually clever in this case, do we want to keep it?

fabianfett:
@ktoso @tomerd I added `--disable enumNamespace`. Though I think we should keep the enum namespaces that were introduced by the former `swiftformat` run. wdyt?

Sources/AWSLambdaEvents/DynamoDB.swift
weissi: changing an `enum` to a `struct` is also API breaking...

ktoso:
Thanks! That sounds good to me :+1:

ktoso (approved):

Sources/AWSLambdaEvents/DynamoDB.swift
fabianfett: @weissi we go the other way though. struct -> enum. just out of curiosity: in which way would enum -> struct be source breaking, if it is just about a namespace?

weissi:
```
import enum Foo.Bar
```

will break. No tool should ever switch `enum` to or from `struct` (or class or something). The programmer needs to decide.

PR #199, opened 2021-04-11
Small Performance Improvements

fabianfett:
This PR has a number of smaller perf improvements. In my measuring they account for about a 3% perf improvement with tiny string payloads. The perf improvement should be much larger for larger invocation payloads.

### Modifications:

- Replaced `HTTPHandler` with `NIOHTTPClientResponseAggregator`, to fix a CoW issue. This will have by far the biggest speed impact. Especially for large payloads.
- Ensured that we only use one `EventLoop` in tests.
- Uses `syncOperations` to setup the `HTTPClient`.

### Result:

Faster Swift Lambda Runtime.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/HTTPClient.swift
+++ b/Sources/AWSLambdaRuntimeCore/HTTPClient.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftAWSLambdaRuntime open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftAWSLambdaRuntime project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -103,8 +103,7 @@ internal final class HTTPClient {
                     //   https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html
                     try channel.pipeline.syncOperations.addHandler(
                         NIOHTTPClientResponseAggregator(maxContentLength: 6 * 1024 * 1024))
-                    try channel.pipeline.syncOperations.addHandler(
-                        UnaryHandler(keepAlive: self.configuration.keepAlive))
+                    try channel.pipeline.syncOperations.addHandler(LambdaChannelHandler())
                     return channel.eventLoop.makeSucceededFuture(())
                 } catch {
                     return channel.eventLoop.makeFailedFuture(error)
@@ -139,10 +138,10 @@ internal final class HTTPClient {
     }
 
     internal struct Response: Equatable {
-        public var version: HTTPVersion
-        public var status: HTTPResponseStatus
-        public var headers: HTTPHeaders
-        public var body: ByteBuffer?
+        var version: HTTPVersion
+        var status: HTTPResponseStatus
+        var headers: HTTPHeaders
+        var body: ByteBuffer?
     }
 
     internal enum Errors: Error {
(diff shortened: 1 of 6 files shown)


Sources/AWSLambdaRuntimeCore/HTTPClient.swift
```diff
@@ -131,10 +138,10 @@ internal final class HTTPClient {
...
 
     internal struct Response: Equatable {
-        public var version: HTTPVersion
-        public var status: HTTPResponseStatus
-        public var headers: HTTPHeaders
-        public var body: ByteBuffer?
+        var version: HTTPVersion
+        var status: HTTPResponseStatus
+        var headers: HTTPHeaders
+        var body: ByteBuffer?
```
fabianfett, line 144: This is a cleanup change. `public` access control in a `internal struct` doesn't make much sense.

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
```diff
@@ -149,133 +156,75 @@ internal final class HTTPClient {
...
-                preconditionFailure("invalid read state \(self.readState)")
-            }
-            self.readState = .head(head)
-        case .body(var bodyPart):
-            switch self.readState {
-            case .head(let head):
-                self.readState = .body(head, bodyPart)
-            case .body(let head, var body):
-                body.writeBuffer(&bodyPart)
-                self.readState = .body(head, body)
```
fabianfett, line 217: We have a fairly large performance issue here... Since `body`'s internal storage is referenced from `self.readState` and from `body`, we incur a CoW on every append to the local buffer. We could fix this by an easy CoW prevention like:

```swift
case .body(let head, var body):
    self.readState = .idle // removes the self.readState reference. Therefore no Cow necessary
    body.writeBuffer(&bodyPart)
    self.readState = .body(head, body)
```

However since October of last year NIO has a `NIOHTTPClientResponseAggregator`. I think we should use that instead. All the http write logic from `HTTPHandler` has been moved into `UnaryHandler` directly.

Sources/AWSLambdaRuntimeCore/LambdaConfiguration.swift
```diff
@@ -64,7 +64,6 @@ extension Lambda {
         struct RuntimeEngine: CustomStringConvertible {
             let ip: String
             let port: Int
-            let keepAlive: Bool
```
fabianfett, line 67: I have removed the `keepAlive` option here, since neither tested nor used anywhere in the project. The HTTPClient should always prefer `keep-alive` but should be able to close a connection after a `connection: close` header. There is a test that ensures that the HTTPClient works correctly when a `connection: close` header is received.

Tests/AWSLambdaRuntimeCoreTests/MockLambdaServer.swift
```diff
@@ -30,7 +30,7 @@ internal final class MockLambdaServer {
     private var shutdown = false
 
     public init(behavior: LambdaServerBehavior, host: String = "127.0.0.1", port: Int = 7000, keepAlive: Bool = true) {
-        self.group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
```
fabianfett, line 33: This makes Xcodes debug view much easier to deal with! And we only need a single core for the MockServer anyway.

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
```diff
@@ -285,14 +236,19 @@ private final class UnaryHandler: ChannelDuplexHandler {
...
-        if !self.keepAlive || serverCloseConnection || response.version != .init(major: 1, minor: 1) {
-            pending.promise.futureResult.whenComplete { _ in
-                _ = context.channel.close()
-            }
+        let closeConnection = serverCloseConnection || response.head.version != .http1_1
+
+        if closeConnection {
+            self.state = .waitForConnectionClose(httpResponse, promise)
+            _ = context.channel.close()
+            return
```
fabianfett, line 247: This is a rather interesting workaround:

If we succeed the promise here and close the connection after succeeding the promise we may run into a race condition:

The lambda runtime will ask for the next work item directly after a succeeded post response request. The desire for the next work item might be faster than the attempt to close the connection. This will lead to a situation where we close the connection but the next request has already been scheduled on the connection that we want to close. For this reason we postpone succeeding the promise until the connection has been closed. This codepath will only be hit in the very, very unlikely event of the server demanding to close connection. (It's more or less only implemented to support http1.1 correctly.) Correct behavior is ensured with the test `testNoKeepAliveServer`.

Sources/AWSLambdaRuntimeCore/HTTPClient.swift
tomerd: moving to `NIOHTTPClientResponseAggregator` makes perfect sense. I wonder tho if moving all the extra functionality to `UnaryHandler` is overloading its intent - to be a unary handler. not sure the performance costs of having another handlers, if low, we can consider splitting that up.

Sources/AWSLambdaRuntimeCore/LambdaConfiguration.swift
tomerd: I think the intent was to be able and force closed connection by sending that header from the client, but this is fine.

tomerd (commented):
good improvement here. one dilemma regarding the expanded role of `UnaryHandler` to decide before we merge

tomerd (approved):
<3

PR #200, opened 2021-04-13
[RFC] Playing with ServiceLifecycle

fabianfett:
Supersedes #141

### Motivation:

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications:

_[Describe the modifications you've done.]_

### Result:

_[After your change, what will change.]_


tomerd:
one interesting venue to explore is how all this looks with async/await

fabianfett:
Superseded by #251

PR #201, opened 2021-04-13
Inline and specialize LambdaHandlers

fabianfett:
Add `@inlinable` to default LambdaHandler implementations. This ensures that those implementations can be inlined into user code and can be specialized for their input and output types.

### Motivation:

We like speed.

### Modifications:

- Add @inlinable to default protocol implementations

### Result:

In my measurement I see another ~4% performance improvement.

### DIFF
--- a/Sources/AWSLambdaRuntime/Lambda+Codable.swift
+++ b/Sources/AWSLambdaRuntime/Lambda+Codable.swift
@@ -80,6 +80,7 @@ internal struct CodableVoidClosureWrapper<In: Decodable>: LambdaHandler {
 
 /// Implementation of  a`ByteBuffer` to `In` decoding
 extension EventLoopLambdaHandler where In: Decodable {
+    @inlinable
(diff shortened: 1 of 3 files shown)


Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -25,18 +25,14 @@ import NIO
 public enum Lambda {
     public typealias Handler = ByteBufferLambdaHandler
 
-    /// `ByteBufferLambdaHandler` factory.
-    ///
-    /// A function that takes a `InitializationContext` and returns an `EventLoopFuture` of a `ByteBufferLambdaHandler`
-    public typealias HandlerFactory = (InitializationContext) -> EventLoopFuture<Handler>
```
tomerd, line 31: since this is public we should put a deprecation instead of deleting

tomerd (approved):
lgtm

PR #205, opened 2021-04-23
[RFC] State machine

fabianfett:
We like speed. 🚀 This pr drives to make Lambda faster by replacing EventLoopFuture chains with a StateMachine.

### Motivation:

- We like speed. 

### Modifications:

- Will be filled out later

### Result:

- Lambda allocates a lot less per invocation.


fabianfett:
This will land eventually in a non breaking way.

PR #216, opened 2021-07-30
split events into spearate package

tomerd:
motivation: the runtime library has a stabel API while the events are still moving target, so in order to provide a 1.0 stable version we should separate them out

changes:
* remove Events module
* update readme
* update Samples


Examples/LambdaFunctions/Package.swift
```diff
@@ -24,6 +24,9 @@ let package = Package(
         // in real-world projects this would say
         // .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", from: "1.0.0")
         .package(name: "swift-aws-lambda-runtime", path: "../.."),
+        // this is the dependency on the swift-aws-lambda-events library
+        // FIXME: update to tag when available
```
tomerd, line 28: 👀 we need to first tag https://github.com/swift-server/swift-aws-lambda-events

tomerd:
cc @bmoffatt @kneekey23

fabianfett (approved):
Looks good to me. However we should have an "integration" test target in AWSLambdaEvents that ensures that the examples in the runtime still compile after a change in events.

tomerd:
> Looks good to me. However we should have an "integration" test target in AWSLambdaEvents that ensures that the examples in the runtime still compile after a change in events.

@fabianfett you mean in the new repo?

fabianfett:
> @fabianfett you mean in the new "events" repo, so that changes there dont break the runtime? could they?

The changes in there can't break the runtime, but the samples in the runtime repository, which would block CI.

tomerd:
> The changes in there can't break the runtime, but the samples in the runtime repository, which would block CI.

it may ben easier (from a CI point of view) to add Examples there as well. WDYT?

tomerd:
https://github.com/swift-server/swift-aws-lambda-events/pull/1

fabianfett:
@tomerd Awesome! Now we should remove the `APIGateway` example from the code here. And maybe add to the soundness script that we never import `AWSLambdaEvents` except in markdown files :)

tomerd:
> tomerd Awesome! Now we should remove the APIGateway example from the code here. All other examples don't depend on AWSLambdaEvents. Maybe we should add to the soundness script that we never import AWSLambdaEvents except in markdown files :)

done

fabianfett:
Failing nightly-main CI because of known issue in SwiftNIO: https://github.com/apple/swift-nio/pull/1943 Not a blocker! Merging!

PR #218, opened 2021-08-13
Clean up dependency graph and imports

Lukasa:
Motivation:

AWSLambdaRuntimeCore and MockServer both have dependencies on the NIO
module, but they don't express this in their Package.swift: they
implicitly rely on these dependencies being met by the NIOHTTP1 import.
This was always brittle: it could lead to timing issues in the build
that could lead to it failing.

Modifications:

- Correctly express the dependency on NIO in Package.swift
- Correctly import _NIOConcurrency in files where it's used

Result:

Better and more accurate builds.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,6 +29,7 @@ let package = Package(
             .product(name: "Logging", package: "swift-log"),
             .product(name: "Backtrace", package: "swift-backtrace"),
             .product(name: "NIOHTTP1", package: "swift-nio"),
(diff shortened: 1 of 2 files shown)


fabianfett (approved):
Thanks for spotting and fixing this!

fabianfett:
Should we cherry pick this into a patch release? @tomerd @Lukasa

tomerd:
> Should we cherry pick this into a patch release? @tomerd @Lukasa

+1

PR #217, opened 2021-08-08
Change Lambda.Context from class to struct

jareyesda:
_Lambda.Context changed from a class to a struct._

### Motivation:

As described in issue #215. Lambda.Context is currently a class that is a wrapper around values. Because it should be a value type, it was changed from a class to a struct.

### Modifications:

The modifications are nearly identical to the changes from #167 - minus the addition of Baggage. The change from a class to a struct also follows CoW semantics to keep performance high. Line 104 of Sources/AWSLambdaRuntimeCore/LambdaRunner.swift had "convenience" and "file private" removed as well.

There are two differences between #167 and this modification: 1) At the time of forking, Tests/AWSLambdaRuntimeTests/Lambda+CodeableTest.swift was not present. 2) There is no Baggage

### Result:

Lambda.Context is now a CoW-boxed struct instead of a class.

### Note:

_This is my first contribution. Please advise on how I can improve future contributions and the proper methodology on testing this change. Thank you all in advance :)_

PR #219, opened 2021-08-18
InitializationContext should be a value not reference type

fabianfett:
Same treatment as #217 for `InitializationContext`. Since it is just a small struct, I didn't bother to create a CoW backing. Happy to do so, if wanted.


tomerd:
>  Since it is just a small struct, I didn't bother to create a CoW backing. Happy to do so, if wanted.

I think its only used once while initializing, so should not impact performance too much.

tomerd (approved):

fabianfett:

PR #220, opened 2021-08-18
Use explicit NIO imports

fabianfett:
With SwiftNIO 2.32.0, we now have explicit modules for generic NIO structures and the Posix implementations. This PR replaces all usages of `import NIO` with explicit import of `NIOCore` and `NIOPosix`.


tomerd (approved):

fabianfett:
Failing swift-main CI, because of an issue in SwiftNIO - not a blocker: https://github.com/apple/swift-nio/pull/1943

PR #222, opened 2021-08-19
Drop sync and closure APIs

### LINKED ISSUE #212: Remove callback based LambdaHandler
We should remove the Callback LambdaHandler.

fabianfett:
Fixes #212.

### Motivation:

For `1.0` we want to have an API that we are happy with going forward. Everything we are not 100% happy with should go out before `1.0`. With the release of Swift 5.5, we don't want to have callback based APIs anymore. For this reason we drop support for callback based APIs with this PR.

This also includes the way we deal with Lambda lifecycle: We drop the API's in which a user could create a Handler themselves and pass it to Lambda. The Handler creation is now always a part of the Lambda's lifecycle. For advanced users (users who wish to implement bridges to server frameworks) the advanced Lambda.Lifecycle API will stay.

### Modifications:

- drop callback based LambdaHandler
- rename AsyncLambdaHandler to LambdaHandler
- remove synchronous factory methods

### Result:

Smaller, cleaner API for 1.0.


Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -138,15 +138,15 @@ extension Lambda {
...
-        internal init(requestID: String,
-                      traceID: String,
-                      invokedFunctionARN: String,
-                      deadline: DispatchWallTime,
-                      cognitoIdentity: String? = nil,
-                      clientContext: String? = nil,
-                      logger: Logger,
-                      eventLoop: EventLoop,
-                      allocator: ByteBufferAllocator) {
+        public init(requestID: String,
```
fabianfett, line 141: @tomerd We probably need to discuss this one! Maybe we want to add a `__forTest_Only()` helper method instead of the public initializer.

Sources/AWSLambdaTesting/Lambda+Testing.swift
```diff
@@ -60,45 +64,25 @@ extension Lambda {
...
-        with event: In,
-        using config: TestConfig = .init()
-    ) throws -> Out {
-        try Self.test(CodableClosureWrapper(closure), with: event, using: config)
-    }
-
-    public static func test<In: Decodable>(
-        _ closure: @escaping Lambda.CodableVoidClosure<In>,
-        with event: In,
+    public static func test<Handler: LambdaHandler>(
```
fabianfett, line 67: @tomerd Lambda.test only supports new async/await lambdas:
```swift
XCTAssertNoThrow(result = try Lambda.test(MyLambda.self, with: input))
```
The benefit is that the Lambda goes through the complete lifecycle with this approach. Initializer and handler are called.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -36,7 +36,7 @@ extension Lambda {
         /// `ByteBufferAllocator` to allocate `ByteBuffer`
         public let allocator: ByteBufferAllocator
 
-        internal init(logger: Logger, eventLoop: EventLoop, allocator: ByteBufferAllocator) {
+        public init(logger: Logger, eventLoop: EventLoop, allocator: ByteBufferAllocator) {
```
fabianfett, line 39: @tomerd We probably need to discuss this one! Maybe we want to add a __forTest_Only() helper method instead of the public initializer.

Sources/AWSLambdaTesting/Lambda+Testing.swift
```diff
@@ -18,29 +18,33 @@
...
 //     }
 //
 //     let input = UUID().uuidString
 //     var result: String?
-//     XCTAssertNoThrow(result = try Lambda.test(MyLambda(), with: input))
+//     XCTAssertNoThrow(result = try Lambda.test(MyLambda.self, with: input))
 //     XCTAssertEqual(result, "echo" + input)
 // }
 
-#if DEBUG
```
tomerd, line 36: removal intentional?

Sources/CodableSample/main.swift
```diff
@@ -35,7 +35,7 @@ struct Handler: EventLoopLambdaHandler {
     }
 }
 
-Lambda.run(Handler())
+Lambda.run { $0.eventLoop.makeSucceededFuture(Handler()) }
```
tomerd, line 38: this is kind of meh

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
tomerd: yes this not ideal. why is it needed?

Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -78,27 +54,19 @@ public enum Lambda {
         return String(cString: value)
     }
 
+    #if swift(>=5.5)
```
tomerd, line 57: should we just require 5.5? feels wrong to drop support for closures *and* support < 5.5

fabianfett: We talked about this last time. We wanted to stay with `5.2` to allow existing frameworks (Vapor, Noze, Hummingbird, Smoke, and friends) to integrate with Lambda without needing to bump their requirements. For integration into those frameworks developers will always use the `Lambda.Lifecycle` APIs.

Normal Lambda users should require 5.5 and just use the async APIs.

I guess with this we will do what most of the Swift Server ecosystem is going todo... Make async/await the normal, leave escape hatch for advanced use cases using futures.

Sources/CodableSample/main.swift
fabianfett: Will readd with a separate PR.

Sources/AWSLambdaTesting/Lambda+Testing.swift
fabianfett: YES

tomerd (approved):

PR #224, opened 2021-08-24
Reorder Event and Context.

### LINKED ISSUE #89: Reorder context, future-proof for a tracing context world

fabianfett:
Fixes #89, for `ByteBufferLambdaHandler` and `EventLoopLambdaHandler` protocol.

In the new `LambdaHandler` protocol, the `event` already is in front of the `context`. Let's do the same for `ByteBufferLambdaHandler` and `EventLoopLambdaHandler`.

### DIFF
--- a/Examples/LambdaFunctions/Sources/Benchmark/main.swift
+++ b/Examples/LambdaFunctions/Sources/Benchmark/main.swift
@@ -25,7 +25,7 @@ struct BenchmarkHandler: EventLoopLambdaHandler {
     typealias In = String
     typealias Out = String
 
-    func handle(context: Lambda.Context, event: String) -> EventLoopFuture<String> {
+    func handle(event: String, context: Lambda.Context) -> EventLoopFuture<String> {
         context.eventLoop.makeSucceededFuture("hello, world!")
     }
 }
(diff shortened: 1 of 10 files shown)


tomerd (approved):

ktoso (commented):
Great to see it, thanks 👍

PR #225, opened 2021-08-25
[RFC] Drop event label from handle methods in LambdaHandlers

fabianfett:
Drop event label from handle methods in LambdaHandlers protocols

### Motivation:

In Lambda functions user's handle events from different sources. Those events can be `S3Event`s, but also `APIGatewayRequest`s. Let's look at a Lambda, that integrates with APIGateway.

```swift
@main
struct EchoLambda: AsyncLambdaHandler {

    init(context: Lambda.InitializationContext) async throws {}

    func handle(event: APIGatewayV2Request, context: Lambda.Context) async throws -> APIGatewayV2Response {
        APIGatewayV2Response(statusCode: .ok, body: event.body)
    }
}
```

The argument in the `handle` method is labeled `event` even though, in this case `request` would be clearly a better name. The user could use `request` as a parameter name by explicitly specifying the parameter name:

```swift
func handle(event request: APIGatewayV2Request, context: Lambda.Context) async throws -> APIGatewayV2Response
```

However `event request` is not easy to read and the parameter name overwrite is not the most known swift feature. For this reason I would advice against this option.

For this reason I propose dropping the argument label and specifying a parameter name `event` to guide users. New function signature:

```swift
func handle(_ event: In, context: Lambda.Context) async throws -> Out
```

This would allow users to write lambdas, that integrate with APIGateway as such:

```swift
func handle(_ request: APIGatewayV2Request, context: Lambda.Context) async throws -> APIGatewayV2Response
```

### Modifications:

- Drop the `event` label in the `handle` LambdaHandler protocol methods
- Rename typealias in `EventLoopHandler` protocol to `Event` and `Output` (from `In` and `Out`)

### Result:

A cleaner, more flexible API.

### DIFF
--- a/Examples/LambdaFunctions/Sources/Benchmark/main.swift
+++ b/Examples/LambdaFunctions/Sources/Benchmark/main.swift
@@ -22,8 +22,8 @@ import NIO
 Lambda.run { $0.eventLoop.makeSucceededFuture(BenchmarkHandler()) }
 
 struct BenchmarkHandler: EventLoopLambdaHandler {
-    typealias In = String
-    typealias Out = String
+    typealias Event = String
+    typealias Output = String
 
     func handle(_ event: String, context: Lambda.Context) -> EventLoopFuture<String> {
         context.eventLoop.makeSucceededFuture("hello, world!")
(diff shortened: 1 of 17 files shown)


adam-fowler:
This change makes sense. It'll cause some annoyance as it changes the signature of a protocol requirement but I can live with that.

fabianfett:

fabianfett:
@tomerd Updated the name of the event `In` type to `Event`. `Out` is now called `Output`. This should make adoption even easier. Please re-review.

tomerd (approved):
looks great!

tomerd:
some CI failures, seems like some test code needs to be further adjusted

PR #228, opened 2021-09-23
update and add examples to new APIs

tomerd:
motivation: keep code up to date with API changes

changes:
* take advantage of @main where possible
* move the top level Sample code to the examples subdirectory
* extract a few examples form the "LambdaFunctions" directory (which is really a deployment demo) and move them to the top level Examples directory
*  rename "LambdaFunctions" examples as "Deployments" to make their intent clearer
* add a sample that demonstrates how to test a lambda now that SwiftPM can test executables directly
* update the test-sample docker setup to build & test th new samples
* fix a few typos and In/Out typealias left overs
* remove LinuxMain since its no longer required


fabianfett:
What's the benefit of moving the String and Codable sample out of the main `Package.swift`?

tomerd:
> What's the benefit of moving the String and Codable sample out of the main Package.swift?

its much less noisy. these are really just samples and so they should not be part of the main package. One area where this causes headache is that tools that scan packages (like package collections and package indexes) and cannot tell that these modules are "not important" so they show up as valid modules you may want to use, which his not the intent

Examples/LambdaFunctions/Sources/Benchmark/main.swift
```diff
@@ -19,7 +19,6 @@ import NIO
 // use this example which is more performant.
 // `EventLoopLambdaHandler` does not offload the Lambda processing to a separate thread
 // while the closure-based handlers do.
```
stevapple, line 21: I think the description here should also be modified?

Examples/Simple/Testing/Sources/Lambda.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import AWSLambdaRuntime
+import NIOCore
```
stevapple, line 16: Is `NIOCore` really necessary here (and in other `LambdaHandler` examples)?

stevapple (commented):
Looks much neater!

Examples/Simple/String/Lambda.swift
```diff
@@ -13,17 +13,20 @@
 //===----------------------------------------------------------------------===//
 
 import AWSLambdaRuntimeCore
```
stevapple, line 15: Encourage using `AWSLambdaRuntime` here?

Examples/LambdaFunctions/Sources/Benchmark/main.swift
```diff
@@ -19,7 +19,6 @@ import NIO
 // use this example which is more performant.
```
stevapple, line 19: `NIOCore` is enough I think.

stevapple (commented):
We shall encourage `AWSLambdaRuntimeCore` with `NIOCore` with `EventLoopLambdaHandler`, and `AWSLambdaRuntime` with `LambdaHandler`?

fabianfett (approved):
Looks great. Much cleaner! Thanks!

PR #229, opened 2021-09-23
Remove dependency on _NIOConcurrency

fabianfett:
SwiftNIO 2.33.0 was just released. The `_NIOConcurrency` is now empty, since its functionality was moved to NIOCore. Remove the non API stable dependency.


tomerd (approved):

fabianfett:
Swift Nightly CI fails because of Sendable requirements. Not fixed in this pr.

PR #232, opened 2021-09-28
Rename `Lambda.Lifecycle` to `LambdaRuntime`

fabianfett:
Rename `Lambda.Lifecycle` to `LambdaRuntime`. As follow ups we should remove the namespacing on all public types and just plainly prefix them with Lambda.


tomerd (approved):
seems fine, too bad git thinks it a completely new file

PR #239, opened 2021-12-06
Add ControlPlaneRequestEncoder

fabianfett:
Add a new `ControlPlaneRequestEncoder` that encodes all requests into an existing buffer. 

### Motivation:

We want to write our request to the ControlPlaneAPI into one buffer, that we can reuse over and over. 

### Modifications:

- Add `ControlPlaneRequestEncoder` that writes `ControlPlaneRequest`s into a buffer that is reused.

### Result:

- Fewer messages on the ChannelPipeline.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/ControlPlaneRequestEncoder.swift
+++ b/Sources/AWSLambdaRuntimeCore/ControlPlaneRequestEncoder.swift
@@ -1,0 +1,126 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+struct ControlPlaneRequestEncoder: _EmittingChannelHandler {
+    typealias OutboundOut = ByteBuffer
+
+    private var host: String
+    private var byteBuffer: ByteBuffer!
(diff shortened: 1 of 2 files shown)


Sources/AWSLambdaRuntimeCore/ControlPlaneRequestEncoder.swift
```diff
@@ -0,0 +1,126 @@
...
+            context.flush()
+
+        case .invocationResponse(let requestID, let payload):
+            let contentLength = payload?.readableBytes ?? 0
+            self.byteBuffer.writeInvocationResultRequestLine(requestID)
+            self.byteBuffer.writeHostHeader(host: self.host)
+            self.byteBuffer.writeString(.userAgentHeader)
+            self.byteBuffer.writeContentLengthHeader(length: contentLength)
+            self.byteBuffer.writeString(.CRLF) // end of head
+            if contentLength > 0 {
```
tomerd, line 46: `if let payload = payload && contentLength > 0` would be nice so we can drop the force unwrap

Sources/AWSLambdaRuntimeCore/ControlPlaneRequestEncoder.swift
```diff
@@ -0,0 +1,126 @@
...
+            self.byteBuffer.writeString(.CRLF) // end of head
+            self.byteBuffer.writeBytes(payload)
+            context.write(self.wrapOutboundOut(self.byteBuffer), promise: promise)
+            context.flush()
+        }
+    }
+
+    mutating func writerAdded(context: ChannelHandlerContext) {
+        self.byteBuffer = context.channel.allocator.buffer(capacity: 256)
+    }
```
tomerd, line 82: nice

tomerd (approved):
😎

Sources/AWSLambdaRuntimeCore/ControlPlaneRequestEncoder.swift
fabianfett: Fixed.

PR #243, opened 2021-12-10
Add LambdaRequestID

fabianfett:
If we want to minimize allocations for every invocation, we need to look at types that currently allocate. Currently we use a String to hold the request id. However since the request id is a uuid, that string is 36 characters long. This is way above the 15 character string allocation threshold. The go to type in this case would be `UUID`. However `UUID` is in Foundation and we want to keep the lambda runtime Foundation free.

This pr introduces a LambdaRequestID to represent a uuid. One nice side effect of having our own uuid case is: We can make writing the uuid-string to a ByteBuffer allocation free (since no intermediate translation to a string is needed first).

This pr does not add any public API.


tomerd (approved):

tomerd:
lgtm 👍🏼

PR #244, opened 2022-01-03
Lambda factory as a protocol requirement.

fabianfett:
As previously discussed...

We want to use the same lambda application lifecycle for all lambda implementations. Currently we support `@main` only for the `LambdaHandler` protocol. This PR brings the `@main` functionality to the lower level protocols.

### Modifications:

- Add a factory method requirement to the `ByteBufferLambdaHandler` protocol
- Remove the factory callback (`HandlerFactory`) from the `LambdaRuntime` and `Lambda`
- Make the `LambdaRuntime` generic over the `LambdaHandler` type
- Adjust tests
- Write better docu

### Result:

- The application lifecycle is the same for all protocol levels.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -34,13 +34,17 @@ public enum Lambda {
 
     /// Run a Lambda defined by implementing the ``ByteBufferLambdaHandler`` protocol.
     /// The Runtime will manage the Lambdas application lifecycle automatically. It will invoke the
-    /// ``ByteBufferLambdaHandler/factory(context:)`` to create a new Handler.
+    /// ``ByteBufferLambdaHandler/makeHandler(context:)`` to create a new Handler.
     ///
     /// - parameters:
-    ///     - factory: A `ByteBufferLambdaHandler` factory.
+    ///     - configuration: A Lambda runtime configuration object
+    ///     - handlerType: The Handler to create and invoke.
     ///
     /// - note: This is a blocking operation that will run forever, as its lifecycle is managed by the AWS Lambda Runtime Engine.
-    internal static func run<Handler: ByteBufferLambdaHandler>(configuration: Configuration = .init(), handlerType: Handler.Type) -> Result<Int, Error> {
+    internal static func run<Handler: ByteBufferLambdaHandler>(
+        configuration: Configuration = .init(),
+        handlerType: Handler.Type
+    ) -> Result<Int, Error> {
         let _run = { (configuration: Configuration) -> Result<Int, Error> in
             Backtrace.install()
             var logger = Logger(label: "Lambda")
@@ -89,7 +93,7 @@ public enum Lambda {
             return _run(configuration)
         }
         #else
-        return _run(configuration, factory)
+        return _run(configuration)
         #endif
     }
 }
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -24,27 +24,6 @@ import NIOCore
 import NIOPosix
 
 public enum Lambda {
-    public typealias Handler = ByteBufferLambdaHandler
-
-    /// `ByteBufferLambdaHandler` factory.
-    ///
-    /// A function that takes a `InitializationContext` and returns an `EventLoopFuture` of a `ByteBufferLambdaHandler`
-    public typealias HandlerFactory = (InitializationContext) -> EventLoopFuture<Handler>
-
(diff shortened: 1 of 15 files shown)


Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -53,30 +32,23 @@ public enum Lambda {
...
-    internal static func run(configuration: Configuration = .init(), factory: @escaping HandlerFactory) -> Result<Int, Error> {
-        let _run = { (configuration: Configuration, factory: @escaping HandlerFactory) -> Result<Int, Error> in
+    /// Run a Lambda defined by implementing the ``ByteBufferLambdaHandler`` protocol.
+    /// The Runtime will manage the Lambdas application lifecycle automatically. It will invoke the
+    /// ``ByteBufferLambdaHandler/factory(context:)`` to create a new Handler.
+    ///
+    /// - parameters:
+    ///     - factory: A `ByteBufferLambdaHandler` factory.
+    ///
+    /// - note: This is a blocking operation that will run forever, as its lifecycle is managed by the AWS Lambda Runtime Engine.
```
tomerd, line 42: update docs ^^

Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -53,30 +32,23 @@ public enum Lambda {
...
+    internal static func run<Handler: ByteBufferLambdaHandler>(configuration: Configuration = .init(), handlerType: Handler.Type) -> Result<Int, Error> {
```
tomerd, line 43: is `handlerType` used?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -135,9 +154,18 @@ extension EventLoopLambdaHandler where Output == Void {
...
+///         ``LambdaHandler`` based APIs.
 ///         Most users are not expected to use this protocol.
 public protocol ByteBufferLambdaHandler {
+    /// Create your Lambda handler for the runtime.
+    ///
+    /// Use this to initialize all your resources that you want to cache between invocations. This could be database
+    /// connections and HTTP clients for example. It is encouraged to use the given `EventLoop`'s conformance
+    /// to `EventLoopGroup` when initializing NIO dependencies. This will improve overall performance, as it
+    /// minimizes thread hopping.
+    static func factory(context: Lambda.InitializationContext) -> EventLoopFuture<Self>
```
tomerd, line 167: maybe `makeHandler`? factory is generally not popular term in Swift. cc @weissi for ideas

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -163,6 +191,20 @@ extension ByteBufferLambdaHandler {
...
+    /// Initializes and runs the lambda function.
+    ///
+    /// If you precede your ``ByteBufferLambdaHandler`` conformer's declaration with the
+    /// [@main](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#ID626)
+    /// attribute, the system calls the conformer's `main()` method to launch the lambda function.
+    ///
+    /// The lambda runtime provides a default implementation of the method that manages the launch
+    /// process.
+    public static func main() {
+        _ = Lambda.run(configuration: .init(), handlerType: Self.self)
```
tomerd, line 204: should we mark Lambda::run @discardableResult?

Sources/AWSLambdaRuntimeCore/Lambda.swift
tomerd: I guess needed for the generics?

tomerd (approved):
nice! some small suggestions

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
ktoso: Yeah, generally "make" is the word to use; `makeHandler` would work too.

Sources/AWSLambdaRuntimeCore/Lambda.swift
fabianfett: Yes. It is there to make the user declare the Generic type. Otherwise the user would need to write:
```swift
Lambda.run<Handler>()
```

Sources/AWSLambdaRuntimeCore/Lambda.swift
fabianfett: fixed

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
fabianfett: I think the `Lambda.run` method should not return anything. But I haven't made my mind up about this fully. Would like to tackle later.

PR #246, opened 2022-02-10
Add default value for traceID header

StefanNienhuis:
Adds a default value for the Lambda-Runtime-Trace-Id header, to make it compatible with the AWS SAM CLI and other applications depending on the Lambda Runtime Interface Emulator. 

### Motivation:

Currently the Swift Runtime is incompatible with the Lambda Runtime Interface Emulator.

### Modifications:

Adds a default value for the Lambda-Runtime-Trace-Id header.

This is similar to what the [Rust runtime](https://github.com/awslabs/aws-lambda-rust-runtime/blob/f33fa059a88869b38646d875688fab58885208de/lambda-runtime/src/types.rs#L157) is doing.

### Result:

The Swift Runtime will be compatible with the Lambda Runtime Interface Emulator.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
+++ b/Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
@@ -54,7 +54,8 @@ struct Invocation: Hashable {
         self.requestID = requestID
         self.deadlineInMillisSinceEpoch = unixTimeInMilliseconds
         self.invokedFunctionARN = invokedFunctionARN
-        self.traceID = headers.first(name: AmazonHeaders.traceID) ?? ""
+        self.traceID = headers.first(name: AmazonHeaders.traceID) ?? 
+            "Root=\(AmazonHeaders.generateXRayTraceID());Sampled=0"
         self.clientContext = headers["Lambda-Runtime-Client-Context"].first
         self.cognitoIdentity = headers["Lambda-Runtime-Cognito-Identity"].first
     }
--- a/Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
+++ b/Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
@@ -51,14 +51,10 @@ struct Invocation: Hashable {
             throw Lambda.RuntimeError.invocationMissingHeader(AmazonHeaders.invokedFunctionARN)
         }
 
-        guard let traceID = headers.first(name: AmazonHeaders.traceID) else {
-            throw Lambda.RuntimeError.invocationMissingHeader(AmazonHeaders.traceID)
-        }
-
         self.requestID = requestID
         self.deadlineInMillisSinceEpoch = unixTimeInMilliseconds
         self.invokedFunctionARN = invokedFunctionARN
-        self.traceID = traceID
+        self.traceID = headers.first(name: AmazonHeaders.traceID) ?? ""
         self.clientContext = headers["Lambda-Runtime-Client-Context"].first
         self.cognitoIdentity = headers["Lambda-Runtime-Cognito-Identity"].first
     }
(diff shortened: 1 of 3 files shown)


Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
```diff
@@ -51,14 +51,10 @@ struct Invocation: Hashable {
...
 
-        guard let traceID = headers.first(name: AmazonHeaders.traceID) else {
-            throw Lambda.RuntimeError.invocationMissingHeader(AmazonHeaders.traceID)
-        }
-
         self.requestID = requestID
         self.deadlineInMillisSinceEpoch = unixTimeInMilliseconds
         self.invokedFunctionARN = invokedFunctionARN
-        self.traceID = traceID
+        self.traceID = headers.first(name: AmazonHeaders.traceID) ?? ""
```
tomerd, line 57: should this emit a warning in the case its missing? wdyt @fabianfett ?

tomerd:

Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
fabianfett: I think, we should just add a self created traceID. We already have a function for this:
```suggestion
        self.traceID = headers.first(name: AmazonHeaders.traceID) ?? 
            "Root=\(AmazonHeaders.generateXRayTraceID());Sampled=0"
```

fabianfett (commented):
Also we should a test to ensure, this continues to work.

tomerd:
thanks @StefanNienhuis see feedback inline

fabianfett:

tomerd:
thanks @StefanNienhuis looks like there are some formatting issues the validation script found. you can also run it locally with `docker-compose -f docker/docker-compose.yaml run --rm soundness`

.vscode/launch.json
```diff
@@ -0,0 +1,33 @@
...
+            "name": "Test swift-aws-lambda-runtime",
+            "program": "/Applications/Xcode.app/Contents/Developer/usr/bin/xctest",
+            "args": [
+                ".build/debug/swift-aws-lambda-runtimePackageTests.xctest"
+            ],
+            "cwd": "${workspaceFolder:swift-aws-lambda-runtime}",
+            "preLaunchTask": "swift: Build All"
+        }
+    ]
+}
```
tomerd, line 33: .gitignore ?

Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
```diff
@@ -54,8 +54,7 @@ struct Invocation: Hashable {
         self.requestID = requestID
         self.deadlineInMillisSinceEpoch = unixTimeInMilliseconds
         self.invokedFunctionARN = invokedFunctionARN
-        self.traceID = headers.first(name: AmazonHeaders.traceID) ?? 
-            "Root=\(AmazonHeaders.generateXRayTraceID());Sampled=0"
+        self.traceID = headers.first(name: AmazonHeaders.traceID) ?? "Root=\(AmazonHeaders.generateXRayTraceID());Sampled=0"
```
tomerd, line 57: should we add a small unit test too so this does not regress?

.vscode/launch.json
StefanNienhuis: I removed it from the commit for now. Do you want me to add it to .gitignore?

Sources/AWSLambdaRuntimeCore/ControlPlaneRequest.swift
tomerd: testing that we generate a default traceID if no header is provided

StefanNienhuis: I have that ready. I used the header from another file and changed the year to 2022, but now soundness.sh is complaining that it's not an acceptable year. Should I add `s/2022/YEARS/` to the `replace_acceptable_years()` function?

StefanNienhuis: See 374dc71

fabianfett (approved):

fabianfett:

tomerd (approved):

PR #251, opened 2022-03-17
terminator handler

tomerd:
motivation: make it simpler to register shutdown hooks

changes:
* introduce Terminator helper that allow registering and de-registaring shutdown handlers
* expose the new terminator handler on the InitializationContext and deprecate ShutdownContext
* deprecate the Handler::shutdown protocol requirement
* update the runtime code to use the new terminator instead of calling shutdown on the handler
* add and adjust tests

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Terminator.swift
+++ b/Sources/AWSLambdaRuntimeCore/Terminator.swift
@@ -57,7 +57,7 @@ extension Lambda {
         /// - parameters:
         ///     - eventLoop: The `EventLoop` to run the termination on.
         ///
-        /// - Returns: An`EventLoopFuture` with the result of the termination cycle.
+        /// - Returns: An `EventLoopFuture` with the result of the termination cycle.
         internal func terminate(eventLoop: EventLoop) -> EventLoopFuture<Void> {
             func terminate(_ iterator: IndexingIterator<[(name: String, handler: Handler)]>, errors: [Error], promise: EventLoopPromise<Void>) {
                 var iterator = iterator
--- a/Sources/AWSLambdaRuntimeCore/Terminator.swift
+++ b/Sources/AWSLambdaRuntimeCore/Terminator.swift
@@ -104,7 +104,7 @@ extension LambdaTerminator {
         private var index: [RegistrationKey]
         private var map: [RegistrationKey: (name: String, handler: Handler)]
 
-        public init() {
+        init() {
             self.lock = .init()
             self.index = []
             self.map = [:]
--- a/Sources/AWSLambdaRuntimeCore/Terminator.swift
+++ b/Sources/AWSLambdaRuntimeCore/Terminator.swift
@@ -15,79 +15,94 @@
 import NIOConcurrencyHelpers
 import NIOCore
 
-extension Lambda {
-    /// Lambda terminator.
-    /// Utility to manage the lambda shutdown sequence.
-    public final class Terminator {
-        public typealias Handler = (EventLoop) -> EventLoopFuture<Void>
-        public typealias RegistrationKey = String
+/// Lambda terminator.
+/// Utility to manage the lambda shutdown sequence.
+public final class LambdaTerminator {
+    private typealias Handler = (EventLoop) -> EventLoopFuture<Void>
 
-        private var storage: Storage
+    private var storage: Storage
 
-        public init() {
-            self.storage = Storage()
-        }
-
-        /// Register a shutdown handler with the terminator
-        ///
-        /// - parameters:
-        ///     - name: Display name for logging purposes
-        ///     - handler: The shutdown handler to call when terminating the Lambda.
-        ///             Shutdown handlers are called in the reverse order of being registered.
-        ///
-        /// - Returns: A `RegistrationKey` that can be used to de-register the handler when its no longer needed.
-        @discardableResult
-        public func register(name: String, handler: @escaping Handler) -> RegistrationKey {
-            let key = LambdaRequestID().uuidString // UUID basically
-            self.storage.add(key: key, name: name, handler: handler)
-            return key
-        }
-
-        /// De-register a shutdown handler with the terminator
-        ///
-        /// - parameters:
-        ///     - key: A `RegistrationKey` obtained from calling the register API.
-        public func deregister(_ key: RegistrationKey) {
-            self.storage.remove(key)
-        }
-
-        /// Begin the termination cycle
-        /// Shutdown handlers are called in the reverse order of being registered.
-        ///
-        /// - parameters:
-        ///     - eventLoop: The `EventLoop` to run the termination on.
-        ///
-        /// - Returns: An `EventLoopFuture` with the result of the termination cycle.
-        internal func terminate(eventLoop: EventLoop) -> EventLoopFuture<Void> {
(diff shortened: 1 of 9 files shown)


tomerd:
@fabianfett this is still wip but wanted to get an initial feedback on the direction / approach

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeTest.swift
```diff
@@ -66,23 +66,37 @@ class LambdaRuntimeTest: XCTestCase {
...
+                context.terminator.register(name: "test 1", handler: { eventLoop in
+                    eventLoop.makeFailedFuture(ShutdownError(description: "error 1"))
+                })
+                context.terminator.register(name: "test 2", handler: { eventLoop in
+                    eventLoop.makeSucceededVoidFuture()
+                })
+                context.terminator.register(name: "test 3", handler: { eventLoop in
+                    eventLoop.makeFailedFuture(ShutdownError(description: "error 2"))
+                })
+                context.terminator.register(name: "test 3", handler: { eventLoop in
```
yim-lee, line 88: These two terminators have the same name. Intended?

yim-lee (commented):
Overall looks reasonable. Got one question about the tests.

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeTest.swift
tomerd: good catch

tomerd, line 88:
```suggestion
                context.terminator.register(name: "test 4", handler: { eventLoop in
```

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeTest.swift
```diff
@@ -66,23 +66,37 @@ class LambdaRuntimeTest: XCTestCase {
...
+                    eventLoop.makeSucceededVoidFuture()
+                })
+                context.terminator.register(name: "test 4", handler: { eventLoop in
```
tomerd, line 91:
```suggestion
                context.terminator.register(name: "test 5", handler: { eventLoop in
```

yim-lee (approved):

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIOConcurrencyHelpers
+import NIOCore
+
+extension Lambda {
+    /// Lambda terminator.
+    /// Utility to manage the lambda shutdown sequence.
+    public final class Terminator {
```
fabianfett, line 21: I think, I would prefer, if we were calling this `LambdaTerminator` instead of `Lambda.Terminator`.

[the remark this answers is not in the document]
fabianfett: I think the Terminator should be `Sendable` for 5.6 and onwards.

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -0,0 +1,133 @@
...
+        public typealias Handler = (EventLoop) -> EventLoopFuture<Void>
+        public typealias RegistrationKey = String
```
fabianfett, line 23: I think, to allow easier API evolution, we should make this a real type:

```
extension LambdaTerminator {
  struct RegistrationKey: Hashable, Sendable {
    var key: String
  }
}
```

fabianfett, line 23: I think, to allow easier API evolution, we should make this a real type:

```
extension LambdaTerminator {
  struct RegistrationKey: Hashable, Sendable {
    var key: String
  }
}
```

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -0,0 +1,133 @@
...
+            return promise.futureResult
+        }
+    }
+
+    private final class Storage {
+        private let lock: Lock
+        private var index: [String]
+        private var map: [String: (name: String, handler: Terminator.Handler)]
+
+        public init() {
```
fabianfett, line 92:
```suggestion
        init() {
```

fabianfett, line 92:
```suggestion
       init() {
```

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -0,0 +1,133 @@
...
+extension Result {
+    fileprivate var error: Error? {
+        switch self {
+        case .failure(let error):
+            return error
+        case .success:
+            return .none
+        }
+    }
+}
```
fabianfett, line 133: Is that used?

fabianfett, line 133: Is that used?

fabianfett (commented):
This is much better, than the ShutdownHandler method.

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -0,0 +1,133 @@
...
+        /// Register a shutdown handler with the terminator
+        ///
+        /// - parameters:
+        ///     - name: Display name for logging purposes
+        ///     - handler: The shutdown handler to call when terminating the Lambda.
+        ///             Shutdown handlers are called in the reverse order of being registered.
+        ///
+        /// - Returns: A `RegistrationKey` that can be used to de-register the handler when its no longer needed.
+        @discardableResult
+        public func register(name: String, handler: @escaping Handler) -> RegistrationKey {
```
fabianfett, line 40: As you know, I'm not a big fan of the typealiases for closures, since they destroy code completion.

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -0,0 +1,133 @@
...
+    /// Utility to manage the lambda shutdown sequence.
+    public final class Terminator {
+        public typealias Handler = (EventLoop) -> EventLoopFuture<Void>
+        public typealias RegistrationKey = String
+
+        private var storage: Storage
+
+        public init() {
+            self.storage = Storage()
+        }
```
fabianfett, line 29: Do we need the init to be public?

Sources/AWSLambdaRuntimeCore/Terminator.swift
fabianfett: Was this missed?

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -0,0 +1,139 @@
...
+
+        public var description: String {
+            self.value
+        }
+    }
+}
+
+extension LambdaTerminator {
+    private final class Storage {
```
fabianfett, line 102: NIT: I don't really see the need for the Storage class here, since the TerminationHandler is a class as well. But I don't think that this should stop us from merging.

fabianfett (approved):
Two small nits, left, that shouldn't stop us from merging.

Sources/AWSLambdaRuntimeCore/Terminator.swift
tomerd: I'd like to keep that abstraction

PR #252, opened 2022-03-18
adoption of sendable

tomerd:
motivation: adopt to sendable requirments in swift 5.6

changes:
* define sendable shims for protocols and structs that may be used in async context
* adjust tests
* add a test to make sure no warning are emittted

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -26,7 +26,7 @@ import NIOCore
 ///         level protocols ``EventLoopLambdaHandler`` and
 ///         ``ByteBufferLambdaHandler``.
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-public protocol LambdaHandler: EventLoopLambdaHandler {
+public protocol LambdaHandler: EventLoopLambdaHandler where Event: _AWSLambdaSendable {
     /// The Lambda initialization method
     /// Use this method to initialize resources that will be used in every request.
     ///
@@ -157,7 +157,7 @@ extension EventLoopLambdaHandler where Output == Void {
 /// - note: This is a low level protocol designed to power the higher level ``EventLoopLambdaHandler`` and
 ///         ``LambdaHandler`` based APIs.
 ///         Most users are not expected to use this protocol.
-public protocol ByteBufferLambdaHandler {
+public protocol ByteBufferLambdaHandler: _ByteBufferLambdaHandlerSendable {
     /// Create your Lambda handler for the runtime.
     ///
     /// Use this to initialize all your resources that you want to cache between invocations. This could be database
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -26,7 +26,7 @@ import NIOCore
 ///         level protocols ``EventLoopLambdaHandler`` and
 ///         ``ByteBufferLambdaHandler``.
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-public protocol LambdaHandler: EventLoopLambdaHandler where Event: _AWSLambdaSendable {
+public protocol LambdaHandler: EventLoopLambdaHandler {
     /// The Lambda initialization method
     /// Use this method to initialize resources that will be used in every request.
     ///
@@ -69,15 +69,16 @@ extension LambdaHandler {
 }
 
 /// unchecked sendable wrapper for the handler
+/// this is safe since lambda runtime is designed to calls the handler serially
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-fileprivate struct UncheckedSendableHandler<Underlying: LambdaHandler>: @unchecked Sendable {
+fileprivate struct UncheckedSendableHandler<Underlying: LambdaHandler, Event, Output>: @unchecked Sendable where Event == Underlying.Event, Output == Underlying.Output {
     let underlying: Underlying
 
     init(underlying: Underlying) {
         self.underlying = underlying
     }
 
-    func handle(_ event: Underlying.Event, context: LambdaContext) async throws -> Underlying.Output {
+    func handle(_ event: Event, context: LambdaContext) async throws -> Output {
         try await self.underlying.handle(event, context: context)
     }
 }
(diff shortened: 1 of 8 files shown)


Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -26,7 +26,7 @@ import NIOCore
 ///         level protocols ``EventLoopLambdaHandler`` and
 ///         ``ByteBufferLambdaHandler``.
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-public protocol LambdaHandler: EventLoopLambdaHandler {
+public protocol LambdaHandler: EventLoopLambdaHandler where Event: _AWSLambdaSendable {
```
tomerd, line 29: @fabianfett this is an interesting bit. since we are using it in an async context in L62 it needs to be sendable, or we can wrap it if we are sure access to it is safe

fabianfett: The Event does not need to be `Sendable`, access will always be serial.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -157,7 +157,7 @@ extension EventLoopLambdaHandler where Output == Void {
 /// - note: This is a low level protocol designed to power the higher level ``EventLoopLambdaHandler`` and
 ///         ``LambdaHandler`` based APIs.
 ///         Most users are not expected to use this protocol.
-public protocol ByteBufferLambdaHandler {
```
fabianfett, line 160: IMO the `ByteBufferLambdaHandler` does not need to be sendable. The runtime will always call it in a serial manner. We should wrap the protocol in a `UncheckedSendableBox`.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
fabianfett: Another question IMO is, should we require the Output to be Sendable? Users could do something like this:

```
struct MyHandlerThatTriesToDoSomethingBad {
  func handle(_ event: Event, context: LambdaContext) async throws -> MyOutput {
    var output = Foo()
    Task {
      output.bar() // <--- compiler will require Sendable here!
    }
    return output
  }
}
```

I don't think we need to require Output to be Sendable as well. As the example above shows, as soon as the user would do something illegal, the compiler would step in and require sendability, if it is needed.

fabianfett (commented):
I think the really important type is `LambdaRuntime`, which we need to make `Sendable`.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
tomerd: @fabianfett what if the users want to use the event payload in the task?

struct MyHandlerThatTriesToDoSomethingBad {
  func handle(_ event: Event, context: LambdaContext) async throws -> MyOutput {
    var output = Foo()
    Task {
        someFunc(event)
    }
    return output
  }
}

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -58,13 +58,27 @@ extension LambdaHandler {
...
+
+    init(underlying: Underlying) {
+        self.underlying = underlying
+    }
+
+    func handle(_ event: Underlying.Event, context: LambdaContext) async throws -> Underlying.Output {
+        try await self.underlying.handle(event, context: context)
+    }
+}
+
```
tomerd, line 81: @fabianfett wdyt?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
fabianfett: @tomerd That is a valid point. For this reason we should make all the events we expose in `swift-aws-lambda-events` Sendable. But the *Runtime* does not and should not need to require it.

fabianfett (commented):
Great progress! I think we must not forget to mark `LambdaRuntime` and `LambdaTerminator` also as `Sendable`.

tomerd:
>  I think we must not forget to mark LambdaRuntime and LambdaTerminator also as Sendable.

yes, I will merge the terminator PR first than rebase and update

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -137,3 +137,10 @@ extension LambdaTerminator {
         let underlying: [Error]
     }
 }
+
+// Ideally this would not be @unchecked Sendable, but Sendable checks do not understand locks
+// We can transition this to an actor once we drop support for older Swift versions
```
fabianfett, line 142: We can transition this to an actor once we have custom executors :)

tomerd: that too

tomerd:
> I think we must not forget to mark LambdaRuntime and LambdaTerminator also as Sendable.

done for LambdaTerminator. why would we want LambdaRuntime to be Sendable? it is not used directly by the users atm afaik.

fabianfett:
> Why would we want LambdaRuntime to be Sendable? it is not used directly by the users atm afaik.

@tomerd `LambdaRuntime` is supposed to be used by framework developers that wish to integrate with Lambda, since it allows adopters to hook it into their own lifecycle methods. The [vapor lambda bridge](https://github.com/vapor-community/vapor-aws-lambda-runtime) works this way. For this reason LambdaRuntime is also part of the public API.

LambdaRuntime should be `@unchecked Sendable`, since it is thread safe by dispatching to the same eventLoop. This is currently not true for the start method, which needs to be fixed with the pattern:

```
    public func start() -> EventLoopFuture<Void> {
        if self.eventLoop.inEventLoop {
            return self.start0()
        } else {
            return self.flatSubmit { self.start0() }
        }
    }

    private func start0() -> EventLoopFuture<Void> {
        // current start implementation
    }
```

After that we should have marked everything in the public API as Sendable that needs to be Sendable.

Sources/AWSLambdaRuntimeCore/LambdaRuntime.swift
```diff
@@ -189,3 +196,8 @@ public final class LambdaRuntime<Handler: ByteBufferLambdaHandler> {
         }
     }
 }
+
+/// This is safe since lambda runtime is intended to be used within a single `EventLoop`
```
fabianfett, line 200:
```suggestion
/// This is safe since lambda runtime synchronizes by dispatching all methods to a single `EventLoop`
```

fabianfett (approved):
Awesome! I think we can make one comment a little clearer. Besides that, great work!

PR #254, opened 2022-03-18
[wip] packaging plugin

tomerd:
motivation: an easy way to archive lambda for uploading to AWS

changes:
* add SwiftPM plugin with the verb "archive"
* on macOS use docker to build and package the lambda(s)
* on Linux, build and package directly using the plugin bridge into SwiftPM

TODO
* use libc instead of zip
* read configuration flag
* cleanup and tests


tomerd:
cc @abertelrud - this requires escaping the sandbox because of docker. would appreciate your feedback on the direction here

Examples/Echo/Package.swift
```diff
@@ -1,7 +1,16 @@
...
 import PackageDescription
 
+// this is the dependency on the swift-aws-lambda-runtime library
+var dependencies = [Package.Dependency]()
+if FileManager.default.fileExists(atPath: "../../Package.swift") {
+    dependencies.append(Package.Dependency.package(name: "swift-aws-lambda-runtime", path: "../.."))
+} else {
+    dependencies.append(Package.Dependency.package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", branch: "main"))
+}
+
```
fabianfett, line 13: Why do we need this?

Package@swift-5.2.swift
```diff
@@ -0,0 +1,55 @@
+// swift-tools-version:5.2
```
fabianfett, line 1: Let's go to 5.4 directly here and drop 5.2 and 5.3.

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,344 @@
...
+            buildConfiguration: configuration.buildConfiguration,
+            verboseLogging: configuration.verboseLogging
+        )
+        #elseif os(Linux)
+        let builtProducts = try self.build(
+            products: configuration.products,
+            buildConfiguration: configuration.buildConfiguration,
+            verboseLogging: configuration.verboseLogging
+        )
+        #else
```
fabianfett, line 48: Not sure this is the right condition... What happens if I am on Ubuntu Linux?

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,344 @@
...
+        return results
+    }
+
+    private func package(
+        products: [LambdaProduct: Path],
+        toolsProvider: (String) throws -> Path,
+        outputDirectory: Path,
+        verboseLogging: Bool
+    ) throws -> [LambdaProduct: Path] {
+        let zipToolPath = try toolsProvider("zip")
```
fabianfett, line 162: If we do this, then we need to install zip in the AmazonLinux docker container first, since it should not be preinstalled.

fabianfett (commented):
Great start. Love to see this arriving! A couple of quick thoughts.

Plugins/AWSLambdaPackager/Plugin.swift
tomerd: good point. do we want to check for distro?

Plugins/AWSLambdaPackager/Plugin.swift
tomerd: right. i wasn't to explore what using zlib would look like

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,344 @@
...
+
+             */
+            // docker run --rm -v "$workspace":/workspace -w /workspace/Examples/Deployment builder \
+            //               bash -cl "./scripts/package.sh $executable"
+        }
+        return archives
+    }
+
+    @discardableResult
+    private func execute(
```
abertelrud, line 214: This kind of thing seems common enough that I wonder if PackagePlugins should provide a convenience cover.

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,344 @@
...
+    case unknownExecutable(String)
+    case buildError(String)
+    case failedWritingDockerfile
+    case processFailed(Int32)
+    case invalidProcessOutput
+}
+
+extension PackageManager.BuildResult {
+    // find the executable produced by the build
+    func executableArtifact(for product: Product) throws -> PackageManager.BuildResult.BuiltArtifact? {
```
abertelrud, line 314: Should this be in PackagePlugin in the future?  Seems like a common thing to want.

Plugins/AWSLambdaPackager/Plugin.swift
fabianfett: I guess we have to...

Package@swift-5.4.swift
```diff
@@ -0,0 +1,55 @@
+// swift-tools-version:5.4
```
yim-lee, line 1: And we are effectively dropping support for Swift < 5.4?

Package@swift-5.5.swift
```diff
@@ -0,0 +1 @@
+Package@swift-5.4.swift
```
yim-lee, line 1: 👍

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,390 @@
...
+        )
+
+        print("\(archives.count > 0 ? archives.count.description : "no") archive\(archives.count != 1 ? "s" : "") created")
+        for (product, archivePath) in archives {
+            print("  * \(product.name) at \(archivePath.string)")
+        }
+    }
+
+    private func buildInDocker(
+        packageIdentity: Package.ID,
```
yim-lee, line 68: Unused?

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,390 @@
...
+        var archives = [LambdaProduct: Path]()
+        for (product, artifactPath) in products {
+            print("-------------------------------------------------------------------------")
+            print("archiving \"\(product.name)\"")
+            print("-------------------------------------------------------------------------")
+
+            // prep zipfile location
+            let workingDirectory = outputDirectory.appending(product.name)
+            let zipfilePath = workingDirectory.appending("\(product.name).zip")
+            if FileManager.default.fileExists(atPath: workingDirectory.string) {
```
yim-lee, line 165: Do we mean to check the working dir exists or the zip file?

What if `workingDirectory` is not a dir?

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,390 @@
...
+                try FileManager.default.removeItem(atPath: workingDirectory.string)
```
yim-lee, line 166: It seems bad to delete a dir/file quietly. Should we throw error or at least print warning?

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,390 @@
...
+            // rename artifact to "bootstrap"
+            let relocatedArtifactPath = workingDirectory.appending(artifactPath.lastComponent)
+            let symbolicLinkPath = workingDirectory.appending("bootstrap")
+            try FileManager.default.copyItem(atPath: artifactPath.string, toPath: relocatedArtifactPath.string)
+            try FileManager.default.createSymbolicLink(atPath: symbolicLinkPath.string, withDestinationPath: relocatedArtifactPath.lastComponent)
+
+            #if os(macOS) || os(Linux)
+            let arguments = ["--junk-paths", "--symlinks", zipfilePath.string, relocatedArtifactPath.string, symbolicLinkPath.string]
+            #else
+            throw Error.unsupportedPlatform("can't or don't know how to create a zipfile on this platform")
```
yim-lee, line 179:
```suggestion
            throw Error.unsupportedPlatform("can't or don't know how to create a zip file on this platform")
```

Package@swift-5.4.swift
tomerd: yea that is the intention

Plugins/AWSLambdaPackager/Plugin.swift
tomerd: this is a working directory like SwiftPM `.build` directory, the deletion is to make sure we have a clean state and dont end up with old zip files etc. happy to print something tho if we think it s useful.

Plugins/AWSLambdaPackager/Plugin.swift
tomerd: we need a working directory here to create several files, and finally the zip archive

Plugins/AWSLambdaPackager/Plugin.swift
fabianfett: I agree with Tom here.

tomerd:
@fabianfett I think this is ready to merge as iteration 1. next iteration we can explore how to integrate ziplib

fabianfett (commented):
As part of my review, I tried to run it in the EchoHandler... However I got the following output:

```
➜  Echo git:(feature/packaging-plugin) ✗ swift package archive     
building "echo" in docker
error: processFailed(126)
```

I think we should at least print the processes output, in the error case. This error is undebugable for me.

tomerd:
> I think we should at least print the processes output

fixed

fabianfett:
Interesting, just pulled the newest version and tried to rerun:

```bash
➜  Echo git:(feature/packaging-plugin) ✗ swift package archive
building "echo" in docker
docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/create": dial unix /var/run/docker.sock: connect: operation not permitted.
See 'docker run --help'.
error: /usr/local/bin/docker run --rm -v /Users/fabian/Developer/swift-server/swift-aws-lambda-runtime/Examples/Echo:/workspace -w /workspace swift:amazonlinux2 bash -cl swift build -c release --show-bin-path failed with code 126
➜  Echo git:(feature/packaging-plugin) ✗ docker container create swift:5.6-focal
98decac8da409010202c45428d18e06f9761f1b130c649afc14778b6d74aa72b
➜  Echo git:(feature/packaging-plugin) ✗ 
```

Apparently from the plugin I don't have enough rights to create a new container, but just from the shell I have. Do you have any idea what this might be related to @tomerd.

stevapple:
I think the reason’s here: 
https://github.com/apple/swift-evolution/blob/55f9bb7eab8e89bb8cd2794083aba101ac07777a/proposals/0303-swiftpm-extensible-build-tools.md?plain=1#L1018-1020

Wonder if we can surpass the restriction.

tomerd:
right, this is because talking to docker is networking which is defined by the sandbox.  for local testing I have been using the plugin without the sandbox, e.g:  `swift package --disable-sandbox --package-path Examples/Echo archive`

@abertelrud would be interesting to see if we can poke a hole for docker in the sandbox, or be able to express it as a plugin requirement that the user can approve

fabianfett:
Trying to use this branch with Xcode 13.3 results in the following error:

> the package manifest at '/Package@swift-5.5.swift' cannot be accessed (InternalError(description: "Internal error. Please file a bug at https://bugs.swift.org with this info. symlinks not supported")) in https://github.com/tomerd/swift-aws-lambda-runtime.git

fabianfett:
Reading a bit into this... Isn't the way we currently try to solve this the wrong way around?

https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#version-specific-manifest-selection

If I understand the docu correctly it should work like this: 

- `Package.swift` (for everything older than Swift 5.6) 
- `Package@swift-5.6.swift` (for 5.6 and above)

Package@swift-5.6.swift
```diff
@@ -0,0 +1,66 @@
...
+            .product(name: "NIOCore", package: "swift-nio"),
+            .product(name: "NIOConcurrencyHelpers", package: "swift-nio"),
+            .product(name: "NIOPosix", package: "swift-nio"),
+        ]),
+        .plugin(
+            name: "AWSLambdaPackager",
+            capability: .command(
+                intent: .custom(
+                    verb: "archive",
+                    description: "Archive the Lambda binary and prepare it for uploading to AWS. Requires docker on macOS."
```
fabianfett, line 41:
```suggestion
                    description: "Archive the Lambda binary and prepare it for uploading to AWS. Requires docker on macOS or non Amazonlinux 2 distributions."
```

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -0,0 +1,425 @@
...
+            if FileManager.default.fileExists(atPath: workingDirectory.string) {
+                try FileManager.default.removeItem(atPath: workingDirectory.string)
+            }
+            try FileManager.default.createDirectory(atPath: workingDirectory.string, withIntermediateDirectories: true)
+
+            // rename artifact to "bootstrap"
+            let relocatedArtifactPath = workingDirectory.appending(artifactPath.lastComponent)
+            let symbolicLinkPath = workingDirectory.appending("bootstrap")
+            try FileManager.default.copyItem(atPath: artifactPath.string, toPath: relocatedArtifactPath.string)
+            try FileManager.default.createSymbolicLink(atPath: symbolicLinkPath.string, withDestinationPath: relocatedArtifactPath.lastComponent)
```
fabianfett, line 178: Code comment does not match behavior. I wonder if just renaming the executable (like the code comment says), is actually the better alternative here. What do we gain through the symlink indirection?

fabianfett (approved):
Looks good and works for me!

Plugins/AWSLambdaPackager/Plugin.swift
tomerd: I cant remember tbh, I think I tired it and it caused some issue, will need to try again. in either case its the same end result

PR #255, opened 2022-03-31
Add async LambdaHandler.shutdown

adam-fowler:
Add async shutdown method to `LambdaHandler` protocol

### Motivation:

`LambdaHandler` provides async interfaces for `init` and `handle` but not `shutdown`

### Modifications:

- Add protocol requirement `shutdown(context: Lambda.ShutdownContext) async throws` to `LambdaHandler`
- Add default version that does nothing in extension
- Override ByteBufferLambdaHandler.shutdown to call the async version from `LambdaHandler` in extension

### Result:

Can shutdown LambdaHandler using async methods eg
```swift
final class MyLambda: LambdaHandler {
    init(context: Lambda.InitializationContext) {
        self.httpClient = HTTPClient(eventLoopGroupProvider: .shared(context.eventLoop))
    }

    func shutdown(context: Lambda.ShutdownContext) async throws {
        try await self.httpClient.shutdown()
    }
}
```

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -79,8 +79,7 @@ extension LambdaHandler {
         return promise.futureResult
     }
 
-    public func shutdown(context: Lambda.ShutdownContext) async throws {
-    }
+    public func shutdown(context: Lambda.ShutdownContext) async throws {}
 }
 
 #endif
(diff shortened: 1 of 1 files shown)


fabianfett:
I think we should add the async methods directly to the replacement for the ShutdownHandler methods in: https://github.com/swift-server/swift-aws-lambda-runtime/pull/251

tomerd:
agree with @fabianfett, let get #251 over the finish line.

adam-fowler:
Ok. I'm happy to close this in favour of #251 getting the async support

PR #257, opened 2022-04-14
remove extension for JSONEncoder and JSONDecoder

tomerd:
motivation: we should not publicly extend types we do not own

change: remove extensions which are largely API sugar that is not directly related to Lambda


fabianfett (approved):
I feel much better now! Thanks!

PR #258, opened 2022-04-15
Add script to check for API breaks

fabianfett:
We should check for API breaking changes in CI, to ensure we don't accidentally break API going forwards.

### Motivation:

- We should check for API breaking changes in CI, to ensure we don't accidentally break API going forwards.

### Modifications:

- Add a script that checks for breaking changes (copy and pasted from NIO).

### Result:

- We ensure we don't accidentally break API after 1.0

### Open tasks:

@tomerd we need to add another CI that runs the API checking script.

### DIFF
--- a/scripts/check_no_api_breakages.sh
+++ b/scripts/check_no_api_breakages.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftAWSLambdaRuntime open source project
+##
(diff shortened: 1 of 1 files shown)


tomerd:
thanks @fabianfett, should we add a docker compose task that runs this, or do we need to do that at the CI server level?

fabianfett:
@tomerd I tried to see what we do in AHC. However I did not find a special docker-compose task for it. So I assume it is activated in the CI server.

tomerd:

tomerd:

tomerd (approved):

PR #259, opened 2022-04-15
add new abstraction for simple Lambda that does not need explicit initialization

tomerd:
motivation: make simple things simple

changes:
* add new abstraction names SimpleLambdaHandler that has a default initializer
* add test

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -84,6 +84,31 @@ fileprivate struct UncheckedSendableHandler<Underlying: LambdaHandler, Event, Ou
 }
 #endif
 
+// MARK: - SimpleLambdaHandler
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+/// Strongly typed, processing protocol for a Lambda that takes a user defined
+/// ``EventLoopLambdaHandler/Event`` and returns a user defined
+/// ``EventLoopLambdaHandler/Output`` asynchronously.
+///
+/// - note: Simpler use cases that do not special initialization should implement this protocol instead of the lower
+///         level protocols ``LambdaHandler``,
+///         ``EventLoopLambdaHandler`` and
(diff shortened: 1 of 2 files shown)


tomerd:
we could / should consider doing this on `LambdaHander` instead of the new abstraction, also not in love with the "Simple" prefix.

stevapple:
We can drop `context:` from `LambdaHandler`’s initializer, then the `SimpleLambdaHandler` model will just fit (we can keep `async` and `throws` so users can still initialize some resources if they want).

tomerd:
@stevapple `context` is needed when users need to initialize resources, it provides thinks like the NIO EventLoop and Terminator, so we cannot drop it.

stevapple:
Is there any reason they cannot use async/await for asynchronous startup, and use throw in place of manual termination? I think that's the "simple" way and should work for most people.

tomerd:
termination is for resource cleanup when shutting down, there needs to be an entry point to that. also all NIO based clients (eg db client based on NIO) need access the the `EventLoop` for initialization to avoid thread hops

fabianfett (commented):
While I like the idea of making Lambda easier to use, I wonder if we actually achieve this goal through this change. Until now, IMO, it was clear which Handler protocol we recommend: LambdaHandler. Now the story is becoming harder.

Let's look at the pros and cons of this change:

Pro:

- We can remove 1-3 lines for very simple LambdaHandlers.
- Users don't need to specify an explicit init

Con:

- We need to explain to users the difference between `SimpleLambdaHandler` and `LambdaHandler` even in the Simple case. We need explain to users, that they shall use `SimpleLambdaHandler` if they don't have any resources that they want to share between invocations.
- The way from a very simple lambda to an actual lambda (I think every real lambda, will use some shared network resources) is becoming harder. We need to remind developers, that they need to change the `Handler` protocol and now need to implement a new init method. There is no compiler warning to point them into the right direction.

In my opinion the cons outweigh the pros. Is there something that I have overseen?

tomerd:
superseded by #273

PR #262, opened 2022-04-20
Drop Swift 5.2 and 5.3

fabianfett:
As outlined in a [Swift forums post in November ’21](https://forums.swift.org/t/swiftnio-swift-version-support/53232), SwiftNIO will only support the latest non-patch Swift release and the 2 immediately prior non-patch versions.

In this commit we drop support for Swift 5.2 and 5.3.


stevapple:
Have you re-run SwiftFormat after changing the config file?

fabianfett:
> Have you re-run SwiftFormat after changing the config file?

Yes, otherwise the soundness tests wouldn't pass.

yim-lee (approved):

tomerd (approved):

PR #264, opened 2022-06-04
Fix performance test script

stevapple:
Fix the performance test script

### Motivation:

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications:

_[Describe the modifications you've done.]_

### Result:

_[After your change, what will change.]_


tomerd:

tomerd (approved):

tomerd:
thanks @stevapple

Issue #266, opened 2022-06-22
Use primary associated type to reduce boilerplate


stevapple:
After [SE-0346](https://github.com/apple/swift-evolution/blob/main/proposals/0346-light-weight-same-type-syntax.md) lands in Swift 5.7, it’s possible for us to mark `Event` and `Output` as primary associate types for `LambdaHandler`. This can simplify a demo handler from:

```swift
@main
struct MyLambda: LambdaHandler {
    typealias Event = String
    typealias Output = String

    init(context: LambdaInitializationContext) async throws {
        // setup your resources that you want to reuse for every invocation here.
    }

    func handle(_ input: String, context: LambdaContext) async throws -> String {
        // as an example, respond with the input's reversed
        String(input.reversed())
    }
}
```

into

```swift
@main
struct MyLambda: LambdaHandler<String, String> {
    init(context: LambdaInitializationContext) async throws {
        // setup your resources that you want to reuse for every invocation here.
    }

    func handle(_ input: String, context: LambdaContext) async throws -> String {
        // as an example, respond with the input's reversed
        String(input.reversed())
    }
}
```

sebsto:
Implemented in v2 
https://github.com/swift-server/swift-aws-lambda-runtime/blob/54fd25bdad26ac9d2a0cd593a82ed438dc7e9825/Examples/HelloWorld/Sources/main.swift#L19

PR #272, opened 2022-09-30
Improves Developer Ergonomics for LambdaHandler Conformances

hectormatos2011:
This PR aims to improve developer ergonomics for simple `LambdaHandler` conformances when trying to quickly put together a local server to test against.

### Motivation:

Swift 5.7 added primary associated types for protocols where you can start using generic bracket syntax for protocols. An issue for this was opened not too long ago in this repo (#266) with the primary motivation of improving developer ergonomics around types conforming to `LambdaHander`. However, swift's type inference system _should_ be able to handle the aforementioned issue just fine without using primary associated types (although support for that could still be useful when defining properties conforming to `LambdaHandler` when you want to constrain the types).

Swift's type inference system does not pick up the `Event` and `Output` types for objects conforming to `LambdaHandler` because those types are defined in the covariant protocol `EventLoopLambdaHandler`. This means that you need to add type aliases for `Event` and `Output` when conforming to `LambdaHandler` - resulting in less than ideal developer ergonomics when trying to spin up a quick little local server.

Additionally, the initializer for `LambdaHandler` is currently a required function in the protocol. This is super useful for setup logic for your Handlers but degrades the developer ergonomics of the most simple/minimal use-case for LambdaHandlers. Developers should be able to continue having the capability of writing simple lambda handlers like they could do in 0.5.2:

```swift
import AWSLambdaRuntime
import Foundation

Lambda.run { (context, input: Request, callback: @escaping (Result<String, Error>) -> Void) in
    callback(.success(...))
}
```

The current state of affairs in comparison is to explicitly define the `Event` and `Output` associated types of `LambdaHandler` when writing a simple lambda as well as being forced to include an init that isn't necessarily always required to successfully run a simple lambda:

```swift
import AWSLambdaRuntime
import Foundation

@main
struct EntryHandler: LambdaHandler {
    typealias Event = SomeCodableEventType
    typealias Output = SomeCodableOutputType
   
    init(context: AWSLambdaRuntimeCore.LambdaInitializationContext) async throws {
        // The body of this could be blank and the lambda would still work
    }
   
    func handle(_ event: Event, context: LambdaContext) async throws -> Output {
        // lambda logic
    }
}
```

With the changes on main, a developer loses the ease of writing a three-line lambda that they could do in 0.5.2.

On top of all of this, when needing to run a server locally in Xcode, a developer would need to add an environment variable to their local scheme. This isn't a much used feature in Xcode for developers coming from iOS who are looking to maybe start writing swift on the server. This isn't too big of a deal, but could be a bit of pain to do when trying to set yourself up with Lambda for the first time. It also introduces the possibility of having to manage two different environment variables if you're switching between debugging in Xcode or running the server locally in your Terminal. It would be nice to have the capability of defining this in code so a developer wouldn't have to open a separate window to toggle a server running locally.

I may be wrong about this (not a lot of experience writing swift on the server), but if a developer is running a local server in Xcode, it should be safe to assume that the server isn't running on an environment like AWS Lambda or Docker. I think that if a server is running their Lambda in Xcode, `LOCAL_LAMBDA_SERVER_ENABLED` should be enabled by default. Doing so would enable developers to package ready-to-go example local servers with their example mobile apps (maybe they want to demonstrate the usage of an SDK or something).

Currently, developers in this situation would have to instruct their users to set up this `LOCAL_LAMBDA_SERVER_ENABLED` environment variable. If a user is a swift developer already, they might have the urge to open that Package in Xcode and just hit run without knowing they need to set that variable up. 

With the support of a variable for enabling running as a local server in Lambda directly, a developer could package a ready-to-go local server package with their example apps that would only require opening the package and running in Xcode with no extra set up.

### Modifications:

* Removes ambiguities between `LambdaHandler`, `EventLoopLambdaHandler`, and `ByteBufferLambdaHandler`. The `handle(_:context:)` functions shared across all three prevented the swift compiler from properly inferring the `Event` and `Output` types between `LambdaHandler` and `EventLoopLambdaHandler`. The `handle(_:context:)` functions are now more descriptive with their primary parameter to match the protocol they are associated with:
    * `LambdaHandler.handle(_:context)` has now been changed to `LambdaHandler(request:context:)`
    * `EventLoopLambdaHandler.handle(_:context)` has now been changed to `EventLoopLambdaHandler(event:context:)`
    * `ByteBufferLambdaHandler.handle(_:context)` has now been changed to `ByteBufferLambdaHandler(buffer:context:)`
* Adds an init function to the `LambdaHandler` protocol
    * Adds a default implementation for `init(context:)` that calls through to `init()` - this allows fully qualified structs conforming to `LambdaHandler` to omit their initializers if no setup is required for the Lambda handler
* Adds a static var `isLocalServerEnabled` to the `ByteBufferLambdaHandler` protocol
    * Default is `true` when running in Xcode, but still respects the `LOCAL_LAMBDA_SERVER_ENABLED` environment variable.
    * This defaults to `false` if running outside of Xcode
* Updates tests with new changes 

### Result:

With all of the changes in this PR, a developer could package a local lambda handler Swift Package with an example app that could look like this:

```swift
@main
final class EntryHandler: LambdaHandler {
    func handle(request: YourDecodableRequest, context: LambdaContext) -> YourEncodableResponse {
        // lambda logic
    }
}
```

All previous logic is still retained but the optional code needed is now opt-in with sensible defaults.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Lambda+String.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda+String.swift
@@ -16,7 +16,7 @@ import NIOCore
 extension EventLoopLambdaHandler where Event == String {
     /// Implementation of a `ByteBuffer` to `String` decoding.
     @inlinable
-    public func decode(buffer: ByteBuffer) throws -> String {
+    public func decode(buffer: ByteBuffer) throws -> Event {
         var buffer = buffer
         guard let string = buffer.readString(length: buffer.readableBytes) else {
             fatalError("buffer.readString(length: buffer.readableBytes) failed")
@@ -28,7 +28,7 @@ extension EventLoopLambdaHandler where Event == String {
 extension EventLoopLambdaHandler where Output == String {
     /// Implementation of `String` to `ByteBuffer` encoding.
     @inlinable
-    public func encode(allocator: ByteBufferAllocator, value: String) throws -> ByteBuffer? {
+    public func encode(allocator: ByteBufferAllocator, value: Output) throws -> ByteBuffer? {
         // FIXME: reusable buffer
         var buffer = allocator.buffer(capacity: value.utf8.count)
         buffer.writeString(value)
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -73,7 +73,7 @@ public enum Lambda {
 
         // start local server for debugging in DEBUG mode only
         #if DEBUG
-        if Lambda.env("LOCAL_LAMBDA_SERVER_ENABLED").flatMap(Bool.init) ?? false {
+        if Lambda.isLocalServer {
             do {
                 return try Lambda.withLocalServer {
                     _run(configuration)
@@ -93,6 +93,13 @@ public enum Lambda {
 // MARK: - Public API
 
 extension Lambda {
+    #if DEBUG
+    static var isLocalServer: Bool {
+        get { env("LOCAL_LAMBDA_SERVER_ENABLED").flatMap(Bool.init) ?? false }
+        set { setenv("LOCAL_LAMBDA_SERVER_ENABLED", String(newValue), 1) }
+    }
+    #endif
+    
     /// Utility to access/read environment variables
     public static func env(_ name: String) -> String? {
         guard let value = getenv(name) else {
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -43,7 +43,7 @@ public enum Lambda {
         // start local server for debugging in DEBUG mode only
         #if DEBUG
         var localServer: LocalLambda.Server? = nil
-        if Handler.isLocalServer || env("LOCAL_LAMBDA_SERVER_ENABLED").flatMap(Bool.init) ?? false {
+        if Handler.isLocalServer {
             localServer = try Lambda.startLocalServer()
         }
         #endif
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -33,11 +33,12 @@ public enum Lambda {
     ///     - handlerType: The Handler to create and invoke.
     ///
     /// - note: This is a blocking operation that will run forever, as its lifecycle is managed by the AWS Lambda Runtime Engine.
+    @discardableResult
     internal static func run<Handler: ByteBufferLambdaHandler>(
         configuration: LambdaConfiguration = .init(),
         handlerType: Handler.Type
-    ) throws {
-        var result: Result<Int, Error>?
+    ) throws -> Int {
+        var result: Result<Int, Error> = .success(0)
         
         // start local server for debugging in DEBUG mode only
         #if DEBUG
@@ -66,12 +67,12 @@ public enum Lambda {
                 #if DEBUG
                 signalSource.cancel()
                 #endif
-                result = lifecycleResult
                 eventLoop.shutdownGracefully { error in
                     if let error = error {
                         preconditionFailure("Failed to shutdown eventloop: \(error)")
                     }
                 }
+                result = lifecycleResult
             }
         }
 
@@ -81,7 +82,10 @@ public enum Lambda {
         try localServer?.stop()
         #endif
         
-        if case .failure(let error) = result {
+        switch result {
+        case .success(let count):
+            return count
+        case .failure(let error):
             throw error
         }
     }
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -36,70 +36,60 @@ public enum Lambda {
     internal static func run<Handler: ByteBufferLambdaHandler>(
         configuration: LambdaConfiguration = .init(),
         handlerType: Handler.Type
-    ) -> Result<Int, Error> {
-        let _run = { (configuration: LambdaConfiguration) -> Result<Int, Error> in
-            Backtrace.install()
-            var logger = Logger(label: "Lambda")
-            logger.logLevel = configuration.general.logLevel
-
-            var result: Result<Int, Error>!
-            MultiThreadedEventLoopGroup.withCurrentThreadAsEventLoop { eventLoop in
-                let runtime = LambdaRuntime<Handler>(eventLoop: eventLoop, logger: logger, configuration: configuration)
-                #if DEBUG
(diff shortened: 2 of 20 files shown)


Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -33,60 +33,61 @@ public enum Lambda {
...
+    @discardableResult
     internal static func run<Handler: ByteBufferLambdaHandler>(
         configuration: LambdaConfiguration = .init(),
         handlerType: Handler.Type
-    ) -> Result<Int, Error> {
-        let _run = { (configuration: LambdaConfiguration) -> Result<Int, Error> in
-            Backtrace.install()
-            var logger = Logger(label: "Lambda")
-            logger.logLevel = configuration.general.logLevel
+    ) throws -> Int {
```
hectormatos2011, line 40: The changes here is just code cleanup that should make the binary a little smaller. The logic here is exactly the same

Sources/AWSLambdaRuntimeCore/LambdaConfiguration.swift
```diff
@@ -17,42 +17,23 @@ import Logging
...
-    }
-
-    init(general: General? = nil, lifecycle: Lifecycle? = nil, runtimeEngine: RuntimeEngine? = nil) {
-        self.general = general ?? General()
-        self.lifecycle = lifecycle ?? Lifecycle()
-        self.runtimeEngine = runtimeEngine ?? RuntimeEngine()
-    }
+    var general: General = .init()
+    var lifecycle: Lifecycle = .init()
+    var runtimeEngine: RuntimeEngine = .init()
```
hectormatos2011, line 22: This is probably unnecessary if you don't like it but I updated this to be easier to read. You don't necessarily need the init functions you had before if these are just structs. Updating to vars for these internal structs still keeps the same ability to initialize these structs with the default values you had defined in your previous init functions.

Let me know if you don't want this!

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -14,13 +14,12 @@
 
 #if compiler(>=5.6)
 @preconcurrency import Dispatch
-@preconcurrency import Logging
-@preconcurrency import NIOCore
```
hectormatos2011, line 18: Xcode 14 complains about these. I don't think they're needed anymore? If this breaks something before Xcode 14 I may have to remove this

Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift
```diff
@@ -13,13 +13,8 @@
 //===----------------------------------------------------------------------===//
 
 import Logging
-#if compiler(>=5.6)
-@preconcurrency import NIOCore
-@preconcurrency import NIOHTTP1
-#else
```
hectormatos2011, line 19: Same as before, Xcode 14 says these aren't needed. Will revert this change if it breaks something before Swift 5.7

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -20,11 +20,7 @@ import NIOCore
 public final class LambdaTerminator {
     fileprivate typealias Handler = (EventLoop) -> EventLoopFuture<Void>
 
-    private var storage: Storage
-
-    init() {
-        self.storage = Storage()
-    }
```
hectormatos2011, line 27: More code cleanup. You don't need the init if all you're doing is providing a default value to the private `storage` variable

Tests/AWSLambdaRuntimeCoreTests/Utils.swift
```diff
@@ -32,18 +32,20 @@ func runLambda<Handler: ByteBufferLambdaHandler>(behavior: LambdaServerBehavior,
...
-        XCTFail("should fail with \(shouldFailWithError!)", file: file, line: line)
-    case .success(let count) where shouldFailWithError == nil:
-        XCTAssertEqual(shoudHaveRun, count, "should have run \(shoudHaveRun) times", file: file, line: line)
-    case .failure(let error) where shouldFailWithError == nil:
-        XCTFail("should succeed, but failed with \(error)", file: file, line: line)
-    case .failure(let error) where shouldFailWithError != nil:
-        XCTAssertEqual(String(describing: shouldFailWithError!), String(describing: error), "expected error to mactch", file: file, line: line)
-    default:
-        XCTFail("invalid state")
+func assertLambdaRuntimeResult(_ result: @autoclosure () throws -> Int, shouldHaveRun: Int = 0, shouldFailWithError: Error? = nil, file: StaticString = #file, line: UInt = #line) {
```
hectormatos2011, line 35: Using an autoclosure, we can guarantee that the `result` closure can get called with `try` without having to update the test function definitions to using `throws`. 

This also takes advantage of swift error handling which allows us to clean up this logic. It's more readable in my opinion but I can revert this if you don't like it.

Sources/AWSLambdaRuntimeCore/Lambda+String.swift
```diff
@@ -16,7 +16,7 @@ import NIOCore
 extension EventLoopLambdaHandler where Event == String {
     /// Implementation of a `ByteBuffer` to `String` decoding.
     @inlinable
-    public func decode(buffer: ByteBuffer) throws -> String {
+    public func decode(buffer: ByteBuffer) throws -> Event {
```
hectormatos2011, line 19: These changes help to solve ambiguities in the swift compiler across default implementations of `EventLoopLambdaHandler` when importing `AWSLambdaRuntime` or running tests for the package in Xcode

tomerd:
thanks @hectormatos2011, @fabianfett  and myself have been discussing some API changes to the main branch that should help address some of the points you bring up. should be able to put up a PR soon and then we can maybe see what else is missing and you can help get it over the finish line. wdyt?

hectormatos2011:
@tomerd I'd love to contribute! Thanks for taking a look. Would you like me to close this one? There's some good nuggets of code in here that should at least help with the type inference issue

tomerd:
lets keep it open and focus on landing https://github.com/swift-server/swift-aws-lambda-runtime/pull/273. then we can see what else we need to bring over

tomerd:
hi @hectormatos2011 would you like to revisit this now that we have the 1.0 API in place?

hectormatos2011:
Let me take a look at 1.0 and close the PR if need be

tomerd:
> @tomerd Is it currently possible to have code in the Lambda that makes it always run with LOCAL_LAMBDA_SERVER_ENABLED or do we still have to put that in the target itself? I'd like to be able to deploy example documentation that a developer could just copy and paste into Xcode without having to finagle with Edit Schemes or anything like that

that is not possible right now, to protect users from accidentally deploying the mock / debug server

PR #273, opened 2022-10-07
API Refactoring

tomerd:
refactoring the API for 1.0


tomerd:
@fabianfett ptal

Sources/AWSLambdaRuntime/Lambda+Codable.swift
```diff
@@ -19,7 +19,45 @@ import class Foundation.JSONEncoder
...
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension LambdaHandler where Output: Encodable {
+    public var encoder: LambdaCodableEncoder {
+        Lambda.defaultJSONEncoder
+    }
+}
+
+// MARK: - EventLoopLambdaHandler Codable support
 
 /// Implementation of  a`ByteBuffer` to `Event` decoding.
```
yim-lee, line 62:
```suggestion
/// Implementation of `ByteBuffer` to `Event` decoding.
```

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -44,45 +46,129 @@ public protocol LambdaHandler: EventLoopLambdaHandler {
...
+    /// Examples for this can be HTTP or database clients.
+    /// - parameters:
+    ///     - context: Runtime ``LambdaInitializationContext``.
+    public init(context: LambdaInitializationContext) async throws {
+        self.init()
+    }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+final class CodableLambdaHandler<Underlying: LambdaHandler>: ByteBufferLambdaHandler {
```
stevapple, line 87: I would suggest a name change here to avoid confusion with Swift’s `Codable`, but I’ve got no suitable name yet and it’s `internal` so we indeed can correct it later.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -44,45 +46,129 @@ public protocol LambdaHandler: EventLoopLambdaHandler {
     ///
     /// - Returns: A Lambda result ot type `Output`.
     func handle(_ event: Event, context: LambdaContext) async throws -> Output
```
stevapple, line 48: I would like to see `Context` and `InitializationContext` added as `typealias`es of lambda handlers, which should further simplify the interface.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -44,45 +46,129 @@ public protocol LambdaHandler: EventLoopLambdaHandler {
...
+    func decode(buffer: ByteBuffer) throws -> Event
 }
 
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 extension LambdaHandler {
-    public static func makeHandler(context: LambdaInitializationContext) -> EventLoopFuture<Self> {
-        let promise = context.eventLoop.makePromise(of: Self.self)
+    public init() {
+        self.init()
+    }
```
fabianfett, line 73: This looks very much like infinite recursion to me:

```swift
import XCTest
import AWSLambdaRuntime

struct Foo: Codable {}
struct Bar: Codable {}

struct FooHandler: LambdaHandler {
    typealias Event = Foo
    typealias Output = Bar

    var haha: String

        self.haha = "hahaha"
    }

    func handle(_ event: Foo, context: LambdaContext) async throws -> Bar {
        preconditionFailure()
    }
}

@available(macOS 12, *)
final class FooTests: XCTestCase {

    func testInitialization() {
        let handler = FooHandler()

        print(handler)
    }
}
```

This compiles but crashes then.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
```diff
@@ -26,13 +25,16 @@ import NIOCore
...
+    /// The lambda functions input. In most cases this should be `Codable`. If your event originates from an
+    /// AWS service, have a look at [AWSLambdaEvents](https://github.com/swift-server/swift-aws-lambda-events),
+    /// which provides a number of commonly used AWS Event implementations.
+    associatedtype Event
+    /// The lambda functions output. Can be `Void`.
+    associatedtype Output
+
+    init()
+
     init(context: LambdaInitializationContext) async throws
```
fabianfett, line 38: I still think we need two different handler protocols to enable this easy use. See below re infinite recursion.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
fabianfett: `Context` is a very generic name. I'm not sure, if I would be a fan of this change.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -33,7 +33,7 @@ internal final class LambdaRunner {
     /// Run the user provided initializer. This *must* only be called once.
     ///
     /// - Returns: An `EventLoopFuture<LambdaHandler>` fulfilled with the outcome of the initialization.
-    func initialize<Handler: ByteBufferLambdaHandler>(logger: Logger, terminator: LambdaTerminator, handlerType: Handler.Type) -> EventLoopFuture<Handler> {
+    func initialize(handlerType: (some ByteBufferLambdaHandler).Type, logger: Logger, terminator: LambdaTerminator) -> EventLoopFuture<any ByteBufferLambdaHandler> {
```
fabianfett, line 36: We must enforce 5.7 if we want to use this syntax. Currently we require 5.6. Also IMO we def. should not return `any` here. We should preserve the type.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -43,7 +43,9 @@ internal final class LambdaRunner {
             allocator: self.allocator,
             terminator: terminator
         )
-        return Handler.makeHandler(context: context)
+
+        return handlerType.makeHandler(context: context)
+            .map { $0 as any ByteBufferLambdaHandler }
```
fabianfett, line 48: don't drop types here.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -58,7 +60,7 @@ internal final class LambdaRunner {
             }
     }
 
-    func run<Handler: ByteBufferLambdaHandler>(logger: Logger, handler: Handler) -> EventLoopFuture<Void> {
+    func run(logger: Logger, handler: any ByteBufferLambdaHandler) -> EventLoopFuture<Void> {
```
fabianfett, line 63: no type dropping here.

Sources/AWSLambdaRuntimeCore/LambdaRuntime.swift
```diff
@@ -19,34 +19,58 @@ import NIOCore
 /// `LambdaRuntime` manages the Lambda process lifecycle.
 ///
 /// Use this API, if you build a higher level web framework which shall be able to run inside the Lambda environment.
-public final class LambdaRuntime<Handler: ByteBufferLambdaHandler> {
+public final class LambdaRuntime {
```
fabianfett, line 22: We did we drop the type here?

fabianfett (changes requested):
Great progress! Love the Handlers that don't depend on each other anymore. However we should preserve the types were feasible. And also I don't think we can get rid of the required initializer that easily.

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
stevapple: I think they’re specific enough for a `LambdaHandler`👀 Especially when AWS Lambda and other platforms are widely adopting the term “context” and “context object”.

stevapple: I would see this as a potential extension point, where we can introduce custom `Context` types by `ExpessibleByLambdaContext` protocol to customize their own context instead of extending `LambdaContext` (which is bad practice).

Examples/JSON/Lambda.swift
```diff
@@ -30,10 +30,6 @@ struct MyLambda: LambdaHandler {
     typealias Event = Request
     typealias Output = Response
```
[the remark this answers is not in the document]
tomerd: unfortuently they are, need to see with the compiler team why this is the case

Examples/LocalDebugging/MyLambda/Lambda.swift
```diff
@@ -23,10 +23,6 @@ struct MyLambda: LambdaHandler {
     typealias Event = Request
     typealias Output = Response
```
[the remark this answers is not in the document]
tomerd: ditto

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -71,8 +65,8 @@ public struct LambdaInitializationContext: _AWSLambdaSendable {
 
 /// Lambda runtime context.
 /// The Lambda runtime generates and passes the `LambdaContext` to the Lambda handler as an argument.
-public struct LambdaContext: CustomDebugStringConvertible, _AWSLambdaSendable {
-    final class _Storage: _AWSLambdaSendable {
+public struct LambdaContext: CustomDebugStringConvertible, Sendable {
+    final class _Storage: Sendable {
```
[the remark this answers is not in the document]
tomerd: this PR does not change that aspect. are you saying it was a mistake originally? could `LambdaContext` be `Sendable` if the `_Storage` is not?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
tomerd: we could, I was debating this myself. I was also expecting infinite recursion but strangely it was not the case

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
tomerd: infinite recursion.thanks was also my exception but the tests in `LambdaHandlerTest` work fine. specifically see testBootstrapSuccess. testBootstrapFailure, testBootstrapInitSimple, testBootstrapNoInit. thoughts?

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
tomerd: this is internal, so should not be an issue

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
tomerd: this PR already changes the requirement to 5.7, unless I missed something?

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
tomerd: this was required to make the type system happy. do you have a better way to do this?

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
tomerd: ditto

Sources/AWSLambdaRuntimeCore/LambdaRuntime.swift
tomerd: this was required to remove the conformance to `ByteBufferLambdaHandler` from the other protocols, as otherwise it would require it. Do you have a better solution in mind?

Sources/AWSLambdaRuntimeCore/Terminator.swift
```diff
@@ -100,7 +100,7 @@ extension LambdaTerminator {
 
 extension LambdaTerminator {
     fileprivate final class Storage {
-        private let lock: Lock
+        private let lock: NIOLock
```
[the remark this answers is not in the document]
tomerd: sure

Examples/ErrorHandling/Lambda.swift
```diff
@@ -21,8 +21,6 @@ struct MyLambda: LambdaHandler {
     typealias Event = Request
     typealias Output = Response
```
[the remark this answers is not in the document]
tomerd: not any more now!

Examples/JSON/Lambda.swift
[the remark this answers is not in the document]
tomerd: not any more now!

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
tomerd: added `SimpleLambdaHandler` back

Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
tomerd: I think this is a good topic for a separate PR

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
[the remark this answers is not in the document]
fabianfett: Oh stupid me. The `Storage` has only `let` properties. We are exactly right here.

Sources/AWSLambdaRuntimeCore/Lambda+String.swift
```diff
@@ -13,25 +13,46 @@
 //===----------------------------------------------------------------------===//
 import NIOCore
 
-extension EventLoopLambdaHandler where Event == String {
+// MARK: - LambdaHandler String support
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
```
fabianfett, line 18: Do we just want to require macOS 12 and friends in the `Package.swift`? Will be a breaking change after 1.0...

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
fabianfett: It only changes this for all the examples. It also drops the 5.4 and 5.5 Package.swift. But the Package.swift without version only requires 5.6.

fabianfett: Also I think you are required to use the old generic syntax here, if we don't want to drop the types. Since we need create a connection between the input and the output type.

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
fabianfett: we don't need to drop here, if we keep the old syntax above.

Sources/AWSLambdaRuntimeCore/Lambda+String.swift
tomerd: sure

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
tomerd: > But the Package.swift without version only requires 5.6.

good catch

tomerd: > Also I think you are required to use the old generic syntax here, if we don't want to drop the types. Since we need create a connection between the input and the output type.

I don't think this can work with the old syntax. the Handler type is stored in LambdaRuntime which needs to loose it old generic signature, so they do not align any more

Sources/AWSLambdaRuntimeCore/Lambda+String.swift
hectormatos2011: This makes sense to me since most of the functionality I look to use is barred behind this anyways

tomerd:
@fabianfett now with your changes. thank you

benrosen78:
Hi there. Is there any update on when v1.0 will be released with the Swift async/await support? I am excited to use this in my software!

PR #274, opened 2022-12-02
Add .spi.yml for Swift Package Index DocC support

yim-lee:


.spi.yml
```diff
@@ -0,0 +1,4 @@
+version: 1
+builder:
+  configs:
+    - documentation_targets: [AWSLambdaRuntime]
```
yim-lee, line 4: Do we need to include `AWSLambdaRuntimeCore`?

tomerd: yes that is good idea

tomerd (approved):

PR #278, opened 2022-12-23
[plugin] add a deployer plugin + SAM example

sebsto:
This PR shows proof-of-concept code to add a deployer plugin, in addition to the existing archiver plugin. The deployer plugin generates a SAM deployment descriptor and calls the SAM command line to deploy the lambda function and it's dependencies.

## Motivation

The existing `archive` plugin generates a ZIP to be deployed on AWS. While it removes undifferentiated heavy lifting to compile and package Swift code into a Lambda function package, it does not help Swift developers to deploy the Lambda function to AWS, nor define how to invoke this function from other AWS services.  Deploying requires knowledge about AWS, and deployment tools such as the AWS CLI, the CDK, the SAM CLI, or the AWS console.

Furthermore, most developers will deploy a Lambda function together with some front end infrastructure allowing to invoke the Lambda function. Most common invocation methods are through an HTTP REST API (provided by API Gateway) or processing messages from queues (SQS).  This means that, in addition of the deployment of the lambda function itself, the Lambda function developer must create, configure, and link to the Lambda function an API Gateway or a SQS queue.

SAM is an open source  command line tool that solves this problem. It allows developers to describe the function runtime environment and the additional resources that will trigger the lambda function in a simple YAML file.  SAM CLI allows to validate the descriptor file and to deploy the infrastructure into the AWS cloud.

It also allows for local testing, by providing a Local Lambda runtime environment and a local API Gateway mock in a docker container.

The `deploy` plugin leverages SAM to create an end-to-end infrastructure and to deploy it on AWS.  It relies on  configuration provided by the Swift lambda function developer to know how to expose the Lambda function to the external world. Right now, it support a subset of HTTP API Gateway v2 and SQS queues.

The Lambda function developer describes the API gateway or SQS queue using the Swift programming language by writing a `Deploy.swift` file (similar to `Package.swift` used by SPM).  The plugin transform the `Deploy.swift` data structure into a SAM template. It then calls the SAM CLI to validate and to deploy the template.

## Modifications:

I added two targets to `Package.swift` : 

- `AWSLambdaDeployer` is the plugin itself. I followed the same structure and code as the `archive` plugin. Common code between the two plugins has been isolated in a shared `PluginUtils.swift` file. Because of [a limitation in the current Swift package systems for plugins](https://forums.swift.org/t/difficulty-sharing-code-between-swift-package-manager-plugins/61690/11), I symlinked the file from one plugin directory to the other.

- `AWSLambdaDeploymentDescriptor` is a shared library that contains the data structures definition to describe and to generate a JSON SAM deployment file. It models SAM resources such as a Lambda functions and its event sources : HTTP API and SQS queue. It contains the logic to generate the SAM deployment descriptor, using minimum information provided by the Swift lambda function developer. At the moment it provides a very minimal subset of the supported SAM configuration. I am ready to invest more time to cover more resource types and more properties if this proposal is accepted.

I added a new Example project : `SAM`. It contains two Lambda functions, one invoked through HTTP API, and one invoked through SQS.  It also defines shared resources such as SQS Queue and a DynamoDB Table. It provides a `Deploy.swift` example to describe the required HTTP API and SQS code and to allow `AWSLambdaDeploymentDescriptor` to generate the SAM deployment descriptor. The project also contains unit testing for the two Lambda functions.

## Result:

As a Swift function developer, here is the workflow to use the new `deploy` plugin.

1. I create a Lambda function as usual.  I use the Lambda Events library to write my code. Here is an example (nothing changed -  this is just to provide a starting point) : 

```swift
import AWSLambdaEvents
import AWSLambdaRuntime
import Foundation

@main
struct HttpApiLambda: SimpleLambdaHandler {

  init() {}
    context.logger.info(
      "Log Level env var : \(ProcessInfo.processInfo.environment["LOG_LEVEL"] ?? "info" )")
  }

  func handle(_ event: Event, context: AWSLambdaRuntimeCore.LambdaContext) async throws -> Output {

    var header = HTTPHeaders()
    do {

      let data = try JSONEncoder().encode(event)
      let response = String(data: data, encoding: .utf8)

      return Output(statusCode: .accepted, headers: header, body: response)

    } catch {
      header["content-type"] = "text/plain"
      return Output(statusCode: .badRequest, headers: header, body: "\(error.localizedDescription)")
    }
  }
}
```

2. I create a `Deploy.swift` file to describe the SAM deployment descriptor.  Most of the deployment descriptor will be generated automatically from context, I just have to provide the specifics for my code.  In this example, I want the Lambda function to be invoked from an HTTP REST API. I want the code to be invoked on `GET` HTTP method for the `/test` path. I also want to position the LOG_LEVEL environment variable to `debug`.

I add the new `Deploy.swift` file at the top of my project. Here is a simple deployment file. A more complex one is provided in the `Examples/SAM` sample project.

```swift
import AWSLambdaDeploymentDescriptor

let _ = DeploymentDefinition(
    
    functions: [        
        .function(
            name: "HttpApiLambda",
            architecture: .arm64, // optional, defaults to current build platform
            eventSources: [
                .httpApi(method: .GET, path: "/test"),
            ],
            environment: .variable(["LOG_LEVEL":"debug"]) //optional
        )
    ]
)
```

3. I add a dependency in my project's `Package.swift`. On a `testTarget`, I add this dependency:

```swift
  // on the testTarget()
  dependencies: [
      // other dependencies 
      .product(name: "AWSLambdaDeploymentDescriptor", package: "swift-aws-lambda-runtime")
  ]
```

I also might add this dependency on one of my Lambda functions `executableTarget`. In this case, I make sure it is added only when building on macOS.

```swift
  .product(name: "AWSLambdaDeploymentDescriptor", package: "swift-aws-lambda-runtime", condition: .when(platforms: [.macOS]))
```

4. I invoke the archive plugin and the deploy plugin from the command line.

```bash

swift build 

# first create the zip file
swift package --disable-sandbox archive

# second deploy it with an HTTP API Gateway
swift package --disable-sandbox deploy
```

Similarly to the archiver plugin, the deployer plugin must escape the sandbox because the SAM CLI makes network calls to AWS API (IAM and CloudFormation) to validate and to deploy the template.

5. (optionally) Swift lambda function developer may also use SAM to test the code locally.

```bash
sam local invoke -t sam.json -e Tests/LambdaTests/data/apiv2.json HttpApiLambda 
```

## Command Line Options

The deployer plugin accepts multiple options on the command line.

```bash
swift package plugin deploy --help
OVERVIEW: A swift plugin to deploy your Lambda function on your AWS account.
          
REQUIREMENTS: To use this plugin, you must have an AWS account and have `sam` installed.
              You can install sam with the following command:
              (brew tap aws/tap && brew install aws-sam-cli)

USAGE: swift package --disable-sandbox deploy [--help] [--verbose] [--nodeploy] [--configuration <configuration>] [--archive-path <archive_path>] [--stack-name <stack-name>]

OPTIONS:
    --verbose       Produce verbose output for debugging.
    --nodeploy      Generates the JSON deployment descriptor, but do not deploy.
    --nolist        Do not call sam list to list endpoints
    --configuration <configuration>
                    Build for a specific configuration.
                    Must be aligned with what was used to build and package.
                    Valid values : [ debug, release ] (default: debug)
    --archive-path <archive-path>
                    The path where the archive plugin created the ZIP archive.
                    Must be aligned with the value passed to archive --output-path.
                    (default: .build/plugins/AWSLambdaPackager/outputs/AWSLambdaPackager)
    --stack-name <stack-name>
                    The name of the CloudFormation stack when deploying.
                    (default: the project name)
    --help          Show help information.
```

#### SAM

SAM is already broadly adopted, well maintained and documented. It does the job.  I think it is easier to ask Swift Lambda function developers to install SAM (it is just two `brew` commands) rather than having this project investing in its own mechanism to describe a deployment and to generate the CloudFormation or CDK code to deploy the Lambda function and its dependencies. In the future, we might imagine a multi-framework solution where the plugin could generate code for SAM, or CDK, or Serverless etc ... 

#### Deploy.swift

Swift Lambda function developers must be able to describe the additional infrastructure services required to deploy their functions: a SQS queue, an HTTP API etc.

I assume the typical Lambda function developer knows the Swift programming language, but not the AWS-specific DSL (such as SAM or CloudFormation) required to describe and deploy the project dependencies. I chose to ask the Lambda function developer to describe its deployment with a Swift struct in a top-level `Deploy.swift` file. The `deploy` plugin dynamically compiles this file to generate the SAM JSON deployment descriptor.

The source code to implement this approach is in the `AWSLambdaDeploymentDescriptor` library. This approach is similar to `Package.swift` used by the Swift Package Manager.

This is a strong design decision and [a one-way door](https://shit.management/one-way-and-two-way-door-decisions/). It engages the maintainer of the project on the long term to implement and maintain (close) feature parity between SAM DSL and the Swift `AWSLambdaDeploymentDescriptor` library.

One way to mitigate the maintenance work would be to generate the `AWSLambdaDeploymentDescriptor` library automatically, based on the [the SAM schema definition](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/validator/sam_schema/schema.json). The core structs might be generated automatically and we would need to manually maintain only a couple of extensions providing syntactic sugar for Lambda function developers. This approach is similar to AWS SDKs code generation ([Soto](https://github.com/soto-project/soto-codegenerator) and the [AWS SDK for Swift](https://github.com/awslabs/aws-sdk-swift/tree/main/codegen)). This would require a significant one-time engineering effort however and I haven't had time to further explore this idea.

**Alternatives Considered** 

One alternative proposed during early reviews is to use a DSL instead of a programmatic approach. A Swift-based DSL might be implemented using [Result Builders](https://github.com/apple/swift-evolution/blob/main/proposals/0289-result-builders.md), available since Swift 5.4.
After having read [tutorials](https://theswiftdev.com/result-builders-in-swift/) and watch a [WWDC 2021 session](https://developer.apple.com/videos/play/wwdc2021/10253/), I understand this might provide Swift Lambda function developer a strongly typed, Swift-friendly, DSL to express their SAM deployment descriptor.  I will give a try to this approach and propose it in a PR.

Another alternative is to not use a programmatic approach to describe the deployment at all (i.e. remove `Deploy.swift` and the `AWSLambdaDeploymentDescriptor` from this PR). In this scenario, the `deploy` plugin would generate a minimum SAM deployment template with default configuration for the current Lambda functions in the build target. The plugin would accept command-line arguments for basic pre-configuration of dependant AWS services, such as `--httpApi` or `--sqs <queue_name>` for example. The Swift Lambda function developer could leverage this SAM template to provide additional infrastructure or configuration elements as required. After having generated the initial SAM template, the `deploy` plugin will not overwrite the changes made by the developer.

This approach removes the need to maintain feature parity between the SAM DSL and the `AWSLambdaDeploymentDescriptor` library.

Please comment on this PR to share your feedback about the current design decisions and the proposed alternatives (or propose other alternatives :-) ) 

### What is missing

If this proposal is accepted in its current format, Swift Lambda function developers would need a much larger coverage of the SAM template format. I will add support for resources and properties. We can also look at generating the Swift data structures automatically from the AWS-provided SAM schema definition (in JSON)

### Future directions 

Here are a list of todo and thoughts for future implementations.

- Both for the `deploy` and the `archive` plugin, it would be great to have a more granular permission mechanism allowing to escape the SPM plugin sandbox for selected network calls. SPM 5.8 should make this happen.

- For HTTPApi, I believe the default SAM code and Lambda function examples must create Authenticated API by default. I believe our duty is to propose secured code by default and not encourage bad practices such as deploying open endpoints. But this approach will make the initial developer experience a bit more complex.

- This project should add sample code to demonstrate how to use the Soto SDK or the AWS SDK for Swift. I suspect most Swift Lambda function will leverage other AWS services.

- What about bootstrapping new projects? I would like to create a plugin or command line tool that would scaffold a new project, create the `Package.swift` file and the required project directory and files.  We could imagine a CLI or SPM plugin that ask the developer a couple of questions, such as how she wants to trigger the Lambda function and generate the corresponding code.

---

Happy to read your feedback and suggestions. Let's make the deployment of Swift Lambda functions easier for Swift developers.


Package.swift
```diff
@@ -49,6 +54,23 @@ let package = Package(
                 )
             )
         ),
+        .target(
+            name: "AWSLambdaDeploymentDescriptor",
+            dependencies: [
+                .product(name: "Yams", package: "Yams")
+            ],
+            path: "Sources/AWSLambdaDeploymentDescriptor"
```
tomerd, line 62: redundant?

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+        if configuration.verboseLogging {
+            print("-------------------------------------------------------------------------")
+            print("executables")
+            print("-------------------------------------------------------------------------")
+            print("SAM Executable : \(samExecutableUrl)")
+            print("Deployment Descriptor Executable : \(deploymentDescriptorExecutableUrl)")
+        }
+
+        let currentDirectory = FileManager.default.currentDirectoryPath
+        let samDeploymentDescriptorUrl = URL(fileURLWithPath: currentDirectory)
```
tomerd, line 40: not sure we need a URL here. we can probably use Path from the PackagePlugin API

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+struct AWSLambdaPackager: CommandPlugin {
+    func performCommand(context: PackagePlugin.PluginContext, arguments: [String]) async throws {
+
+        let configuration = try Configuration(context: context, arguments: arguments)
+        guard !configuration.products.isEmpty else {
+            throw Errors.unknownProduct("no appropriate products found to deploy")
+        }
+
+        let samExecutable = try context.tool(named: "sam")
+        let samExecutableUrl = URL(fileURLWithPath: samExecutable.path.string)
```
tomerd, line 29: not sure we need a URL here. we can probably use Path from the PackagePlugin API

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+        let deploymentDescriptorExecutableUrl = configuration.deployExecutable
```
tomerd, line 30: do you think there is an opportunity to do this via configuration instead of code? if so, we have another type of plugin that helps with source generation. 

if we stay with the approach that the user must create a module named in specific way, we should probably add validation step to check if the executable (deploymentDescriptorExecutableUrl) exists on disk

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+                print("Run the command again with --verbose argument to receive more details.")
+            }
+        } catch {
+            print("Can not execute file at:")
+            print("\(deploymentDescriptorExecutableUrl.path)")
+            print("or at:")
+            print("\(samExecutableUrl.path)")
+            print("Is SAM installed ? (brew tap aws/tap && brew install aws-sam-cli)")
+            print("Did you add a 'Deploy' executable target into your project's Package.swift ?")
+            print("Did you build the release version ? (swift build -c release)")
```
tomerd, line 85: we can probably do some guard checks upfront to see if these exist on disk and throw more fine tuned error messages

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+
+    @discardableResult
+    private func execute(
+        executable: URL,
```
tomerd, line 91: I dont think we need to use a URL anywhere. Path has most operations needed? What did you find missing?

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+        arguments: [String],
+        customWorkingDirectory: URL? = nil,
+        logLevel: ProcessLogLevel
+    ) throws -> String {
+        try self.execute(
+            executable: Path(executable.path),
+            arguments: arguments,
+            customWorkingDirectory: customWorkingDirectory == nil
```
tomerd, line 99: flatMap

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+            ? nil : Path(customWorkingDirectory!.path),
+            logLevel: logLevel)
+    }
+
+    // **************************************************************
+    // Below this line, the code is copied from the archiver plugin
+    // **************************************************************
+    @discardableResult
+    private func execute(
```
tomerd, line 108: I assume this is copy/paste from the other plugin? maybe we can combine them to a single implementation?

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+        let arch = "x86_64-apple-macosx"
+#endif
+        var deployExecutableURL = URL(fileURLWithPath: deployProducts[0].targets[0].directory.string)
+        // while there is /Sources or /Deploy => remove last path component 
+        while deployExecutableURL.path.contains("/Sources") ||
+              deployExecutableURL.path.contains("/Deploy") {
+
+            deployExecutableURL = deployExecutableURL.deletingLastPathComponent()
+        }
+        self.deployExecutable = deployExecutableURL.appendingPathComponent(".build/\(arch)/\(self.buildConfiguration)/Deploy")
```
tomerd, line 270: this seems a bit fragile. what information is it trying to obtain? SwiftPM can return the executables path if needed

Plugins/AWSLambdaDeployer/Plugin.swift
```diff
@@ -0,0 +1,348 @@
...
+            return "\(arguments.joined(separator: " ")) failed with code \(code)"
+        }
+    }
+}
+
+// **************************************************************
+// Below this line, the code is copied from the archiver plugin
+// **************************************************************
+
+private enum ProcessLogLevel: Comparable {
```
tomerd, line 321: make DRYer with other plugin?

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -326,7 +326,7 @@ private struct Configuration: CustomStringConvertible {
             self.products = products
 
         } else {
-            self.products = context.package.products.filter { $0 is ExecutableProduct }
+            self.products = context.package.products.filter { $0 is ExecutableProduct && $0.name != "Deploy" }
```
tomerd, line 329: this seems fragile. what other options do we have?

Sources/AWSLambdaDeploymentDescriptor/DeploymentDefinition.swift
```diff
@@ -0,0 +1,526 @@
...
+public protocol SAMResource: Encodable {}
+public protocol SAMResourceProperties: Encodable {}
+
+public struct Resource: SAMResource, Equatable {
+    
+    let type: String
+    let properties: SAMResourceProperties
+    let name: String
+        
+    public static func none() -> [Resource] { return [] }
```
tomerd, line 66: variable instead of func?

Sources/AWSLambdaDeploymentDescriptor/DeploymentDefinition.swift
```diff
@@ -0,0 +1,526 @@
...
+    let codeUri: String
+    let autoPublishAlias: String
+    var eventSources: [String: EventSource]
+    var environment: EnvironmentVariable
+    
+    public init(codeUri: String,
+                eventSources: [EventSource] = [],
+                environment: EnvironmentVariable = EnvironmentVariable.none()) {
+        
+        #if arch(arm64) //when we build on Arm, we deploy on Arm
```
tomerd, line 125: this is probably okay for now, but may not always be true

Sources/AWSLambdaDeploymentDescriptor/DeploymentDefinition.swift
```diff
@@ -0,0 +1,526 @@
...
+    
+    public var variables: [String:EnvironmentVariableValue] = [:]
+    public init() {}
+    public init(_ variables: [String:String]) {
+        for key in variables.keys {
+            self.variables[key] = .string(value: variables[key] ?? "")
+        }
+    }
+    public static func none() -> EnvironmentVariable { return EnvironmentVariable([:]) }
+    public func isEmpty() -> Bool { return variables.count == 0 }
```
tomerd, line 182: var instead of func?

Sources/AWSLambdaDeploymentDescriptor/DeploymentDefinition.swift
```diff
@@ -0,0 +1,526 @@
...
+public struct EnvironmentVariable: Encodable {
+    
+    public var variables: [String:EnvironmentVariableValue] = [:]
+    public init() {}
+    public init(_ variables: [String:String]) {
+        for key in variables.keys {
+            self.variables[key] = .string(value: variables[key] ?? "")
+        }
+    }
+    public static func none() -> EnvironmentVariable { return EnvironmentVariable([:]) }
```
tomerd, line 181: var instead of func?

Sources/AWSLambdaDeploymentDescriptor/DeploymentDescriptor.swift
```diff
@@ -0,0 +1,132 @@
...
+    }
+    
+    // returns additional resources to create in the cloud (ex : a dynamoDB table)
+    // This is a default implementation to avoid forcing the lambda developer to implement it when not needed
+    public func addResource() -> [Resource] {
+        return Resource.none()
+    }
+    
+    // entry point and main function. Lambda developer must call this function.
+    public func run() {
```
tomerd, line 68: can we use an @main based protocol instead?

Sources/AWSLambdaDeploymentDescriptor/DeploymentDescriptor.swift
```diff
@@ -0,0 +1,132 @@
...
+        } catch {
+            throw DeploymentEncodingError.yamlError(causedBy: error)
+        }
+    }
+
+    private func createServerlessFunctionResource(_ name: String) -> [Resource] {
+        
+        // the default output dir for archive plugin
+        // FIXME: add support for --output-path option on packager
+        let package = ".build/plugins/AWSLambdaPackager/outputs/AWSLambdaPackager/\(name)/\(name).zip"
```
tomerd, line 106: this seems a bit fragile. what other options do we have?

Sources/AWSLambdaDeploymentDescriptor/DeploymentDescriptor.swift
```diff
@@ -0,0 +1,132 @@
...
+        let lambdaEventSources = eventSources(name)
+        
+        // When SQS event source is specified, the Lambda function developer
+        // might give a queue name, a queue Arn, or a queue resource.
+        // When developer gives a queue name or a queue resource,
+        // the event source eventually creates the queue resource and references the resource.
+        // Now, we need to collect all queue resources created by SQS event sources or passed by Lambda function develper
+        // to add them to the list of resources to synthetize
+        var resources : [Resource] = lambdaEventSources.filter{ lambdaEventSource in
+            lambdaEventSource.type == "SQS" &&
```
tomerd, line 118: can we make this more type safe?

Examples/SAM/Package.swift
```diff
@@ -0,0 +1,62 @@
...
+    .executable(name: "Deploy", targets: ["Deploy"])
+  ],
+  dependencies: [
+    // this is the dependency on the swift-aws-lambda-runtime library
+    // in real-world projects this would say
+    // .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", branch: "main"),
+    .package(name: "swift-aws-lambda-runtime", path: "../.."),
+    .package(url: "https://github.com/swift-server/swift-aws-lambda-events.git", branch: "main")
+  ],
+  targets: [
```
tomerd, line 37: should we include some testing examples?

tomerd (commented):
thank you so much for this PR @sebsto, this could be very helpful for users

Left a few ideas and comments to discuss further.

sebsto:
Thank you @tomerd for the review.
I agree the code must be more robust. I will propose another design that do not rely on an executable target and finding it's directory at runtime.

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: I removed all unnecessary usage of URL.  I still use a few when I use `FileManager` or `String` functions that expect an `URL` argument

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: `URL` is required here because I use `String.write(to:encoding:)` to save the generated deployment descriptor

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: Agree.  I will propose a simplified approach where function developer must just provide `Deploy.swift` file, similar to SPM's `Package.swift`.

The plugin will compile and run this file to generate the SAM deployment descriptor.

That will remove the need for the function developer to provide an executable target.  The declarative nature of Deploy.swift will be more familiar and less error prone.

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: done. With three `do {} catch {}` blocks instead of one and one additional `guard else` statement

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: done, I removed unnecessary usage of URL

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: sorry, I am not sure to understand where you suggest to use `flatMap()` in the 4 lines highlighted above

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: Initially I did not want to touch the Archiver plugin.  Now that I know you like the idea of the deploy plugin, I will factor out common codes in a separate shared library

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: I will factor out code common to the two plugin in a shared library

Sources/AWSLambdaDeploymentDescriptor/DeploymentDescriptor.swift
sebsto: Agreed. This is not needed anymore with the `Deploy.swift` declarative approach

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: I wonder what is the best approach to share code between two Plugins ?  Share a common executable to carry on the actual work ? Symlinks ?

Check  this thread https://forums.swift.org/t/difficulty-sharing-code-between-swift-package-manager-plugins/61690/11

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: I wonder what is the best approach to share code between two Plugins ?  Share a common executable to carry on the actual work ? Symlinks ?

Check  this thread https://forums.swift.org/t/difficulty-sharing-code-between-swift-package-manager-plugins/61690/11

Plugins/AWSLambdaDeployer/Plugin.swift
tomerd: yea this is a "hole" in SwiftPM plugins that the team is working to fix afaik, the work around is to put it in a file and symlink it across the targets

Plugins/AWSLambdaDeployer/Plugin.swift
tomerd: yea this is a "hole" in SwiftPM plugins that the team is working to fix afaik, the work around is to put it in a file and symlink it across the targets

Plugins/AWSLambdaDeployer/Plugin.swift
tomerd: does this `customWorkingDirectory.flatMap { Path($0) }` work?

Plugins/AWSLambdaDeployer/Plugin.swift
tomerd: `String.write(to:encoding:)` is syntactic sugar for https://developer.apple.com/documentation/foundation/filemanager/1410695-createfile which takes strings

sebsto:
I just pushed a new version of the deployer plugin.

Now it uses a declarative interface to describe the deployment instead of a programmatic one. 
The example is in `Examples/SAM` directory.  The lambda function developer now describes the deployment by writing a `Deploy.swift` that will be compile and run by the `deploy` plugin.  This is an approach similar to the one used by SPM and `Package.swift`

**TODO**
- clear TODOs in the code 
- find a permanent solution for  a ugly workaround for a problem I do not fully understand (Line 20 of Examples/SAM/Package.swift)
- write tests for the Lambda function 
- update README at the start of this thread 
- find a way to share common code between the two plugins

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: OK - will do that

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: OK - will do this

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: Done with https://github.com/sebsto/swift-aws-lambda-runtime/commit/8a816c5f34d3aefd4d299b6cde42efa651ad9c02

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: Done with https://github.com/sebsto/swift-aws-lambda-runtime/commit/8a816c5f34d3aefd4d299b6cde42efa651ad9c02

Plugins/AWSLambdaDeployer/Plugin.swift
sebsto: done https://github.com/sebsto/swift-aws-lambda-runtime/commit/d09ecbe35bc8cb2c72f3afffd5dec4d3100f87dd

Sources/AWSLambdaDeploymentDescriptor/DeploymentDefinition.swift
sebsto: I made the architecture a parameter in `Deploy.swift`, with default to the current architecture.
https://github.com/sebsto/swift-aws-lambda-runtime/commit/f3f4021dbebb086f4f33efac58b3b4e6a5cf4d1e

sebsto:
Here is a new version, using a declarative mode to describe the deployment, inspired by `Package.swift`. It also addresses all the comments and suggestions made at the first review

Examples/SAM/Package.swift
sebsto: Done https://github.com/sebsto/swift-aws-lambda-runtime/commit/464fd01bb4d10060c4bb19f0f42fcac5270eeae7

sebsto:
Here is a new version, using a declarative mode to describe the deployment, inspired by `Package.swift`. It also addresses all the comments and suggestions made at the first review. This is ready for a second review. Thank you.

sebsto:
After further design discussions, @tomerd suggested to use a DSL-based approach to describe the deployment descriptor in the Swift language.  A SAM DSL might be defined using the [result builder library](https://github.com/apple/swift-evolution/blob/main/proposals/0289-result-builders.md)

I will submit more commit to implement this approach.

sebsto:
Closing this PR in favor of https://github.com/swift-server/swift-aws-lambda-runtime/pull/291
The new PR implements a Swift-based DSL approach to describe and generate the deployment descriptor.

PR #277, opened 2022-12-13
Rearrange Examples/LocalDebugging package structure to work with archive plugin

sebsto:
I propose to re-arrange the file and package structure in `Examples/LocalDebugging` to make it work with the `archive` plugin.

### Motivation:

The current directory and package structure in the `Examples/LocalDebugging` example does not work out of the box with the `archive` plugin. 
See https://github.com/swift-server/swift-aws-lambda-runtime/issues/276 

### Modifications:

I re-arranged the package structure to share the same `Package.swift` for the targets `MyLambda` and `Shared`.  This allows the plugin to build and archive MyLamba.  I also made minimum changes to the Xcode workspace to continue to be able to build and debug from Xcode.

Let me know if there is a better way to arrange package and file in this project to work with : Xcode builds, command line builds, and the `archive` plugin.

### Result:

Project now builds with `swift build`
It runs with  `export LOCAL_LAMBDA_SERVER_ENABLED=true  && swift run MyLambda`. 
It packages with `swift package --disable-sandbox archive`
It runs by selecting the `MyLambda` target in Xcode
It runs by selecting the `MyApp` target in Xcode

Issue #276, opened 2022-12-13
[plugin] LocalDebugging example - archive fails because of Shared code


sebsto:
### Expected behavior
I expect the archive plugin to package the `MyLambda` function part of the `Examples/Localdebugging` 
The archive command fails because part of the `MyLambda` function relies on `Shared` code which is not under `MyLambda` directory.

### Actual behavior
The archive fails because the Lambda function relies on shared code outside of the `MyLambda` directory structure.

```
 swift package --disable-sandbox archive --verbose 1 --output-path dist                    
-------------------------------------------------------------------------
configuration
-------------------------------------------------------------------------
{
  outputDirectory: dist
  products: ["MyLambda"]
  buildConfiguration: release
  baseDockerImage: swift:amazonlinux2
}
-------------------------------------------------------------------------
building "mylambda" in docker
-------------------------------------------------------------------------
updating "swift:amazonlinux2" docker image
  amazonlinux2: Pulling from library/swift
  Digest: sha256:0b645a7589aa5f1759f976c42a9ffba48fd930b27b3481b0ab32e271e701d444
  Status: Image is up to date for swift:amazonlinux2
  docker.io/library/swift:amazonlinux2
/usr/local/bin/docker run --rm -v /Users/stormacq/Documents/amazon/code/lambda/swift/swift-aws-lambda-runtime/Examples/LocalDebugging/MyLambda:/workspace -w /workspace swift:amazonlinux2 bash -cl swift build -c release --show-bin-path
  /workspace/.build/aarch64-unknown-linux-gnu/release
building "MyLambda"
/usr/local/bin/docker run --rm -v /Users/stormacq/Documents/amazon/code/lambda/swift/swift-aws-lambda-runtime/Examples/LocalDebugging/MyLambda:/workspace -w /workspace swift:amazonlinux2 bash -cl swift build -c release --product MyLambda --static-swift-stdlib
  error: the package at '/Shared' cannot be accessed (Error Domain=NSCocoaErrorDomain Code=260 "The file doesn’t exist.")
error: /usr/local/bin/docker run --rm -v /Users/stormacq/Documents/amazon/code/lambda/swift/swift-aws-lambda-runtime/Examples/LocalDebugging/MyLambda:/workspace -w /workspace swift:amazonlinux2 bash -cl swift build -c release --product MyLambda --static-swift-stdlib failed with code 1
```

### Steps to reproduce

[...]

PR #292, opened 2023-03-10
improve examples

tomerd:
motivation: examples can be confusing since they use relative path to the library for CI purposes

changes:
* update examples to use the library URL, expect when env variable is set for CI purposes
* rename docker compose job to test-examples since it is more accurate

### DIFF
--- a/Examples/Benchmark/Package.swift
+++ b/Examples/Benchmark/Package.swift
@@ -1,5 +1,6 @@
 // swift-tools-version:5.7
 
+import class Foundation.ProcessInfo // needed for CI to test the local version of the library
 import PackageDescription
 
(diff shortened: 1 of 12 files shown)


tomerd:

fabianfett (approved):
LGTM. Though I wish PackageDescription would allow some kind of env/setting lookup :). The Foundation import always gives me the chills ;)

sebsto:
This change does not allow to use the archive plugin for the local debugging example 
See https://github.com/swift-server/swift-aws-lambda-runtime/issues/276
and https://github.com/swift-server/swift-aws-lambda-runtime/pull/277 for a possible fix

PR #308, opened 2023-09-27
Allow custom initialisation of the HandlerType of the LambdaRuntime.

tachyonics:
Allow custom initialisation of the HandlerType of the LambdaRuntime.

### Motivation:

Provide the flexibility for custom initialisation of the HandlerType as this will often be required by higher level frameworks.

### Modifications:

Modify the `LambdaRuntime` type to accept a closure to provide the handler rather than requiring that it is provided by a static method on the Handler type. This is achieved by a new base protocol `NonFactoryByteBufferLambdaHandler` that does not have the `makeHandler` function requirement.

Also provide a new base protocol `NonFactoryLambdaHandler` that doesn't have the initialisation requirements of either `LambdaHandler` or `SimpleLambdaHandler`. This will allow higher level frameworks to initialise their own handlers that however can take advantage of the `Codable` conveniences of this package.

```swift
@Sendable
func handlerProvider(context: LambdaInitializationContext) 
-> EventLoopFuture<some NonFactoryByteBufferLambdaHandler> {
    let underlyingHandler = <initialise a type conforming to NonFactoryLambdaHandler>
    let codableHandler = underlyingHandler.withWrappingCodableHandler(allocator: context.allocator)

    return context.eventLoop.makeSuceededFuture(codableHandler)
}
```

### Result:

Existing high level use cases (`LambdaHandler` and `SimpleLambdaHandler`) will continue to be provided but a lower level integration will be possible aimed at higher level frameworks.

### DIFF
--- a/Sources/AWSLambdaRuntime/Lambda+Codable.swift
+++ b/Sources/AWSLambdaRuntime/Lambda+Codable.swift
@@ -19,10 +19,10 @@ import class Foundation.JSONEncoder
 import NIOCore
 import NIOFoundationCompat
 
-// MARK: - SimpleLambdaHandler Codable support
+// MARK: - NonFactoryLambdaHandler Codable support
 
 /// Implementation of `ByteBuffer` to `Event` decoding.
-extension SimpleLambdaHandler where Event: Decodable {
+extension NonFactoryLambdaHandler where Event: Decodable {
     @inlinable
     public func decode(buffer: ByteBuffer) throws -> Event {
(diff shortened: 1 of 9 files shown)


tomerd:
thanks for putting together @tachyonics, I finally had time to review and came up with a slightly more minimal change in https://github.com/swift-server/swift-aws-lambda-runtime/pull/310. If the minimalistic approach is good enough for your use case we can update #308 to that end and close #310 or vice versa.

tachyonics:
Closing as this has been replaced by #310

PR #310, opened 2023-10-10
allow custom initialization of the HandlerType of the LambdaRuntime

tomerd:
Motivation:  Provide the flexibility for custom initialization of the HandlerType as this will often be required by higher level frameworks.

Modifications:
* Modify the LambdaRuntime type to accept a closure to provide the handler rather than requiring that it is provided by a static method on the Handler type
* Update downstream code to use HandlerProvider
* Update upstream code to support passing Hanlder Type of Handler Provider
* Add and update tests

Originally suggested and coded by @tachyonics in https://github.com/swift-server/swift-aws-lambda-runtime/pull/308

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -89,18 +89,15 @@ public enum Lambda {
         Self.run(configuration: configuration, handlerProvider: handlerType.makeHandler(context:))
     }
 
-    /// Run a Lambda defined by implementing the ``ByteBufferLambdaHandler`` protocol.
-    /// The Runtime will manage the Lambdas application lifecycle automatically. It will invoke the
-    /// ``ByteBufferLambdaHandler/makeHandler(context:)`` to create a new Handler.
-    ///
+    /// Run a Lambda defined by implementing the ``LambdaRuntimeHandler`` protocol.
     /// - parameters:
     ///     - configuration: A Lambda runtime configuration object
-    ///     - handlerProvider: A provider of the Handler to invoke.
+    ///     - handlerProvider: A provider of the ``LambdaRuntimeHandler``` to invoke.
     ///
     /// - note: This is a blocking operation that will run forever, as its lifecycle is managed by the AWS Lambda Runtime Engine.
     internal static func run(
         configuration: LambdaConfiguration = .init(),
-        handlerProvider: @escaping (LambdaInitializationContext) -> EventLoopFuture<some NonInitializingByteBufferLambdaHandler>
+        handlerProvider: @escaping (LambdaInitializationContext) -> EventLoopFuture<some LambdaRuntimeHandler>
     ) -> Result<Int, Error> {
         let _run = { (configuration: LambdaConfiguration) -> Result<Int, Error> in
             #if swift(<5.9)
--- a/Sources/AWSLambdaRuntimeCore/Lambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda.swift
@@ -92,7 +92,7 @@ public enum Lambda {
     /// Run a Lambda defined by implementing the ``LambdaRuntimeHandler`` protocol.
     /// - parameters:
     ///     - configuration: A Lambda runtime configuration object
-    ///     - handlerProvider: A provider of the ``LambdaRuntimeHandler``` to invoke.
+    ///     - handlerProvider: A provider of the ``LambdaRuntimeHandler`` to invoke.
     ///
     /// - note: This is a blocking operation that will run forever, as its lifecycle is managed by the AWS Lambda Runtime Engine.
     internal static func run(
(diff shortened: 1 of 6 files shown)


tomerd:
@tachyonics wdyt?

tachyonics:
> @tachyonics wdyt?

This will work for our use case although not quite as convenient as what I had. My only concern is that `handlerProvider` returns a `ByteBufferLambdaHandler` which has a `makeHandler` function requirement that will not actually get used. In our code we can just not implement that method but conceptually it seems a bit odd at the runtime level.

tomerd:
@tachyonics wdyt about https://github.com/swift-server/swift-aws-lambda-runtime/pull/310/commits/792d1c1adb0f49704a28f9093e87b618d48b5f5b? I don love the name `CoreByteBufferLambdaHandler` but hopefully this address the need better?

tachyonics (approved):
@tomerd I am not really a fan of the name either but otherwise the changes look good and are more consistent.

tomerd:
thanks @tachyonics, so now to the hard part - choosing a name.

cc @fabianfett @sebsto @yim-lee @ktoso

any good ideas?

tachyonics:
> so now to the hard part - choosing a name

@tomerd `NonInitializingByteBufferLambdaHandler`?

tomerd:
@tachyonics updated to `NonInitializingByteBufferLambdaHandler`. wdyt?

Sources/AWSLambdaRuntimeCore/LambdaRuntime.swift
```diff
@@ -226,8 +267,79 @@ public enum LambdaRuntimeFactory {
...
+            handlerProvider: Handler.makeHandler(context:),
+            eventLoop: eventLoop,
+            logger: logger
+        )
+    }
+
+    /// Create a new `LambdaRuntime`.
+    ///
+    /// - parameters:
+    ///     - handlerProvider: A provider of the ``CoreByteBufferLambdaHandler``  the `LambdaRuntime` will manage.
```
tachyonics, line 304: Should this be `NonInitializingByteBufferLambdaHandler`?

Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -85,6 +85,22 @@ public enum Lambda {
     internal static func run(
         configuration: LambdaConfiguration = .init(),
         handlerType: (some ByteBufferLambdaHandler).Type
+    ) -> Result<Int, Error> {
+        Self.run(configuration: configuration, handlerProvider: handlerType.makeHandler(context:))
+    }
+
+    /// Run a Lambda defined by implementing the ``ByteBufferLambdaHandler`` protocol.
```
tachyonics, line 92: Should this be `NonInitializingByteBufferLambdaHandler`?

Sources/AWSLambdaRuntimeCore/LambdaRuntime.swift
```diff
@@ -35,18 +37,41 @@ public final class LambdaRuntime<Handler: ByteBufferLambdaHandler> {
...
+            logger: logger,
+            configuration: .init()
+        )
     }
 
-    init(handlerType: Handler.Type, eventLoop: EventLoop, logger: Logger, configuration: LambdaConfiguration) {
+    /// Create a new `LambdaRuntime`.
+    ///
+    /// - parameters:
+    ///     - handlerProvider: A provider of the ``ByteBufferLambdaHandler``  the `LambdaRuntime` will manage.
```
tachyonics, line 60: Same as the other comments.

tachyonics:
Just some small comments about comments. Otherwise lgtm.

tomerd:
@tachyonics new name idea: `LambdaRuntimeHandler`. wdyt?

Sources/AWSLambdaRuntimeCore/Lambda.swift
```diff
@@ -85,6 +85,19 @@ public enum Lambda {
...
         configuration: LambdaConfiguration = .init(),
         handlerType: (some ByteBufferLambdaHandler).Type
+    ) -> Result<Int, Error> {
+        Self.run(configuration: configuration, handlerProvider: handlerType.makeHandler(context:))
+    }
+
+    /// Run a Lambda defined by implementing the ``LambdaRuntimeHandler`` protocol.
+    /// - parameters:
+    ///     - configuration: A Lambda runtime configuration object
+    ///     - handlerProvider: A provider of the ``LambdaRuntimeHandler``` to invoke.
```
tachyonics, line 95: Extra tick...

tomerd:
@fabianfett any concerns?

Sources/AWSLambdaRuntimeCore/Lambda.swift
tomerd:
```suggestion
    ///     - handlerProvider: A provider of the ``LambdaRuntimeHandler`` to invoke.
```

fabianfett:
Tbh. I think this doesn't solve the actual issue. The lifetime handling in Lambda was designed and implemented before ServiceLifecycle came about. For this reason, using long running clients in Lambda currently feels absolutely weird.

For this reason, I think, we should reconsider what the Lambda API should look like taking into account that it should play nicely with ServiceLifecycle.

General ideas:

1. `LambdaRuntime` should implement `Service`
2. If a Lambda runs locally it should set up a `ServiceGroup` to handle signal handling
3. If a user needs long running Services within their Lambda they should be able to use ServiceGroups to manage their lifetimes.

PR #318, opened 2024-02-01
[plugin] Feature: Build docker image

tadelv:
# Add option to PackagePlugin, to build image before building Lambda package

### Motivation:

I have a Swift Lambda that has dependencies that require the base docker image to have openssl-dev installed. The default implementation did not work for me - I must have missed something somewhere. Therefore I added another optional step to the packaging plugin to also build the image before the compilation step.

### Modifications:

Added two flags, `--build-docker-image` and `--docker-image-platform`, which when set will run an additional build step using `docker-buildx` and use the `Dockerfile` in the project root to build the image

### Result:

When additional arguments are passed to `swift package archive` and a Dockerfile is present, the plugin will build the image before compiling the Lambda

### DIFF
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -105,8 +105,8 @@ struct AWSLambdaPackager: CommandPlugin {
         if buildDockerImage {
             try self.execute(
               executable: dockerToolPath.removingLastComponent().appending(subpath: "docker-buildx"),
-              arguments: ["--platform=\(buildDockerImagePlatform)", ".", "-t", baseImage],
-              logLevel: .debug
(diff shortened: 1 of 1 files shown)


tomerd:
thanks @tadelv, thanks for the suggestion and PR. the original design is that you can provide a `baseImage` which the plugin would use. this could be a vanilla image or one that you created yourself for example with additional system dependencies. building a custom base image is in principal a one-time thing, so wrapping that in the plugin does not seem to add a huge amount of value compared to just running the docker commands on your own and making the image available for future use. thoughts?

tadelv:
Hi @tomerd, thank you for chiming in! Like I mentioned before, I was sure I missed something :) I'm not familiar enough with Docker to have thought of using a different prebuilt base image. Thank you, this seems like a much better way. 

Do I close the PR?

PR #322, opened 2024-04-15
Avoid warning when enabling strict concurrency check

sebsto:
Avoid strict concurrency check warnings

### Motivation:

My Lambda function is an actor to protect instances variables from race condition.
```
@main
actor MyLambda: LambdaHandler { ... }
```
However, when I compile with
```
swift build -Xswiftc -strict-concurrency=complete
```
I receive this warning
```
warning: non-sendable type 'Self.Event' in parameter of the protocol requirement satisfied by actor-isolated instance method 'handle(_:context:)' cannot cross actor boundary
    func handle(_ event: APIGatewayV2Request, context: AWSLambdaRuntimeCore.LambdaContext) async throws -> APIGatewayV2Response {
```

What I understand from the error message is that `Event` must be `Sendable`. But `Event` is an associated type on LambdaHandler, see 
https://github.com/swift-server/swift-aws-lambda-runtime/blob/main/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift#L146

In this case, it is an `APIGatewayV2Request` which correctly implements `Sendable` but it looks like the compiler can not detect that.

### Modifications:

I changed all occurences of `associatedtype Event` and `associatedtype Output` in `LambdaHandler.swift` to add `:Sendable`.
I did not check for Swift >=5.6 because I understand Swift 5.7 is the minimum version to compile the runtime.

### Result:

No compiler warnings when using `Xswiftc -strict-concurrency=complete`

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaHandler.swift
@@ -27,9 +27,9 @@ public protocol SimpleLambdaHandler {
     /// The lambda function's input. In most cases this should be `Codable`. If your event originates from an
     /// AWS service, have a look at [AWSLambdaEvents](https://github.com/swift-server/swift-aws-lambda-events),
     /// which provides a number of commonly used AWS Event implementations.
-    associatedtype Event
+    associatedtype Event: Sendable
     /// The lambda function's output. Can be `Void`.
-    associatedtype Output
+    associatedtype Output: Sendable
 
     init()
 
@@ -143,9 +143,9 @@ public protocol LambdaHandler {
     /// The lambda function's input. In most cases this should be `Codable`. If your event originates from an
     /// AWS service, have a look at [AWSLambdaEvents](https://github.com/swift-server/swift-aws-lambda-events),
     /// which provides a number of commonly used AWS Event implementations.
-    associatedtype Event
+    associatedtype Event: Sendable
     /// The lambda function's output. Can be `Void`.
-    associatedtype Output
+    associatedtype Output: Sendable
 
     /// The Lambda initialization method.
     /// Use this method to initialize resources that will be used in every request.
@@ -287,9 +287,9 @@ public protocol EventLoopLambdaHandler {
     /// The lambda functions input. In most cases this should be `Codable`. If your event originates from an
     /// AWS service, have a look at [AWSLambdaEvents](https://github.com/swift-server/swift-aws-lambda-events),
     /// which provides a number of commonly used AWS Event implementations.
-    associatedtype Event
+    associatedtype Event: Sendable
     /// The lambda functions output. Can be `Void`.
-    associatedtype Output
+    associatedtype Output: Sendable
 
     /// Create a Lambda handler for the runtime.
     ///


tomerd:
this seems reasonable to me. @fabianfett thoughts?

ktoso (approved):

ktoso:
looks sensible to me 👍

sebsto:
It looks like there are some errors I did not capture locally.

First API change. Is there something I can do to tell the CI system this is indeed an intended API change ?

```
09:25:45 3 breaking changes detected in AWSLambdaRuntime:
09:25:45   💔 API breakage: extension SimpleLambdaHandler has generic signature change from  to <Self.Event : Swift.Sendable, Self.Output : Swift.Sendable>
09:25:45   💔 API breakage: extension LambdaHandler has generic signature change from  to <Self.Event : Swift.Sendable, Self.Output : Swift.Sendable>
09:25:45   💔 API breakage: extension EventLoopLambdaHandler has generic signature change from  to <Self.Event : Swift.Sendable, Self.Output : Swift.Sendable>
09:25:45 
09:25:45 3 breaking changes detected in AWSLambdaRuntimeCore:
09:25:45   💔 API breakage: protocol SimpleLambdaHandler has generic signature change from  to <Self.Event : Swift.Sendable, Self.Output : Swift.Sendable>
09:25:45   💔 API breakage: protocol LambdaHandler has generic signature change from  to <Self.Event : Swift.Sendable, Self.Output : Swift.Sendable>
09:25:45   💔 API breakage: protocol EventLoopLambdaHandler has generic signature change from  to <Self.Event : Swift.Sendable, Self.Output : Swift.Sendable>
```

The second set of error seems to deal with strict concurrency checks

```
11:30:08 /code/Sources/AWSLambdaRuntimeCore/LambdaRuntime.swift:180:29: error: reference to captured var 'logger' in concurrently-executing code; this is an error in the Swift 6 language mode
11:30:08 178 |                             // invocation request might have been cancelled. For this reason we
11:30:08 179 |                             // succeed the promise here.
11:30:08 180 |                             logger.log(level: .info, "lambda invocation sequence has been cancelled for shutdown")
11:30:08     |                             `- error: reference to captured var 'logger' in concurrently-executing code; this is an error in the Swift 6 language mode
11:30:08 181 |                             return promise.succeed(count)
11:30:08 182 |                         }
```

Any idea how to address these ?

fabianfett (changes requested):
I don't think we should land this, as this is fundamentally a Swift issue. The problem here is that Sendable checking is too aggressive in 5.10, as the compiler does not understand transfer semantics around actors. This will be fixed in Swift 6, with [region based isolation](https://github.com/apple/swift-evolution/blob/main/proposals/0414-region-based-isolation.md).

As this introduces a requirement on lambda handlers that is not needed, I'm not in support of this patch.

fabianfett:
> First API change. Is there something I can do to tell the CI system this is indeed an intended API change ?
This CI step isn't required. It is fine to be red. We use it as a warning signal. The only way to make it not red, is to not make breaking changes. But even then it sometimes has false positives.

sebsto:
as per discussion with @fabianfett, let's not merge this change. It should be fixed in Swift 6.0

PR #333, opened 2024-06-21
Add support for resources when packaging using the SwiftPM plugin

camdenfullmer:
This change updates the SwiftPM plugin to include the executable's resources in the archive.

### Motivation:

When deploying a Lambda function that generates HTML content I needed a way to package template HTML files along with the executable. Defining the resources in the `Package.swift` and then packaging them in the archive in a way so that Swift can access them using `Bundle.main.resourceURL` felt like the best way to accomplish this.

Example:

```
.executableTarget(
            name: "Website",
            dependencies: [
                .product(name: "AWSLambdaRuntime", package: "swift-aws-lambda-runtime"),
            ],
            resources: [
                .process("template.stencil"),
            ]),
```

### Modifications:

Plugin code has been modified to generate a `Contents/Resources` directory inside the final archive which includes the resources defined in the executable's target from `Package.swift`.

### Result:

The zip archive includes the defined resources in `Package.swift`.

### DIFF
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -67,6 +67,7 @@ struct AWSLambdaPackager: CommandPlugin {
 
         // create the archive
         let archives = try self.package(
+            packageName: context.package.displayName,
             products: builtProducts,
             toolsProvider: { name in try context.tool(named: name).path },
             outputDirectory: configuration.outputDirectory,
@@ -183,6 +184,7 @@ struct AWSLambdaPackager: CommandPlugin {
 
     // TODO: explore using ziplib or similar instead of shelling out
     private func package(
+        packageName: String,
         products: [LambdaProduct: Path],
         toolsProvider: (String) throws -> Path,
         outputDirectory: Path,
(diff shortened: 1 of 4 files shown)


sebsto:
Looks like this address issue #312

sebsto:
@camdenfullmer  Thank you for this PR. It looks like a legitimate idea and simple implementation

I would propose to add a minimal example Lambda function to show how to use this capability and a section in the README file to document this.

Question: why `Contents/Resources` and not just `Resources` like the mac bundles ? (I know, naming things is hard) 

I left comments on the code

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -202,8 +202,27 @@ struct AWSLambdaPackager: CommandPlugin {
             try FileManager.default.copyItem(atPath: artifactPath.string, toPath: relocatedArtifactPath.string)
             try FileManager.default.createSymbolicLink(atPath: symbolicLinkPath.string, withDestinationPath: relocatedArtifactPath.lastComponent)
 
+            // add resources
+            let contentsDirectory = workingDirectory.appending("Contents")
```
sebsto, line 206: do we really need a `Contents` directory ? Maybe `Resources` is enough

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -202,8 +202,27 @@ struct AWSLambdaPackager: CommandPlugin {
...
+            try FileManager.default.createDirectory(atPath: contentsDirectory.string, withIntermediateDirectories: true)
+            let relocatedResourcesDirectory = contentsDirectory.appending("Resources")
+            let artifactDirectory = artifactPath.removingLastComponent()
+            let resourcesDirectoryName = try FileManager.default.contentsOfDirectory(atPath: artifactDirectory.string)
+                .first(where: { $0.hasSuffix(".resources") && $0.contains(product.name) })
+            if let resourcesDirectoryName {
```
sebsto, line 212: I would use `guard` or `if` on line 210. This line is not necessary

sebsto (commented):
A few questions and suggestions. Thank you for having submitted this

sebsto:

sebsto:

sebsto:
@camdenfullmer Thank you for having submitted this PR. We're considering merging it in the main branch before we start work on v2.
Can you address the two suggestions left above ?

camdenfullmer:
> @camdenfullmer Thank you for having submitted this PR. We're considering merging it in the main branch before we start work on v2. Can you address the two suggestions left above ?

Yep will get right on it. Will try to have changes and feedback by the end of the week!

Plugins/AWSLambdaPackager/Plugin.swift
camdenfullmer: After some more testing the correct bundle to use is `Bundle.module`. So instead of recreating the `Contents` directory I am just copying the `.resources` directory that is in the artifacts directory to the working directory to be zipped. This has the benefit of working on Lambda as well as when running locally on macOS.

camdenfullmer: https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package#Access-a-resource-in-code

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -67,6 +67,7 @@ struct AWSLambdaPackager: CommandPlugin {
 
         // create the archive
         let archives = try self.package(
+            packageName: context.package.displayName,
```
camdenfullmer, line 70: Wasn't sure if there was a better way to get the package name when creating the resources directory name.

sebsto: I think this is the only way to retrieve a package name. 
See https://github.com/swiftlang/swift-package-manager/blob/b24cc8b547f952ee4a954db939f115a26946fe2b/Sources/PackagePlugin/PackageModel.swift#L16

The other option is to use the directory name but it's not necessarily the same.

sebsto:

sebsto:
@camdenfullmer Thank you for the update. I tested the code on different variations of your sample project.  It works like a charm,
Can you make sure the `scripts/soundness.sh` script runs without error ? It's just the license headers missing from the example project.

```
19:06:17    * swift-or-c... missing headers in file './Examples/ResourcePackaging/Lambda.swift'!
```

camdenfullmer:
> @camdenfullmer Thank you for the update. I tested the code on different variations of your sample project. It works like a charm, Can you make sure the `scripts/soundness.sh` script runs without error ? It's just the license headers missing from the example project.
> 
> ```
> 19:06:17    * swift-or-c... missing headers in file './Examples/ResourcePackaging/Lambda.swift'!
> ```

I changed the year to something that will pass, but maybe the script should be updated to allow for `2024`?

sebsto:

sebsto (commented):
LGTM

sebsto (approved):

PR #334, opened 2024-06-24
[Draft] Detached tasks

Buratti:
Introduces a system that allows to keep the execution environment running after submitting the response for a synchronous invocation.

### Note:

This is a draft PR, only meant to gather feedbacks. This code requires some polishing and testing.
The lines commented with `// To discuss:` require careful consideration.

### Motivation:

It is common for a Lambda function to depend on several other serverless services (SQS, APIGW WebSockets, SNS, EventBridge, X-Ray just to name a few). In many occasions, such services might perform non-critical work like sending a push notification to the user, store metrics or flush a set of spans. In these scenarios, the occurrence of a non-recoverable error usually doesn't mean that the whole invocation should fail and result in a sad 500. At the same time, awaiting for such tasks to finish gives no benefits and drastically increases the latency for the API consumers.

This PR implements a hook in the runtime which allows the developer to dispatch code that can continue its execution **after** the invocation of `/response` and before having the environment being frozen by `/next`. This is a common practice described [here](https://aws.amazon.com/blogs/compute/running-code-after-returning-a-response-from-an-aws-lambda-function/).

### Modifications:

A new class called `DetachedTaskContainer` has been added, which follows the design of `LambdaTerminator` as close as possible. A `LambdaContext` now owns an instance of `DetachedTaskContainer`, which can be used by the handler to dispatch asynchrouns non-critical work that can finish **after** the invocation of `/response` and before `/next`.

### Result:

It is now possible to return the result of a synchronous invocation (like API Gateway integration, CloudFront origin, Lambda function URL, Cognito Custom Auth, etc), as soon as it is ready, reducing the overall API latency.

```swift
func handle(_ buffer: ByteBuffer, context: LambdaContext) -> EventLoopFuture<ByteBuffer?> {
    // EventLoop API
    context.tasks.detached(name: "Example 1") { eventLoop in
        return performSomeWork(on: eventLoop).map { result in
            context.logger.info("\(result) has been provided at \(Date())")
        }
    }
    // Async API
    context.tasks.detached(name: "Example 2") {
        try await Task.sleep(for: .seconds(3))
        context.logger.info("Hello World!")
    }
    let response = encodedAPIGatewayResponse()
    return context.eventLoop.makeSucceededFuture(response)
}
```

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
+++ b/Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
@@ -1,0 +1,169 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import Foundation
+import NIOConcurrencyHelpers
+import NIOCore
+import Logging
+
+/// A container that allows tasks to finish after a synchronous invocation
+/// has produced its response.
+public final class DetachedTasksContainer {
+
+    struct Context {
+        let eventLoop: EventLoop
+        let logger: Logger
+    }
+    
+    private var context: Context
+    private var storage: Storage
+
+    init(context: Context) {
+        self.storage = Storage()
+        self.context = context
+    }
+
+    /// Adds a detached task that runs on the given event loop.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the task.
+    ///   - task: The task to execute. It receives an `EventLoop` and returns an `EventLoopFuture<Void>`.
+    /// - Returns: A `RegistrationKey` for the registered task.
+    @discardableResult
+    public func detached(name: String, task: @escaping (EventLoop) -> EventLoopFuture<Void>) -> RegistrationKey {
+        let key = RegistrationKey()
+        let task = task(context.eventLoop).always { _ in
+            self.storage.remove(key)
+        }
+        self.storage.add(key: key, name: name, task: task)
+        return key
+    }
+    
+    /// Adds a detached async task.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the task.
+    ///   - task: The async task to execute.
+    /// - Returns: A `RegistrationKey` for the registered task.
+    @discardableResult
+    public func detached(name: String, task: @Sendable @escaping () async throws -> Void) -> RegistrationKey {
+        let key = RegistrationKey()
+        let promise = context.eventLoop.makePromise(of: Void.self)
+        promise.completeWithTask(task)
+        let task = promise.futureResult.always { result in
+            switch result {
+            case .success:
+                break
+            case .failure(let failure):
+                self.context.logger.warning(
+                    "Execution of detached task failed with error.",
+                    metadata: [
+                        "taskName": "\(name)",
+                        "error": "\(failure)"
+                    ]
+                )
+            }
+            self.storage.remove(key)
+        }
+        self.storage.add(key: key, name: name, task: task)
+        return key
+    }
+
+    /// Informs the runtime that the specified task should not be awaited anymore.
+    ///
+    /// - Warning: This method does not actually stop the execution of the
+    ///   detached task, it only prevents the runtime from waiting for it before
+    ///   `/next` is invoked.
+    ///
+    /// - Parameter key: The `RegistrationKey` of the task to cancel.
+    public func unsafeCancel(_ key: RegistrationKey) {
+        // To discuss:
+        // Canceling the execution doesn't seem to be an easy
+        // task https://github.com/apple/swift-nio/issues/2087
+        //
+        // While removing the handler will allow the runtime
+        // to invoke `/next` without actually awaiting for the
+        // task to complete, it does not actually cancel
+        // the execution of the dispatched task.
+        // Since this is a bit counter-intuitive, we might not
+        // want this method to exist at all.
+        self.storage.remove(key)
+    }
+
+    /// Awaits all registered tasks to complete.
+    ///
+    /// - Returns: An `EventLoopFuture<Void>` that completes when all tasks have finished.
+    internal func awaitAll() -> EventLoopFuture<Void> {
+        let tasks = storage.tasks
+        if tasks.isEmpty {
+            return context.eventLoop.makeSucceededVoidFuture()
+        } else {
+            return EventLoopFuture.andAllComplete(tasks.map(\.value.task), on: context.eventLoop).flatMap {
+                self.awaitAll()
+            }
+        }
+    }
+}
+
+extension DetachedTasksContainer {
+    /// Lambda detached task registration key.
+    public struct RegistrationKey: Hashable, CustomStringConvertible {
+        var value: String
+
+        init() {
+            // UUID basically
+            self.value = UUID().uuidString
+        }
+
+        public var description: String {
+            self.value
+        }
+    }
+}
+
+extension DetachedTasksContainer {
+    fileprivate final class Storage {
+        private let lock: NIOLock
+        
+        private var map: [RegistrationKey: (name: String, task: EventLoopFuture<Void>)]
+
+        init() {
+            self.lock = .init()
+            self.map = [:]
+        }
+
+        func add(key: RegistrationKey, name: String, task: EventLoopFuture<Void>) {
+            self.lock.withLock {
+                self.map[key] = (name: name, task: task)
+            }
+        }
+
+        func remove(_ key: RegistrationKey) {
+            self.lock.withLock {
+                self.map[key] = nil
+            }
+        }
+
+        var tasks: [RegistrationKey: (name: String, task: EventLoopFuture<Void>)] {
+            self.lock.withLock {
+                self.map
+            }
+        }
+    }
+}
+
+// Ideally this would not be @unchecked Sendable, but Sendable checks do not understand locks
+// We can transition this to an actor once we drop support for older Swift versions
+extension DetachedTasksContainer: @unchecked Sendable {}
+extension DetachedTasksContainer.Storage: @unchecked Sendable {}
+extension DetachedTasksContainer.RegistrationKey: Sendable {}
--- a/Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
+++ b/Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
@@ -79,27 +79,6 @@ public final class DetachedTasksContainer {
         return key
     }
 
-    /// Informs the runtime that the specified task should not be awaited anymore.
-    ///
-    /// - Warning: This method does not actually stop the execution of the
-    ///   detached task, it only prevents the runtime from waiting for it before
-    ///   `/next` is invoked.
-    ///
-    /// - Parameter key: The `RegistrationKey` of the task to cancel.
-    public func unsafeCancel(_ key: RegistrationKey) {
-        // To discuss:
-        // Canceling the execution doesn't seem to be an easy
-        // task https://github.com/apple/swift-nio/issues/2087
-        //
-        // While removing the handler will allow the runtime
-        // to invoke `/next` without actually awaiting for the
-        // task to complete, it does not actually cancel
-        // the execution of the dispatched task.
-        // Since this is a bit counter-intuitive, we might not
-        // want this method to exist at all.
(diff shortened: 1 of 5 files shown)


Buratti:
@sebsto @adam-fowler any feedback is welcome

sebsto:
I love the idea. I need more time to look at the implementation
FYI, we're planning to rewrite the runtime for a v2. a rewrite to simplify and better embrace Swift  concurrency and service lifecycle. I wonder at which point it is pertinent to include this type of functionality now or wait for the rewrite.
@fabianfett wdyt ?

Buratti:
That's great! I would love to contribute to a more Swift Concurrency friendly runtime.
About this specific feature, I wouldn't mind seeing in the v1 as well. I have been testing it in our staging environment and the difference in latency is indeed noticeable. Also, there are no breaking changes in this PR.

sebsto:

fabianfett (changes requested):
I'm not in favor of this pr for two reasons:

1. AWS Lambda does not support background processing in the way this pr tries to add this functionality. AWS Lambda freezes the users execution environment as soon as a request has finished. This means we can not rely on any background processing in Lambda. The correct way to implement this is to wrap the [Extensions API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html), that allows background processing for which users are then charged correctly.

2. The proposed approach does not integrate nicely with Swift structured concurrency. The correct Swift structured concurrency approach should work as follows:

```swift
enum MyBackgroundEvents {
  case myEvent(String)
}

let (stream, continuation) = AsyncStream.makeStream(of: MyBackgroundEvents.self)
let runtime = LambdaRuntime { (request, context) -> String in
  continuation.yield(.myEvent("whatever"))  
  return "Woohoo"
}

await withTaskGroup {
  taskGroup.addTask {
    try await runtime.run()
  }

  taskGroup.addTask {
    for await event in stream {
      // do the actual background processing
    }
  }
}
```

We intend to build out an API like the one above. However currently we are blocked on this [ServiceLifecycle issue](https://github.com/swift-server/swift-service-lifecycle/issues/185).

**However** even though this is the correct Swift approach, it is very wrong in the Lambda usecase (see 1.). The real thing is to write a wrapper based on the same primitives that correctly implements the [Extensions API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html).

Buratti:
1. Actually that's not correct. AWS Lambda freezes the execution environment when `/next` is invoked. The time elapsed between `/response|error` and `/next` is billed normally. The article from AWS Blogs I linked in the first comment explains in details both approaches (this one and the one leveraging the Internal Extensions API).

2. The proposed design tries to be as much as possible aligned to the way Swift concurrency works.
```swift
// Pure Swift
Task.detached {
  try await fireAndForget()
}

// Lambda handler
context.task.detached {
  try await fireAndForget()
}
```

fabianfett:
Having thought about this a bit more, I think I have to conclusion that I'm open to this patch in v1. For v2, we should 100% make the concurrency approach structured. A new API proposal will be made shortly.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
```diff
@@ -0,0 +1,169 @@
...
+//
+//===----------------------------------------------------------------------===//
+import Foundation
+import NIOConcurrencyHelpers
+import NIOCore
+import Logging
+
+/// A container that allows tasks to finish after a synchronous invocation
+/// has produced its response.
+public final class DetachedTasksContainer {
```
fabianfett, line 21: why does this need to be public?

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+    }
+}
+
+// Ideally this would not be @unchecked Sendable, but Sendable checks do not understand locks
+// We can transition this to an actor once we drop support for older Swift versions
+extension DetachedTasksContainer: @unchecked Sendable {}
+extension DetachedTasksContainer.Storage: @unchecked Sendable {}
+extension DetachedTasksContainer.RegistrationKey: Sendable {}
```
fabianfett, line 169: We must make sure that all those types are `Sendable` without using `@unchecked`. I'm sure we can achieve this.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
```diff
@@ -0,0 +1,169 @@
...
+
+        public var description: String {
+            self.value
+        }
+    }
+}
+
+extension DetachedTasksContainer {
+    fileprivate final class Storage {
+        private let lock: NIOLock
```
fabianfett, line 136: Use a [`NIOLockedValueBox`](https://github.com/apple/swift-nio/blob/96149964cba9982b40729589376050307972f1c8/Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift#L24) here instead.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
```diff
@@ -0,0 +1,169 @@
...
+            return EventLoopFuture.andAllComplete(tasks.map(\.value.task), on: context.eventLoop).flatMap {
+                self.awaitAll()
+            }
+        }
+    }
+}
+
+extension DetachedTasksContainer {
+    /// Lambda detached task registration key.
+    public struct RegistrationKey: Hashable, CustomStringConvertible {
```
fabianfett, line 120: I think we don't need this. We should not support removing background tasks.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -158,6 +161,10 @@ public struct LambdaContext: CustomDebugStringConvertible, Sendable {
     public var allocator: ByteBufferAllocator {
         self.storage.allocator
     }
+    
+    public var tasks: DetachedTasksContainer {
```
fabianfett, line 165: I don't think we need to make the task container a property here. Instead we should add a `func detachedBackgroundTask(_ body: @escaping @Sendable () async -> ())` API here. Which then uses the task container internally.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// Adds a detached task that runs on the given event loop.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the task.
+    ///   - task: The task to execute. It receives an `EventLoop` and returns an `EventLoopFuture<Void>`.
+    /// - Returns: A `RegistrationKey` for the registered task.
+    @discardableResult
+    public func detached(name: String, task: @escaping (EventLoop) -> EventLoopFuture<Void>) -> RegistrationKey {
```
fabianfett, line 43: We should not add a new `EventLoop` API.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+    
+    /// Adds a detached async task.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the task.
+    ///   - task: The async task to execute.
+    /// - Returns: A `RegistrationKey` for the registered task.
+    @discardableResult
+    public func detached(name: String, task: @Sendable @escaping () async throws -> Void) -> RegistrationKey {
```
fabianfett, line 59: We should enforce that users handle errors! For this reason the user's closure should just be `async` but not `throws`

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// Informs the runtime that the specified task should not be awaited anymore.
+    ///
+    /// - Warning: This method does not actually stop the execution of the
+    ///   detached task, it only prevents the runtime from waiting for it before
+    ///   `/next` is invoked.
+    ///
+    /// - Parameter key: The `RegistrationKey` of the task to cancel.
+    public func unsafeCancel(_ key: RegistrationKey) {
```
fabianfett, line 89: We are not using structured concurrency, for this reason we should not try to add cancel support. Cancel support will be added when we move to a fully async/structured concurrency api.

fabianfett:
@Buratti Sorry for not making this explicit. Thanks for pushing on this! Really appreciated. Gave me quite a bit food for thought! Let's try to land this!

Buratti:
Hi @fabianfett! Thank you for giving this a deeper thought and for the feedbacks, I'll make sure to address them in the next couple of days.
I'm very glad to hear you have reconsidered this feature, and yes, I definitely agree on having a fully structured concurrency version of it for the v2!

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
Buratti: It does not. I like the idea you mentioned [here](https://github.com/swift-server/swift-aws-lambda-runtime/pull/334#discussion_r1666710596).

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
Buratti: I believe we can, I took inspiration from [Terminator.swift](https://github.com/swift-server/swift-aws-lambda-runtime/blob/872183bbe99aaa26e52f9d4baef6adaf815f45df/Sources/AWSLambdaRuntimeCore/Terminator.swift#L143) for that part, but I'll change it.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
Buratti: This is one of the point I was most worried about, I like the idea of not supporting cancelation at all, it makes things much easier to handle and it is more aligned with how NIO works.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
Buratti: I did not like this API either, I added it just for consistency with [this](https://github.com/swift-server/swift-aws-lambda-runtime/blob/872183bbe99aaa26e52f9d4baef6adaf815f45df/Sources/AWSLambdaRuntimeCore/Terminator.swift#L38) but I am perfectly fine with removing it, let me know please.

Sources/AWSLambdaRuntimeCore/DetachedTasks.swift
Buratti: Definitely the best approach 👍🏻

Sources/AWSLambdaRuntimeCore/LambdaRunner.swift
```diff
@@ -95,13 +95,19 @@ internal final class LambdaRunner {
...
-                    return (invocation, result)
+                    return (invocation, result, context)
                 }
-        }.flatMap { invocation, result in
+        }.flatMap { invocation, result, context in
             // 3. report results to runtime engine
             self.runtimeClient.reportResults(logger: logger, invocation: invocation, result: result).peekError { error in
                 logger.error("could not report results to lambda runtime engine: \(error)")
-            }
+                // To discuss:
```
Buratti, line 104: This is the only point that remains open @fabianfett. When the runtime fails to report a result to AWS Lambda, do we want to wait for the background tasks to complete before stopping the execution of the whole process?

fabianfett: In the error case we should await all subtasks.

sebsto:
Thank you @Buratti for having submitted this change.
As we discussed, we're going to merge it to the main branch while, at the same time, adopting a different and not compatible approach for v2 
Can you address the last remaining point ?

Buratti:
Hi @sebsto! Please forgive me, last few weeks have been very full! I'll try my best to have them ready before tomorrow EOD!

sebsto:

sebsto:
Thank you for the changes ! 

Can you add a example code + a section in the README doc - that will allows developers to learn about that feature without having to look at the internals of the code :-) 

It looks like the code does not compile on 5.7 - Is this something easy to fix ?
```
17:37:09 /code/Sources/AWSLambdaRuntimeCore/LambdaRunner.swift:113:10: error: type of expression is ambiguous without more context
17:37:09         .flatMap { context in
17:37:09 ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
```

fabianfett:
@Buratti Please have a look at #339.

Buratti:
Hi @sebsto @fabianfett.
I should be able to fix the PR this evening (CEST).

Thanks for sharing the V2 proposal with me, I will read it thoroughly as soon as I finished the changes to this PR.

fabianfett:

Buratti:
Should I run the soundness script?

sebsto:
I’m off this week with just my phone and crappy network. I’ll check the submission next week. 

But to answer your question, yes the soundness script must run without reporting errors. 
the nightly build fail can be ignored.

PR #339, opened 2024-08-08
v2 API Proposal Document

aryan-25:
Hello 👋

I am Aryan, a Computer Science student and currently an intern at Apple on the Swift on Server team. As part of my internship, I have been working with @fabianfett to propose a new v2 API for the `swift-aws-lambda-runtime` library.

`swift-aws-lambda-runtime` is an important library for the Swift on Server ecosystem. The initial API was written before async/await was introduced to Swift. When async/await was introduced, shims were added to bridge between the underlying SwiftNIO `EventLoop` interfaces and async/await.

However, just like `gRPC-swift` and `postgres-nio`, we now want to shift to solely using async/await constructs instead of `EventLoop` interfaces. For this, large parts of the current API have to be reconsidered. This also provides a good opportunity to add support for new AWS Lambda features such as [response streaming](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/). 

We have written a document that explains the current limitations of the library and proposes a detailed design for the v2 API.

Please read the proposal and voice your opinions either here or on [the forum post](https://forums.swift.org/t/aws-lambda-v2-api-proposal/73819). We are looking forward to your feedback!

cc: @sebsto @tachyonics @FranzBusch @Lukasa @tomerd @weissi @adam-fowler @ktoso @0xTim


sebsto:
Thank you @aryan-25 ! I'll have a look this morning

sebsto:
This is not related to this doc, but should we move the `Documentation.docc` directory just underneath `Sources` ?
@fabianfett wdyt ?

Current struct :

```
Sources/AWSLambdaRuntimeCore
└── Documentation.docc
    ├── Resources
    │   ├── code
    │   └── tutorials
    └── tutorials
```

Proposed struct 

```
Sources/Documentation.docc
├── Resources
│   ├── code
│   └── tutorials
└── tutorials
```

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+        context: LambdaContext
+    ) async throws -> LambdaResponse
+}
+```
+
+### Handler as a Closure
+
+To create a Lambda function using the current API, a user first has to create an object and conform it to one of the handler protocols by implementing the initializer and the `handle(...)` function. Now that `LambdaRuntime` is public, this verbosity can very easily be simplified.
+
+**ClosureHandler**
```
0xTim, line 404: Should this be a heading?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+
+However, if it comes to consistency with the larger Swift on server ecosystem, the API that returns a `LambdaResponse` is likely the better choice. Hummingbird v2, OpenAPI and the new Swift gRPC v2 implementation all use this approach. This might be due to the fact that writing middleware becomes easier, if a Response is explicitly returned.
+
+We decided to implement the approach in which a `LambdaResponseWriter` is passed to the function, since the approach in which a `LambdaResponse` is returned can trivially be built on top of it. This is not true vice versa.
+
+We welcome the discussion on this topic and are open to change our minds and API here.
```
0xTim, line 552: It would be interesting to see if anyone has any use cases that integrate something like Swift Middleware

0xTim (commented):
Great steps forward!

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+    /// Report an error in the response stream
+    consuming func reportError(_ error: any Error) async throws
+}
+```
+
+### LambdaContext
+
+`LambdaContext` will be largely unchanged, but the `eventLoop` property as well as the `allocator` property (of type `ByteBuffer`) will be removed.
+
+A new function `backgroundTask()` will also be added. This will allow tasks to be run in the background while and after the response is/has been sent. Please note that `LambdaContext` will not be Sendable anymore.
```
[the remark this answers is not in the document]
fabianfett: In order to allow adding background tasks in a structured way, we will need to back the `LambdaContext` with a `TaskGroup`. Since `TaskGroup` is not `Sendable` we can't mark `LambdaContext` as `Sendable`. However if users need any property (all of them are `Sendable`) out of the `LambdaContext` they can just get that property out of it and then pass it around or close over it.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+        responseWriter.finish()
+    }
+}
+```
+
+The method is not marked as `throws` as we want the users of this API to handle errors themselves and call `LambdaResponseWriter.reportError(_:)` if needed.
+
+### LambdaRuntime
+
+`LambdaRuntime` is the class that communicates with the Lambda control plane as defined in [Building a custom runtime for AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html) and forward the invocations to the provided `StreamingLambdaHandler`. It will conform to `ServiceLifecycle.Service` to provide support for `swift-service-lifecycle`.
```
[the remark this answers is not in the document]
fabianfett: In order to use a `LambdaHandler` users will wrap it in a `LambdaCodableAdapter`. `LambdaCodableAdapter` conforms to `StreamingLambdaHandler`. So the LambdaRuntime type that you will use when using LambdaHandler will be:

```
let runtime: LambdaRuntime<LambdaCodableAdapter<MyLambdaHandler, Event, Output, JSONDecoder, JSONEncoder>>
```

but users basically never have to use that type explicitly. So to answer your question: Yes it works with `LambdaHandler` as well, but via the `LambdaCodableAdapter`.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+    services: [postgresClient, runtime],
+    configuration: .init(gracefulShutdownSignals: [.sigterm]),
+    logger: logger
+)
+try await serviceGroup.run()
+```
+
+#### Simplifying Codable support
+
+A detailed explanation is provided in the Codable Support section. In short, much of the boilerplate code defined for each handler protocol in `Lambda+Codable` and `Lambda+String` will be replaced with a single `LambdaCodableAdapter` class.
```
FranzBusch, line 117: Why is the adapter a `class`?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+/// A writer object to write the Lambda response stream into
+public protocol LambdaResponseWriter: ~Copyable {
+    /// Write a response part into the stream
+    func write(_ buffer: ByteBuffer) async throws
+    /// End the response stream
+    consuming func finish()
+    /// Write a response part into the stream and end the response stream
+    consuming func writeAndFinish(_ buffer: ByteBuffer) async throws
+    /// Report an error in the response stream
+    consuming func reportError(_ error: any Error) async throws
```
FranzBusch, line 143: This seems like it is finishing as well? Do just want to spell this `finish(error: (any Error)?)` instead?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+    /// `Logger` to log with.
+    ///
+    /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
+    public var logger: Logger { get }
+
+    /// Schedule background work in lambda. This work must be completed before
+    /// the runtime can ask for further events. Note that AWS will continue to
+    /// charge you after the response has been returned, but work is still
+    /// processing in the background.
+    public func addBackgroundTask(_ body: sending @escaping () async -> ())
```
FranzBusch, line 183: This doesn't feel necessary to me with the current API proposal. Since we pass in an independent writer. A user can just write some bytes, the call writer then finish and then do any background work before returning from the `handle` method. They can even set that up in child tasks in the `handle` method to start the background work while handling the actual request.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+    /// For invocations from the AWS Mobile SDK, data about the Amazon Cognito identity provider.
+    public var cognitoIdentity: String? { get }
+
+    /// For invocations from the AWS Mobile SDK, data about the client application and device.
+    public var clientContext: String? { get }
+
+    /// `Logger` to log with.
+    ///
+    /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
+    public var logger: Logger { get }
```
FranzBusch, line 177: Wondering how we feel about this in relation to https://github.com/apple/swift-log/pull/315. Also why do we call out the env variable here? Shouldn't we just pass in the logger to the runtime and that logger determines the log level?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+    public var requestID: String { get }
+
+    /// The AWS X-Ray tracing header.
+    public var traceID: String { get }
+
+    /// The ARN of the Lambda function, version, or alias that's specified in the invocation.
+    public var invokedFunctionARN: String { get }
+
+    /// The timestamp that the function times out.
+    public var deadline: DispatchWallTime { get }
```
FranzBusch, line 166: Can we use `ContinuousClock.Instant` here?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+/// The base LambdaHandler protocol
+public protocol StreamingLambdaHandler: ~Copyable {
+    /// The business logic of the Lambda function
+    /// - Parameters:
+    ///   - event: The invocation's input data
+    ///   - responseWriter: A ``LambdaResponseWriter`` to write the invocations response to.
+    ///                     If no response or error is written to the `responseWriter` it will
+    ///                     report an error to the invoker.
+    ///   - context: The LambdaContext containing the invocation's metadata
+    mutating func handle(_ event: ByteBuffer, responseWriter: consuming some LambdaResponseWriter, context: LambdaContext) async
```
FranzBusch, line 201: Shouldn't we make it possible that this method throws? Otherwise it forces every user to wrap any throwing calls into a do/catch. FWIW in [grpc](https://github.com/grpc/grpc-swift/blob/main/Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md) we did exactly the opposite and handle that for the user since it makes it way more ergonomic to write those handlers

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+
+    /// Create a LambdaRuntime by passing a handler, an eventLoop and a logger.
+    /// - Parameter handler: A ``StreamingLambdaHandler`` that will be invoked
+    /// - Parameter eventLoop: An ``EventLoop`` on which the LambdaRuntime will be
+    ///                        executed. Defaults to an EventLoop from
+    ///                        ``NIOSingletons.posixEventLoopGroup``.
+    /// - Parameter logger: A logger
+    public init(
+        handler: consuming sending Handler,
+        eventLoop: EventLoop = Lambda.defaultEventLoop,
```
FranzBusch, line 246: Why are we using our own default EL here instead of the singleton EL?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+        logger: Logger = Logger(label: "Lambda")
```
FranzBusch, line 247: IMO we shouldn't default the logger. Let users pass this in and if they really want to create a fresh logger they can decide to do that

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+
+### Lambda
+
+We also add an enum to store a static function and a property on. We put this on the static `Lambda` because `LambdaRuntime` is generic and thus has bad ergonomics for static properties and functions.
+
+```swift
+enum Lambda {
+    /// This returns the default EventLoop that a LambdaRuntime is scheduled on.
+    /// It uses `NIOSingletons.posixEventLoopGroup.next()` under the hood.
+    public static var defaultEventLoop: any EventLoop { get }
```
FranzBusch, line 275: Like noted I don't see why we would need to create a separate static var for this?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+```
+
+The current API allows for a Lambda function to be tested locally through a mock server by requiring an environment variable named  `LOCAL_LAMBDA_SERVER_ENABLED` to be set to `true`. If this environment variable is not set, the program immediately crashes as the user will not have the `AWS_LAMBDA_RUNTIME_API` environment variable on their local machine (set automatically when deployed to AWS Lambda). However, making the user set the `LOCAL_LAMBDA_SERVER_ENABLED` environment variable is an unnecessary step that can be avoided. In the v2 API, the `run()` function will automatically start the mock server when the `AWS_LAMBDA_RUNTIME_API` environment variable cannot be found.
+
+### Lambda
+
+We also add an enum to store a static function and a property on. We put this on the static `Lambda` because `LambdaRuntime` is generic and thus has bad ergonomics for static properties and functions.
+
+```swift
+enum Lambda {
```
FranzBusch, line 272: Is this namespace really worth it if we remove the EL default and can we then just make the `reportStartupError` global method?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+    /// This returns the default EventLoop that a LambdaRuntime is scheduled on.
+    /// It uses `NIOSingletons.posixEventLoopGroup.next()` under the hood.
+    public static var defaultEventLoop: any EventLoop { get }
+
+    /// Report a startup error to the Lambda Control Plane API
+    public static func reportStartupError(any Error) async
```
FranzBusch, line 278: What if we make this a `with` method so that users can do any setup and we report any error that happens there. Instead of them having to do a `do/catch`?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,556 @@
...
+### Codable support
+
+For v2 we want to introduce a `LambdaCodableAdapter` that:
+
+1. Accepts any generic underlying handler conforming to a new protocol `LambdaHandler` specifying `associatedtype`s `Event` and `Output`, and a `handle(_ event: Event, context: LambdaContext) async throws -> Output` function.
+
+#### LambdaHandler
+
+```swift
+public protocol LambdaHandler {
```
FranzBusch, line 323: Just out of curiosity we can't solve this with protocol inheritance right? We do need the various different protocols and adapters?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
FranzBusch: I think this is a fair point and I would love to see how middleware would work with this writer approach. IMO middleware do make some sense for Lambda

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: Hahaha. Leftover has become a struct in our prototype already.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: Not sure I agree here. I think the `reportError` is more explicit in what it does, than a finish with an error.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: Nope this is epoch time.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: The default EL uses the singleton ELG under the hood. The default EL picks an EL from the singleton ELG. Users can use this to get their invocations onto the same EL as the runtime.

```swift 
try await withTaskExecutor(Lambda.defaultEventLoop) {
    try await runtime.run()
}
```

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: See above.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: We do have protocol inheritance today and I don't like it. The problem here is that there is never a case where the protocol inheritance here makes sense from a user point of view. You either want to use pure streaming or you want to use Codable. But there is no point in which you opt to override the streaming API when you implement `LambdaHandler`. Worst of all injecting custom Encoders and Decoders only works through globals.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: I think that makes sense. Would you default the Logger to nil then? The idea of using a default logger here is that new adopters don't need to learn about Logging first.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: Yeah I think I can get behind that.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: I think your benchmark shows clearly that using TaskLocals is more expensive than passing around a logger time wise. 
Given that Lambdas are payed by the milliseconds I certainly don't want to make using TaskLocals the only approach.
Also I want this API to land sooner rather than later and not block on the above discussion that can be seen as an addition.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
sebsto: As per [Lambda runtime API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next)

> Lambda-Runtime-Deadline-Ms – The date that the function times out in Unix time milliseconds. For example, 1542409706888.

FranzBusch: Got you so we would need a `UTCClock.Instant` here

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
FranzBusch: I don't fully buy this yet. If the default becomes the singleton from NIO users can still do that. They just need to call `any()` on the global singleton and then pass around their own EL. How much do we expect people to do this that we need this extra singleton is the question I guess

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
FranzBusch: Yeah I would default to `nil` and basically have no logging if nothing is passed.

0xTim: I'd disagree with this. Most people (I would assume?) will be using the lambda handler directly rather than via another framework like Hummingbird or Vapor. It would be weird to deploy the simplest thing and get _no_ logging by default as that's the first place people check to see if things are working or not.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
tachyonics: I made this point above, middleware definitely does make sense for lambda (for example tracing should just be a middleware that can be used anywhere) and should just be integrated IMO. My only quasi-concern with the passed-in-writer-approach is that for it to work with middleware, the middleware protocol has to suppress `Copyable` conformance on one of its associated types - its writer type, something that isn't going to be supported in Swift 6.0.

That said I do think the argument that this is a lower-level and more flexible API makes sense.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+struct Greeting: Encodable {
+    var echoedMessage: String
+}
+
+/// A simple Lambda function that echoes the input
+let runtime = LambdaRuntime { (event: Input, context: LambdaContext) in
+    Greeting(echoedMessage: event.message)
+}
+
+try await runtime.run()
```
tachyonics, line 549: For this, the simplest use case, would it make sense to provide a static `main()` on `LambdaRuntime` that just calls its `run` method so users can use `@main` if this is all they need to do?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
adam-fowler: This seems to dovetail with the comments about making handle support throwing of errors. Do we want users to report errors, via `reportError` or via throwing an error. Given this will end the stream I think I agree with Franz.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
adam-fowler: Yep, there'll be a lot of discussion around using a TaskLocal logger and it shouldn't hold this work up.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
adam-fowler: +1 to @0xTim's point also

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+
+## Motivation
+
+### Current Limitations
+
+#### EventLoop interfaces
+
+The current API extensively uses the `EventLoop` family of interfaces from SwiftNIO in many areas. To use these
+interfaces correctly though, it requires developers to exercise great care and be aware of certain details such as never
+running blocking code on the same `EventLoop` the library uses. Developers also need to understand the various transform
```
weissi, line 17: same applies to async/await. I'd scratch this

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+Below are explanations for all types that we want to use in AWS Lambda Runtime v2.
+
+### LambdaResponseWriter
+
+We will introduce a new `LambdaResponseWriter` protocol. It is used in the new `StreamingLambdaHandler` (defined below),
+which is the new base protocol for the `LambdaRuntime` (defined below as well).
+
+```swift
+/// A writer object to write the Lambda response stream into
+public protocol LambdaResponseWriter: ~Copyable {
```
weissi, line 193: I think making this `~Copyable` will lead to a load of issues. Can we have a motivation on why we allow non-copyable types to implement this? If we allow non-copyable types here, then anything accepting `LambdaResponseWriter`s needs to support this which sounds like a tall ask.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: Agree. This is unstructured concurrency and given that the user controls `finish`, they know when the response is done.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+```
+
+### StreamingLambdaHandler
+
+The new `StreamingLambdaHandler` protocol is the base protocol to implement a Lambda function. Most users will not use
+this protocol and instead use the `LambdaHandler` protocol defined in the **Codable Support** section.
+
+```swift
+/// The base StreamingLambdaHandler protocol
+public protocol StreamingLambdaHandler: ~Copyable {
```
weissi, line 258: I think making this `~Copyable` will lead to a load of issues. Can we have a motivation on why we allow non-copyable types to implement this? If we allow non-copyable types here, then anything accepting `StreamingLambdaHandler `s needs to support this which sounds like a tall ask.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+
+2. Accepts _any_ encoder and decoder object conforming to the `LambdaEventDecoder` and `LambdaOutputEncoder` protocols:
+
+#### LambdaEventDecoder and LambdaOutputEncoder protocols
+
+```swift
+public protocol LambdaEventDecoder {
+    /// Decode the ByteBuffer representing the received event into the generic type T
+    /// the handler will receive
+    func decode<T: Decodable>(_ type: T.Type, from buffer: ByteBuffer) throws -> T
```
weissi, line 429: can't we give type variables nice names like normal variables? `T` could maybe be `Event`?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+}
+
+public protocol LambdaOutputEncoder {
+    /// Encode the generic type T the handler has produced into a ByteBuffer
+    func encode<T: Encodable>(_ value: T, into buffer: inout ByteBuffer) throws
```
weissi, line 434: same here, nice variable maybe? `encode<Output: Encodable>...`?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+
+#### ClosureHandler
+
+This handler is generic over any `Event` type conforming to `Decodable` and any `Output` type conforming to `Encodable`
+or `Void`.
+
+```swift
+public struct ClosureHandler<Event, Output>: LambdaHandler {
+    /// Initialize with a closure handler over generic Input and Output types
+    public init(body: @escaping (Event, LambdaContext) async throws -> Output) where Output: Encoda
```
weissi, line 501: nit: some bytes missing on this line (ends in `Encoda`)

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+    /// response chunks on the stream.
+    public init(_ stream: @escaping sending (Writer) async throws -> ())
+}
+```
+
+The `StreamingLambdaHandler` would look like this:
+
+```swift
+/// The base LambdaHandler protocol
+public protocol StreamingLambdaHandler: ~Copyable {
```
weissi, line 629: same question around allowing `~Copyable` types here.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: Will we still have the ability to run everything in just one thread (the main thread)? I think for many lambdas you'll never want to fork off a second thread at all. NIO thread & async/await should just run in the main thread (but as global default executor not as `@MainActor`)

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: After thinking about this more, I think we should just make the handler method throwing. In this case we will do the error reporting, which means users don't have to deal with it, which in turn means we should remove this method here.

@FranzBusch, @tachyonics, @adam-fowler Can you get behind this?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: @weissi this would not be unstructured, as the underlying runtime would inject a `taskGroup` into the context and that `taskGroup` would be used to schedule the child task here. The underlying runtime would then ensure that all tasks have completed, before it would ask for more work. This approach would be necessary, if we used an approach in which the user returns a LambdaResponse:

```swift
protocol LambdaStreamingResponse {
    func handle(_ event: ByteBuffer, context: LambdaContext) async throws -> LambdaResponse
}
```

However this _is_ needed for the `LambdaHandler` API:

```swift
public protocol CodableLambdaHandler: ~Copyable {
    associatedtype Event
    associatedtype Output

    // only way to schedule background work that can continue after returning the Output here
    // is by having the addBackgroundTask on `LambdaContext` here.
    func handle(_ event: Event, context: LambdaContext) async throws -> Output
}
```

@FranzBusch given this, do we want to have two different `LambdaContext`s? Or are we fine with keeping the `LambdaContext.addBackgroundTask` even in the situation where it isn't really needed? I think I would opt for keeping it (so that we have it for the 99% use-case)

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
adam-fowler: The only issue with that is the user can't customise their response to the error, they might want to report the error differently if they haven't started streaming response data versus being in the middle of streaming a response.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: Yes. Sadly we don't have that yet. Only way out would be if Lambda built its own Clock. But I don't think we want to do that.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: @fabianfett I'm pretty sure that it's is unstructured, it just manages to now use a `Task {` to achieve it. It passes the background work into a reference which will then be owned by something _further up the stack_.

The test is easy, if you write this code

```
do {
    thing.addBackgroundWork { print("AAAA") }
} while false
print("BBBB")
```

Is there any way in which `AAAA` could print _after_ `BBBB`? If so: unstructured.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: As mentioned above: https://github.com/swift-server/swift-aws-lambda-runtime/pull/339#discussion_r1711150410

You convinced me that this should `throw`. 

- If an error is thrown before the first write happened, we will use the error endpoint. 
- If an error is thrown after the write has completed, we will log the error.
- If an error is thrown while sending the response we use trailing headers to report the error.

@tachyonics What we hand to the user as a Non-Copyable type can just be a wrapper around what we use internally :)

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: Structured Concurrency means that when you reach the next `}`, then all the work you have scheduled must have terminated or wait for it. There's no going further until that's complete.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
sebsto: Beware that [DispatchWallTime](https://developer.apple.com/documentation/dispatch/dispatchwalltime) has microsecond precision while the Lambda API uses milliseconds. The implementation will have to convert units

weissi: @fabianfett one thing to be aware of though: There's no (documented) way to get the actual time out of a `DispatchWallTime`. The `.rawValue` is opaque and is going backwards, so _not_ UNIX time :)

```
  8> let wt = DispatchWallTime.now() 
wt: Dispatch.DispatchWallTime = {
  rawValue = 16723545949480398616
}
  9> let wt2 = DispatchWallTime.now()
wt2: Dispatch.DispatchWallTime = {
  rawValue = 16723545943513136616
}
 10> wt2 > wt
$R4: Bool = true
 11> wt2.rawValue > wt.rawValue
$R5: Bool = false
```

See how `wt2 > wt` as expected (I asked for `wt2` after `wt`) but  `wt2.rawValue > wt.rawValue == false`!!

So you can only really use `DispatchWallTime` if you use Dispatch itself to schedule events. You cannot convert it (in a documented way) to a timestamp or so.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: > I don't fully buy this yet. If the default becomes the singleton from NIO users can still do that. They just need to call `any()` on the global singleton and then pass around their own EL. How much do we expect people to do this that we need this extra singleton is the question I guess

@FranzBusch to do this correctly users would need to do the following:

```swift
let el = NIOSingleton.posix.any()
let runtime = LambdaRuntime(handler: MyHandler(), eventLoop: el)
try await withTaskExecutor(el) {
    try await runtime.run()
}
```

Not sure if explaining this dance is worth it. What is the big problem with having this property. We don't spawn a different ELG. We just make sure that the _same_ EL is used.

> Will we still have the ability to run everything in just one thread (the main thread)? I think for many lambdas you'll never want to fork off a second thread at all. NIO thread & async/await should just run in the main thread (but as global default executor not as `@MainActor`)

Currently I have not planned to keep this functionality. The reason for this is, that users have tripped over this when using Lambda because of course they import some code that uses `DispatchMain` and `@MainActor`. I think the best way out of this would be for NIO to provide an async/await adapted API for this, so that users who are interested in this behavior can easily achieve it:

```swift
try await withCurrentThreadAsEventLoop { el in
    // task executor preference is auto set to el here
    let runtime = LambdaRuntime(handler: MyHandler(), eventLoop: el)
    try await runtime.run()
}
```

@weissi Do you think this would be a reasonable compromise? Make users opt in?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: @FranzBusch Can you get behind the arguments and rest your case?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: > they import some code that uses DispatchMain and @MainActor.

Uuhg, on the server side? That will be super slow, right?

Maybe we could actually make a NIO EL on the main thread host _both_ the MainActor and the global concurrency executor.

> @weissi Do you think this would be a reasonable compromise? Make users opt in?

I do think opt-in is reasonable but we should have tests that indeed everything (async/await code and EL code) runs on the main thread if you opt in.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: @weissi Okay I get your point. 

But the point of adding this API is that users explicitly ask for it #334 and AWS makes a deal about supporting background work: [Running code after returning a response from an AWS Lambda function](https://aws.amazon.com/blogs/compute/running-code-after-returning-a-response-from-an-aws-lambda-function/)

In Lambdas case this can be used to flush metrics or logs _after_ the response has returned. So from a structured concurrency standpoint, yes this is flawed! But it is a useful feature nonetheless. 

We do use structured concurrency in the level above to ensure all background tasks finish before we get the next invocation.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: > Maybe we could actually make a NIO EL on the main thread host both the MainActor and the global concurrency executor.

@weissi That would require an Evolution Proposal I assume? So nothing to wait for here, I guess. We will add those tests :)

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+>: StreamingLambdaHandler where Handler.Output == Output, Handler.Event == Event {
+
+    /// Register the concrete handler, encoder, and decoder.
+    public init(
+        handler: Handler,
+        encoder: Encoder,
+        decoder: Decoder
+    ) where Output: Encodable
+
+    /// For handler with a void output -- the user doesn't specify an encoder.
```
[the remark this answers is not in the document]
fabianfett: Interesting point, we didn't support this before. Are there Lambdas without input?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: Can't we express this like so:

```
func handle() {
    try await writeResponse()
    try await flushMetrics()
}
```

That way we can run code after writing a response and we're also structured. You could also provide a library function that performs certain jobs after the response has been written.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: Maybe you should create a `LambdaUTCTimeClock` type that conforms to all the new hotness but implements the clock. With the hope that one day you can deprecate it and to `typealias LambdaUTCTimeClock = UTClock` with a bit of luck.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: @fabianfett is there nothing like `swift_task_enqueueGlobal_hook` for the main queue on Linux at least?

weissi: But even without `@MainActor` support, this is still super useful and we should have the tests at least that NIO's I/O code and the user's `async/await` code can run on the same thing.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: @weissi agree that this works for the Streaming case. However I think we should keep the easy Codable support API:

```
public protocol LambdaHandler: ~Copyable {
    associatedtype Event
    associatedtype Output

    // only way to schedule background work that can continue after returning the Output here
    // is by having the addBackgroundTask on `LambdaContext` here.
    func handle(_ event: Event, context: LambdaContext) async throws -> Output
}
```

In that case you can't `flushMetrics()` after the response was returned. And the `addBackgroundTask()` would be handy here.

weissi: why don't you make it

```
public protocol LambdaHandler: ~Copyable {
    associatedtype Event
    associatedtype Output

    func handle(_ event: Event, context: LambdaContext) async throws -> Output

    func performPostHandleWork(...) async throws -> Void
} 

extension LambdaHandler {
    func performPostHandleWork(...) async throws -> Void {
        // nothing to do
    }
}
```

that way the users can implement post-handle work

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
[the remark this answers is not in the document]
sebsto: I think the answer is "no" (or at least an empty JSON `{}` 
Source: the [runtime API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html) and the [invoke API](https://docs.aws.amazon.com/lambda/latest/api/API_Invoke.html). 

I asked the team to confirm

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+
+An implementation that sends the number 1 to 10 every 500ms could look like this:
+
+```swift
+struct SendNumbersWithPause: StreamingLambdaHandler {
+    func handle(
+        _ event: ByteBuffer,
+        responseWriter: consuming some LambdaResponseWriter,
+        context: LambdaContext
+    ) async {
```
czechboy0, line 281: Out of curiosity: at which point is the HTTP response started/the head sent? Before the user handler is invoked? Lazily right before the first `responseWriter.write(...)` call?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+        for i in 1...10 {
+            responseWriter.write(ByteBuffer(string: #"\#(i)\n\r"#))
+            try? await Task.sleep(for: .milliseconds(500))
+        }
+        responseWriter.finish()
```
czechboy0, line 286: What happens if the user doesn't call `finish` explicitly?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: This would be lazily before the first `write()`. If the user calls `writeAndFinish()` just once, we would send the response back in non streaming mode (if it fits into the 6MB limit).

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
FranzBusch: They can still do this by doing a custom do/catch. It just makes the default case easier

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
FranzBusch: The problem with splitting it across two methods is if the post work needs context from the initial work. However, you could also just pass a writer that takes the output to the `handle` method to allow for background work. It's not as nice as returning the Output from the method.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: `LambdaResponseWriter` is not a protocol that users would need to implement. It is marked `NonCopyable` as we want to enforce through the compiler that users can `write()` after they have called `finish()`.

> If we allow non-copyable types here, then anything accepting `LambdaResponseWriter`s needs to support this which sounds like a tall ask.

Would you mind elaborating on this a bit?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: > Can we have a motivation on why we allow non-copyable types to implement this?

It is not necessary to have a non copyable Handler. We just allow it. Tbh. the motivation here is mostly, because we can.

> If we allow non-copyable types here, then anything accepting `StreamingLambdaHandler`s needs to support this which sounds like a tall ask.

I think the only thing that should accept this protocol is `LambdaRuntime` and we have support for non copyable already implemented in our prototype. What else would consume a `StreamingLambdaHandler`?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: agree! Copy and paste from old code. Should have fixed. @aryan-25 can you fix this please?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: Let's discuss here: https://github.com/swift-server/swift-aws-lambda-runtime/pull/339#discussion_r1711515316 once we have a solution we will fix or resolve here.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: > What else would consume a StreamingLambdaHandler?

The user's tests?

> we have support for non copyable already implemented in our prototype.

Sure, but it could cause problems further down the line, no? Like if you figure out that you will need to copy it

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: What about

```
responseWriter.withResponseStream { stream in
   try await stream.write(...)
   try await stream.write(...)
   try await stream.write(...)
   try await stream.write(...)
   try await stream.write(...)
}
```

where the `head` is sent on `{` and the `finish` is sent on `}`? That way things can't go wrong. If you `throw` that'd mean `reportError`.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: > The problem with splitting it across two methods is if the post work needs context from the initial work. 

but we can pass the context, no?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: That'd kinda answer both of @czechboy0's questions pretty clearly. No implicit, lazy `head` anymore and no way to forget `finish`.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: > The user's tests?

I think we would provide a `TestRuntime` that does that for the user.

 > Sure, but it could cause problems further down the line, no? Like if you figure out that you will need to copy it

Makes sense. But in this case requirements would have changed so much that we should investigate a v3.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: I don't think we want to integrate the swift-middleware library directly. I think if users have those needs they should consider using hummingbird/smoke/vapor in lambda mode, which then brings the option to use middleware.

This shall not mean that writing a middleware can be impossible. We should allow users to write middleware that they can stack. 

Also I think swift-middleware is currently not targeted for anytime soon. So in order to make progress we should not depend on it. Integrating swift-middleware (as is) on top of this proposed API is absolutely possible and therefore should not be considered a blocker.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+}
+```
+
+#### Verbose Codable support
+
+In the current API, there are extensions and Codable wrapper classes for decoding events and encoding computed responses
+for _each_ different handler protocol and for both `String` and `JSON` formats. This has resulted in a lot of
+boilerplate code which can very easily be made generic and simplified in v2.
+
+### New features
```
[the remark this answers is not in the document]
fabianfett: Let's do the discussion [here](https://github.com/swift-server/swift-aws-lambda-runtime/pull/339/files#r1711636115)

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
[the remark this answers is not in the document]
weissi: Why does the  context need to hold the `TaskGroup`? I smell structured concurrency violations :)

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: @fabianfett Not sure I understand what "a task that starts immediately but is allowed to run longer" is. Could you provide some example code?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: > @weissi would you build the sleeping on top of dispatch?

No.

I'd just create a custom Clock type (because the language lacks `UTCClock` for some reason). Once you have your own clock (which you could for example back by `clock_gettime` and `struct timespec` but hidden as implementation detail).

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,667 @@
...
+    consuming func writeAndFinish(_ buffer: ByteBuffer) async throws
+    /// Report an error in the response stream
+    consuming func reportError(_ error: any Error) async throws
+}
+```
+
+### LambdaContext
+
+`LambdaContext` will be largely unchanged, but the `eventLoop` property as well as the `allocator` property (of type
+`ByteBufferAllocator`) will be removed.
```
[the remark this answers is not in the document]
weissi: I think it's fine to remove this. The original idea was that you could use more efficient allocators for ByteBuffers but we never actually used that. Swift itself is still not able to use alternative allocators (except if you globally set `malloc`/`free` to a different one) so I don't think this was as useful as we thought it would be (outside of ByteBuffer tests).

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: Really not sure about that. I don't think it gives you any benefit but it restricts your code's evolution. Don't see the point :).

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: `~Copyable` doesn't help you if you _don't_ call `finish`. You can't mark `deinit` as `@available(*, unavailable) deinit {}`. It does prevent double-calling but it also comes with massive strings attached, namely that the user can't copy it if the type's actually `~Copyable`.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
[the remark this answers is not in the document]
sebsto: I confirm a payload is required. Minimum payload is `{}`

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: An API call that 

1. does a short computation which is returned to you immediately 
2. spawns a longer computation, which's result is send to you via mail once it finishes (can take up to 15min)

Both should start as soon as the API is invoked. We can potentially question the API design here. But this is possible with Lambda and we should allow users to do this, if they opt to.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
fabianfett: How would you implement the required sleeping methods though?
https://developer.apple.com/documentation/swift/clock/sleep(for:tolerance:)

weissi: You can just use the normal `Task.sleep` on the non-suspending clock for that. UTC time and the non-suspending clock run at the same speed

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
weissi: Sure, but don't you provide a `~Copyable` `LambdaResponseWriter` to the user?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
tachyonics: We can debate whether we think the swift-middleware library should be integrated directly but I do agree it shouldn't be considered a blocker for this work/release.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
tachyonics: Probably the most consistent approach for @fabianfett's use case is to provide a wrapping `CodableResponseWriter` (potentially with just a `writeAndFinish` API) while allowing work to continue past the result being returned to the control plane. It is not as simple as returning the result directly. We **could** provide both forms but that does add its own complexity to the library overall.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
tachyonics: Wouldn't it look like-

```swift
@main
LambdaRuntime { (event: Input, context: LambdaContext) in
    Greeting(echoedMessage: event.message)
}
```

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
tachyonics: The head has to be sent on the first write because the control plane has different endpoints for a complete failure vs a successful response or a midstream error. So we need to wait until at least a partial response to determine what endpoint to send the response to.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
czechboy0: So it is safe to call `finish()` multiple times?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
tachyonics: The semantics of the writer type we want is of mutable non-shared state. Isn't a non-copyable type the correct mechanism to express these semantics?

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
FranzBusch: The fact that I propose to default the proposed task local logger to something is making it clear to me that this should also default to a standard logger. In the future, if we do the task local we can just default this logger to the task local value.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
FranzBusch: `finish` is consuming so you can't call it multiple times

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
Joannis: +1 on Tim's comment

Joannis:
I don't have further input other than what was already mentioned

fabianfett:
@aryan-25 Thanks so much for continuing to push this forward! 🙏🚀

tachyonics (approved):
Thanks @aryan-25, this is looking great!

I am fine with having the three handler protocols; we will just need to make sure that each is documented with what use case they are recommended for and for the writer protocols what it means for the control plane to finish the response but delay returning from the handle function.

0xTim (approved):
This is looking good!

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,890 @@
...
+### Providing ownership of main() and support for swift-service-lifecycle
+
+- Instead of conforming to a handler protocol, users can now create a `LambdaRuntime` by passing in a handler closure.
+- `LambdaRuntime` conforms to `ServiceLifecycle.Service` by implementing a `run()` method that contains initialization
+  and graceful shutdown logic.
+- This allows the lifecycle of the `LambdaRuntime` to be managed with `swift-service-lifecycle` _alongside_ and in the
+  same way the lifecycles of the required services are managed, e.g.
+  `try await ServiceGroup(services: [postgresClient, ..., lambdaRuntime], ...).run()`.
+- Dependencies can now be injected into `LambdaRuntime` — `swift-service-lifecycle` guarantees that the services will be
+  initialized _before_ the `LambdaRuntime`'s `run()` function is called.
```
adam-fowler, line 148: What guarantees are these? ServiceLifecycle doesn't currently have any initialization order guarantees, unless I missed something.

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,890 @@
...
+
+#### LambdaHandlerAdapter
+
+Any handler conforming to `LambdaHandler` can be conformed to `LambdaWithBackgroundProcessingHandler` through
+`LambdaHandlerAdapter`.
+
+```swift
+/// Wraps an underlying handler conforming to ``LambdaHandler``
+/// with ``LambdaWithBackgroundProcessingHandler``.
+public struct LambdaHandlerAdapter<
```
adam-fowler, line 560: It's not clear how the Event is decoded and Output encoded in the `LambdaHandlerAdapter`.

Joannis (approved):

Sources/AWSLambdaRuntimeCore/Documentation.docc/Proposals/0001-v2-api.md
```diff
@@ -0,0 +1,905 @@
...
+manually initialize services, leading to the issue of the user then perhaps forgetting to gracefully shutdown the
+initialized services:
+
+```swift
+struct MyLambda: LambdaHandler {
+    let pgClient: PostgresClient
+
+    init(context: AWSLambdaRuntimeCore.LambdaInitializationContext) async throws {
+        /// Instantiate service
+        let client = PostgresClient(configuration: ...)
```
Andrea-Scuderi, line 85: Using Postgres as example about something to use with AWS Lambda is quite misleading for the users for two reasons:
- The main point of using Lambda is autoscaling, this is usually something very difficult to obtain from a RDS like Postgres and could require connection pooling features. Ideally, the use case we should present is the connection to a NO-SQL database such as DynamoDB or MongoDB.
- In addition giving the nature of connection oriented protocol of the DB, opening the client connection during the Lambda's initialisation could lead to connection time-out due to the life cycle of the Lambda. There is no guarantee on how long a lambda will be allocated after the first usage. At some point the connection of the DB will be closed from the DB side, leaving subsequent invocation with a closed connection from the Lambda side.

fabianfett: > The main point of using Lambda is autoscaling, this is usually something very difficult to obtain from a RDS like Postgres and could require connection pooling features. Ideally, the use case we should present is the connection to a NO-SQL database such as DynamoDB or MongoDB.

While I generally agree, neither Soto nor MongoDB integrate with swift-service-lifecycle today. The goal of this new API is to integrate better with the shared building blocks of the Swift server ecosystem. Since swift-service-lifecycle is an important one, we opted for PostgresNIO as it integrates with it today.

> In addition giving the nature of connection oriented protocol of the DB, opening the client connection during the Lambda's initialisation could lead to connection time-out due to the life cycle of the Lambda.

PostgresClient can lazily create connections. This is totally configurable.

> At some point the connection of the DB will be closed from the DB side, leaving subsequent invocation with a closed connection from the Lambda side.

That's totally fine, the connection pool will automatically reconnect for the user.

fabianfett:
The discussion here has settled. @aryan-25 addressed most of the feedback. From my point the only open issue concerns the use of `DispatchWallTime` in `LambdaContext`. I've created an extra ticket for this #384.

Last call to all participants: I intend to merge this PR on Monday. Is there any other unaddressed feedback besides #384, that we should create an extra ticket for?

sebsto (approved):

PR #344, opened 2024-08-28
Introduce LambdaRuntimeClientProtocol

aryan-25:
### Motivation:

To start implementing the v2 API (#339), we first want to introduce the `LambdaRuntimeClientProtocol`. Objects conforming to this protocol can then be used with `LambdaRuntime`.

### Modifications:

- Introduced `LambdaRuntimeClientProtocol` and the associated `LambdaResponseStreamWriter` protocol.
- Renamed the existing `Invocation` struct to `InvocationMetadata` since it does not contain the actual invocation event data itself.
- Added an `Invocation` struct which contains `InvocationMetadata` and `ByteBuffer`.
 
### Result:

`LambdaRuntime` can be tested more easily through a mock client conforming to `LambdaRuntimeClientProtocol`.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/LambdaRuntimeClientProtocol.swift
+++ b/Sources/AWSLambdaRuntimeCore/LambdaRuntimeClientProtocol.swift
@@ -1,0 +1,33 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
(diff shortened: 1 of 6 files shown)


Sources/AWSLambdaRuntimeCore/LambdaRuntimeClientProtocol.swift
```diff
@@ -0,0 +1,33 @@
...
+
+package protocol LambdaRuntimeClientProtocol {
+    associatedtype Writer: LambdaResponseStreamWriter
+
+    func nextInvocation() async throws -> (Invocation, Writer)
+}
+
+package struct Invocation {
+    package let metadata: InvocationMetadata
+    package let event: ByteBuffer
```
fabianfett, line 32:
```suggestion
    package var metadata: InvocationMetadata
    package var event: ByteBuffer
```

fabianfett: please also add a `package` initializer here.

fabianfett (approved):

PR #345, opened 2024-08-28
Update LambdaContext

aryan-25:
### Motivation:

As outlined in the v2 API proposal (#339), the `eventLoop` and `allocator` properties are to be removed from `LambdaContext`. The `detachedBackgroundTask(_:)` function and the associated `tasks` property are also not needed since the v2 API will introduce handlers that support background task execution through a writer passed to the handler.

### Modifications:

Introduced `NewLambdaContext`: exactly the same as `LambdaContext` but without the `eventLoop`, `allocator`, and `tasks` properties, and the `detachedBackgroundTask(_:)` function.

`LambdaContext` was not modified directly in order to keep the current functionality untouched.

### Result:

`NewLambdaContext` can be used to continue implementing the new API.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/NewLambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/NewLambdaContext.swift
@@ -20,7 +20,7 @@ import NIOCore
 
 /// Lambda runtime context.
 /// The Lambda runtime generates and passes the `LambdaContext` to the Lambda handler as an argument.
-public struct NewLambdaContext: CustomDebugStringConvertible, Sendable {
+package struct NewLambdaContext: CustomDebugStringConvertible, Sendable {
     final class _Storage: Sendable {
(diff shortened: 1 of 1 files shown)


Sources/AWSLambdaRuntimeCore/NewLambdaContext.swift
```diff
@@ -0,0 +1,140 @@
...
+
+import Dispatch
+import Logging
+import NIOCore
+
+// MARK: - Context
+
+/// Lambda runtime context.
+/// The Lambda runtime generates and passes the `LambdaContext` to the Lambda handler as an argument.
+public struct NewLambdaContext: CustomDebugStringConvertible, Sendable {
```
fabianfett, line 23: please replace all public with package for now

fabianfett (approved):

PR #346, opened 2024-08-28
[CI] Set Swift 6 container image for API breakage + Documentation check

aryan-25:
### Motivation:

Currently, the API Breakage and Documentation checks are failing since these checks run on the `swift:5.10-noble` image while the library requires Swift 6.

With the new container image parameters introduced in https://github.com/apple/swift-nio/pull/2851, the image for the API Breakage and Documentation checks can be configured to `swiftlang/swift:nightly-6.0-jammy`.

### Modifications:

Set the `api_breakage_check_container_image` and `docs_check_container_image` parameters to `swiftlang/swift:nightly-6.0-jammy`.

### Result:

The API Breakage and Documentation checks will no longer immediately fail.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/NewLambdaContext.swift
+++ b/Sources/AWSLambdaRuntimeCore/NewLambdaContext.swift
@@ -1,0 +1,140 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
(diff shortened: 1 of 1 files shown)


fabianfett (approved):

fabianfett:
Even though docs doesn't work currently, let's get this merged so that API breakage check works again.

PR #347, opened 2024-08-29
Add runLoop function

aryan-25:
### Motivation:

As part of the implementation of the new v2 API (#339), the `LambdaRuntimeClientProtocol` was introduced in  #344.

Building upon it, this PR introduces a static `runLoop` function that implements the runtime loop (fetch the next event -> handle the event -> report the result -> repeat).

This function is generic over an object conforming to `LambdaRuntimeClientProtocol` to also allow for mock runtime clients to be used in tests.

### Modifications:

- Add the `StreamingLambdaHandler` protocol as detailed in the v2 API proposal (#339).
- Add the static `runLoop` function.

### Result:

Runtime loop implemented. The library can also be tested more easily, e.g. through a mock runtime client object passed to `runLoop`.

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/NewLambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/NewLambda.swift
@@ -24,17 +24,17 @@ package protocol StreamingLambdaHandler {
     ) async throws
 }
 
-package enum NewLambda {
+extension Lambda {
     package static func runLoop<RuntimeClient: LambdaRuntimeClientProtocol, Handler>(
         runtimeClient: RuntimeClient,
         handler: Handler,
         logger: Logger
-    ) async throws
-    where Handler: StreamingLambdaHandler {
+    ) async throws where Handler: StreamingLambdaHandler {
+        var handler = handler
+
         while !Task.isCancelled {
             let (invocation, writer) = try await runtimeClient.nextInvocation()
 
-            var handler = handler
             do {
                 try await handler.handle(
                     invocation.event,
--- a/Sources/AWSLambdaRuntimeCore/NewLambda.swift
+++ b/Sources/AWSLambdaRuntimeCore/NewLambda.swift
@@ -1,0 +1,56 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Dispatch
+import Logging
+import NIOCore
+
+package protocol StreamingLambdaHandler {
+    mutating func handle(
+        _ event: ByteBuffer,
+        responseWriter: some LambdaResponseStreamWriter,
+        context: NewLambdaContext
+    ) async throws
+}
+
+package enum NewLambda {
+    package static func runLoop<RuntimeClient: LambdaRuntimeClientProtocol, Handler>(
+        runtimeClient: RuntimeClient,
+        handler: Handler,
+        logger: Logger
+    ) async throws
+    where Handler: StreamingLambdaHandler {
+        while !Task.isCancelled {
+            let (invocation, writer) = try await runtimeClient.nextInvocation()
+
+            var handler = handler
+            do {
+                try await handler.handle(
+                    invocation.event,
+                    responseWriter: writer,
+                    context: NewLambdaContext(
+                        requestID: invocation.metadata.requestID,
+                        traceID: invocation.metadata.traceID,
+                        invokedFunctionARN: invocation.metadata.invokedFunctionARN,
+                        deadline: DispatchWallTime(millisSinceEpoch: invocation.metadata.deadlineInMillisSinceEpoch),
+                        logger: logger
+                    )
+                )
+            } catch {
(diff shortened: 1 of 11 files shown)


Sources/AWSLambdaRuntimeCore/NewLambda.swift
```diff
@@ -0,0 +1,56 @@
...
+package protocol StreamingLambdaHandler {
+    mutating func handle(
+        _ event: ByteBuffer,
+        responseWriter: some LambdaResponseStreamWriter,
+        context: NewLambdaContext
+    ) async throws
+}
+
+package enum NewLambda {
+    package static func runLoop<RuntimeClient: LambdaRuntimeClientProtocol, Handler>(
```
fabianfett, line 28: please add tests for this function :)

Sources/AWSLambdaRuntimeCore/NewLambda.swift
```diff
@@ -0,0 +1,56 @@
...
+        runtimeClient: RuntimeClient,
+        handler: Handler,
+        logger: Logger
+    ) async throws
+    where Handler: StreamingLambdaHandler {
+        while !Task.isCancelled {
+            let (invocation, writer) = try await runtimeClient.nextInvocation()
+
+            var handler = handler
```
fabianfett, line 37: we should move this outside the while loop

Sources/AWSLambdaTesting/LambdaMockClient.swift
```diff
@@ -0,0 +1,266 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import AWSLambdaRuntimeCore
+import Foundation
+import Logging
+import NIOCore
+
+package struct LambdaMockWriter: LambdaResponseStreamWriter {
```
fabianfett, line 20: I think this should be internal (aka no access control modifier). No need for package within a test target.

Sources/AWSLambdaTesting/LambdaMockClient.swift
```diff
@@ -0,0 +1,266 @@
...
+    package static let cannotCallNextEndpointWhenAlreadyWaitingForEvent = LambdaError(
+        code: .cannotCallNextEndpointWhenAlreadyWaitingForEvent
+    )
+    package static let cannotCallNextEndpointWhenAlreadyProcessingAnEvent = LambdaError(
+        code: .cannotCallNextEndpointWhenAlreadyProcessingAnEvent
+    )
+    package static let cannotReportResultWhenNoEventHasBeenProcessed = LambdaError(
+        code: .cannotReportResultWhenNoEventHasBeenProcessed
+    )
+}
```
fabianfett, line 77: We don't need this dance in tests. just use an enum

Sources/AWSLambdaTesting/LambdaRunLoopTests.swift
```diff
@@ -0,0 +1,50 @@
...
+        ) async throws {
+            try await responseWriter.writeAndFinish(event)
+        }
+    }
+
+    let mockClient = LambdaMockClient()
+    let mockEchoHandler = MockEchoHandler()
+
+    @Test func testRunLoop() async throws {
+        _ = Task { () in
```
fabianfett, line 37: this runs forever. please cancel the running loop once the invocation has succeeded

Tests/AWSLambdaRuntimeCoreTests/LambdaRunLoopTests.swift
```diff
@@ -0,0 +1,53 @@
...
+                runtimeClient: self.mockClient,
+                handler: self.mockEchoHandler,
+                logger: Logger(label: "RunLoopTest")
+            )
+        }
+
+        let inputEvent = ByteBuffer(string: "Test Invocation Event")
+        let response = try await self.mockClient.invoke(event: inputEvent)
+
+        runLoopTask.cancel()
```
fabianfett, line 49: please use a `TaskGroup` here.

Tests/AWSLambdaRuntimeCoreTests/LambdaRunLoopTests.swift
```diff
@@ -0,0 +1,53 @@
...
+
+        #expect(response == inputEvent)
+    }
```
fabianfett, line 52: add a test that ensures reportError is called if an error is thrown

Tests/AWSLambdaRuntimeCoreTests/LambdaMockClient.swift
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    func writeAndFinish(_ buffer: ByteBuffer) async throws {
+        try await self.underlying.write(buffer)
+        try await self.underlying.finish()
+    }
+
+    func reportError(_ error: any Error) async throws {
+        try await self.underlying.write(ByteBuffer(string: "\(error)"))
+        try await self.underlying.finish()
```
fabianfett, line 42: I think this is wrong. It should trigger a reportError function. And that reportError function should fail the Continuation in `handlerIsProcessing`

Tests/AWSLambdaRuntimeCoreTests/LambdaMockClient.swift
```diff
@@ -0,0 +1,294 @@
...
+            case .waitingForNextEvent(let eventArrivedHandler):
+                self.state = .initialState
+                return .cancelContinuation(eventArrivedHandler)
+            }
+        }
+
+        mutating func failProcessing() -> FailProcessingAction {
+            switch self.state {
+            case .initialState, .waitingForNextEvent:
+                return .none
```
fabianfett, line 197: we should crash here, because this would mean, that we send a failProcessing at the wrong time.

fabianfett (approved):
Great change! Thanks for pushing through!

PR #358, opened 2024-09-05
Make v2 API functions and types public

aryan-25:
### Motivation:

As part of the implementation of the v2 API (#339), we landed new additions with the `package` access modifier so that they could not be mistakenly used while work was still in progress.

We now want to make the new additions introduced `public`.

### Modifications:

- Make new v2 functions and types `public`.

### Result:

The new API can be used by adopters.


Sources/AWSLambdaRuntimeCore/Lambda+JSON.swift
```diff
@@ -67,7 +69,7 @@ package struct LambdaHandlerAdapter<
     ///   - outputWriter: The writer to write the computed response to.
     ///   - context: The ``LambdaContext`` containing the invocation's metadata.
     @inlinable
-    package func handle(
+    public func handle(
```
fabianfett, line 72: the init needs a `public`

Sources/AWSLambdaRuntimeCore/Lambda+JSON.swift
```diff
@@ -136,7 +137,7 @@ package struct LambdaCodableAdapter<
 }
 
 /// A ``LambdaResponseStreamWriter`` wrapper that conforms to ``LambdaResponseWriter``.
-package struct LambdaCodableResponseWriter<Output, Encoder: LambdaOutputEncoder, Base: LambdaResponseStreamWriter>:
+public struct LambdaCodableResponseWriter<Output, Encoder: LambdaOutputEncoder, Base: LambdaResponseStreamWriter>:
```
fabianfett, line 140: can we drop the `Codable` name here.

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -108,15 +108,15 @@ package struct LambdaContext: CustomDebugStringConvertible, Sendable {
         )
     }
 
-    package func getRemainingTime() -> TimeAmount {
+    public func getRemainingTime() -> TimeAmount {
```
fabianfett, line 111: we should change that to return `Duration`

Sources/AWSLambdaRuntimeCore/Lambda+JSON.swift
```diff
@@ -16,7 +16,7 @@ import NIOCore
 
 /// The protocol a decoder must conform to so that it can be used with ``LambdaCodableAdapter`` to decode incoming
```
fabianfett, line 17: Rename file to `Lambda+Codable.swift`

PR #361, opened 2024-09-06
Fix error in Plugin.swift

aryan-25:
Fixes error:
`AWSLambdaPackager/Plugin.swift:276: Fatal error: Unexpectedly found nil while unwrapping an Optional value`


sebsto (approved):

sebsto:
@aryan-25 I propose to close this in favour of https://github.com/swift-server/swift-aws-lambda-runtime/pull/363 that includes that change.

PR #363, opened 2024-09-06
fix deprecation warning for Plugin's Path

sebsto:
Replace the usage of Plugin's `Path` with `URL`

### Motivation:

`Path` is deprecated in Swift 6

### Modifications:

- Replace the use of `Path` with `URL` and adjust path components management accordingly 

### Result:

no more compilation warnings when compiling the `archive` plugin on Swift 6 

### Command to test the plugin with Swift 6 

```sh
	swift package archive \
		--base-docker-image swiftlang/swift:nightly-6.0-amazonlinux2 \
		--disable-sandbox
```


sebsto:

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -12,11 +12,18 @@
...
 import Synchronization
 
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import struct Foundation.URL
+import class Foundation.ProcessInfo
+import class Foundation.FileManager
+import struct Foundation.ObjCBool
+#endif
```
fabianfett, line 25: I'm afraid we need full Foundation here. This is a built time dependency. So therefore we don't really care.
```suggestion
import Foundation
```

sebsto: Correct - just realized that when testing the CI compilation (that failed) - I just pushed a change

Plugins/AWSLambdaPackager/PluginUtils.swift
```diff
@@ -0,0 +1,145 @@
...
+        arguments: [String],
+        customWorkingDirectory: URL? = nil,
+        logLevel: ProcessLogLevel
+    ) throws -> String {
+        if logLevel >= .debug {
+            print("\(executable.absoluteString) \(arguments.joined(separator: " "))")
+        }
+
+        // this shared global variable is safe because we're mutating it in a dispatch group
+        // https://developer.apple.com/documentation/foundation/process/1408746-terminationhandler
```
fabianfett: I think the comment here is misleading. The important thing here is the following. We write to the variable from a single SerialDispatchQueue here. We wait until the process is run below `process.waitUntilExit()`. This means no further writes to `output` will happen. This makes it save for us to read the output.

fabianfett: We should switch to the new Foundation Process API as soon as it is available:
https://github.com/apple/swift-foundation/blob/main/Proposals/0007-swift-subprocess.md

sebsto: Changed to 
```
        // We need to use an unsafe transfer here to get the fd into our Sendable closure.
        // This transfer is fine, because we write to the variable from a single SerialDispatchQueue here.
        // We wait until the process is run below process.waitUntilExit().
        // This means no further writes to output will happen.
        // This makes it save for us to read the output
```

sebsto: Created https://github.com/swift-server/swift-aws-lambda-runtime/issues/364 to keep track of `SubProcess`

PR #368, opened 2024-09-26
add help option on archiver plugin

sebsto:
add a `--help` option to the archiver plugin 

### Motivation:

The only way to know what available options are was to look at the code 

### Modifications:

Add a `swift package archive --help` command line option to print a descriptive help text 

### Result:

```
$ swift package archive --help                                 

OVERVIEW: A SwiftPM plugin to build and package your lambda function.

REQUIREMENTS: To use this plugin, you must have docker installed and started.

USAGE: swift package --disable-sandbox archive [--help] [--verbose]
                                           [--output-directory <path>]
                                           [--products <list of products>]
                                           [--configuration debug | release]
                                           [--swift-version <version>]
                                           [--base-docker-image <docker_image_name>]
                                           [--disable-docker-image-update]
                                          

OPTIONS:
--verbose                     Produce verbose output for debugging.
--output-directory <path>     The path of the binary package.
                              (default is `.build/plugins/AWSLambdaPackager/outputs/...`)
--products <list>             The list of executable targets to build.
                              (default is taken from Package.swift)
--configuration <name>        The build configuration (debug or release)
                              (default is release)
--swift-version               The swift version to use for building. 
                              (default is latest)
                              This parameter cannot be used when --base-docker-image  is specified.
--base-docker-image <name>    The name of the base docker image to use for the build.
                              (default : swift-<version>:amazonlinux2)
                              This parameter cannot be used when --swift-version is specified.
--disable-docker-image-update Do not attempt to update the docker image
--help                        Show help information.
```

PR #379, opened 2024-09-30
Allow docker access in archiver plugin

fabianfett:
Allow plugin to use unix domain socket to interact with docker. This should eliminate the use of `--disable-sandbox` for "normal" docker installs (unchanged docker context).


sebsto:
I would rather first land https://github.com/swift-server/swift-aws-lambda-runtime/pull/368 before tackling this one.

Also, this one should include a slight change in the README and not include changes in LambdaRuntime.
Let's try to have one PR/commit per subject.

May I close it and open a separate one once #368 is merged ?

sebsto:
Closing in favour of a combined README / code change that I will submit shortly

PR #382, opened 2024-09-30
Restructure Examples folder

sebsto:
Move all v1 examples to folder 
`Examples/v1`
Keep only v2 examples in `Examples` folder


Examples/APIGateway/Sources/main.swift
```diff
@@ -0,0 +1,34 @@
...
+    (event: APIGatewayV2Request, context: LambdaContext) -> APIGatewayV2Response in
+
+    var header = HTTPHeaders()
+    context.logger.debug("HTTP API Message received")
+
+    header["content-type"] = "application/json"
+
+    // echo the request in the response
+    let data = try JSONEncoder().encode(event)
+    let response = String(data: data, encoding: .utf8)
```
fabianfett, line 29:
```suggestion
let encoder = JSONEncoder()
let runtime = LambdaRuntime {
    (event: APIGatewayV2Request, context: LambdaContext) -> APIGatewayV2Response in

    var header = HTTPHeaders()
    context.logger.debug("HTTP API Message received")

    header["content-type"] = "application/json"

    // echo the request in the response
    let data = try encoder.encode(event)
    let response = String(decoding: data, as: Unicode.UTF8.self)
```

Examples/APIGateway/Sources/main.swift
```diff
@@ -0,0 +1,34 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import AWSLambdaEvents
+import AWSLambdaRuntime
+import class Foundation.JSONEncoder
```
fabianfett, line 17: Please do the FoundationEssentials dance here. We want small binaries.
```suggestion
#if canImport(FoundationEssentials)
import FoundationEssentials
#else
import Foundation
#endif
```

Examples/v1/Benchmark/Package.swift
```diff
@@ -0,0 +1,53 @@
+// swift-tools-version:5.7
+
+import PackageDescription
```
fabianfett, line 3: Why do we want to keep the v1 examples?

fabianfett (commented):
First round

sebsto:
Done. @fabianfett do you have a hard feeling against keeping v1's ?

fabianfett:
> Done. @fabianfett do you have a hard feeling against keeping v1's ?

Kinda. When users look at the repo, they don't necessarily know which version we are currently at. This might lead to confusion. Examples for v1 are still available in the v1 branch.

sebsto:
@fabianfett There is no 1.0.0 branches, these are just tags.
I propose to create a branch from the last 1.0.0 tag and delete everything v1 specific from main.  OK ?

PR #391, opened 2024-10-08
add docker fine grained permission to the plugin

sebsto:
removes the need to `--disable-sandbox` and update docs and examples


fabianfett (approved):
Love it!

PR #397, opened 2024-10-08
Remove docker compose files + add GH action to compile examples during CI

sebsto:


.github/workflows/pull_request.yml
```diff
@@ -24,6 +24,17 @@ jobs:
             linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error"
             linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error"
 
+    integration-tests:
+        name: Integration Tests
+        # Workaround https://github.com/nektos/act/issues/1875
+        uses: apple/swift-nio/.github/workflows/swift_matrix.yml@main
```
fabianfett, line 30: I don't think we should use the matrix here. Testing examples with latest should be fine.

docker/docker-compose.yaml
```diff
@@ -34,14 +37,10 @@ services:
...
-      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/Echo &&
-      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/ErrorHandling &&
-      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/Foundation &&
-      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/JSON &&
-      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/LocalDebugging/MyLambda &&
-      LAMBDA_USE_LOCAL_DEPS=true swift test --package-path Examples/Testing
+      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/APIGateway &&
+      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/AWSSDK &&
+      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/HelloWorld &&
+      LAMBDA_USE_LOCAL_DEPS=true swift build --package-path Examples/Soto
```
fabianfett, line 43: Let's make this a test matrix :)

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG swift_version=5.7
+ARG swift_version=6.0
```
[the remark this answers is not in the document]
sebsto: IMHO we need an easy way to test SwiftFormat locally before pushing changes.
That's the main reason to keep this, unless you know an easier way ?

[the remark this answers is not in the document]
fabianfett: We now use `swift-format` which is part of the toolchain. Back in the day we used `SwiftFormat`. Not today anymore.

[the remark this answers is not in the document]
sebsto: I know, but we can not just call `swift format -i` to test our. format, we have to call `curl -s https://raw.githubusercontent.com/apple/swift-nio/main/scripts/check-swift-format.sh | bash` which is less convenient, unless you know a better method that is more closely aligned to the CI

docker/docker-compose.yaml
sebsto: I added a loop to automate the discovery of Examples

sebsto:
@swift-server-bot test this

.github/workflows/pull_request.yml
fabianfett: What's the benefit of the matrix here? I still don't think we should use it.

scripts/integration_tests.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set +x -e
+
+for EXAMPLE in $(find Examples -type d -d 1);
+do
+	echo "Building $EXAMPLE"
+	pushd $EXAMPLE
+	LAMBDA_USE_LOCAL_DEPS=../.. swift build
+	popd
+done
```
fabianfett, line 24: what do you think about using a matrix to build them. this should make CI much faster. I don't wat to wait for the AWSSDK and Soto to succeed.

.github/workflows/pull_request.yml
sebsto: @fabianfett because that's the only script I found in Swift-NIO.  Not using it would mean write a specific action for this project.  Is this OK for you ?

scripts/integration_tests.sh
sebsto: are you talking about a CI Action (do you have an example?) or a shell script to run these in parallel ?

sebsto: Is this what you're referring too ?
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

sebsto:
@fabianfett Here is a first version of a Matrix that launch integration tests in parallel. 
Each example is built on swift-6.0-nigthly and swift-main-nightly 

I'm not super happy with the place where the list of example is defined (currently in `.github/workflows/examples_matrix.yml`) because I don't think we can pass an array as parameter between workflow. 
(Ideally, the list of examples should be defined in `.github/workflows/pull_request.yml`)

.github/workflows/examples_matrix.yml
```diff
@@ -0,0 +1,90 @@
...
+        description: "The command of the nightly main Swift version linux matrix job to execute."
+
+## We are cancelling previously triggered workflow runs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.name }}
+  cancel-in-progress: true
+ 
+jobs:
+  linux:
+    name: Linux (${{ matrix.swift.swift_version }})
```
fabianfett, line 51: I think the example name should also be part of the name.

.github/workflows/examples_matrix.yml
```diff
@@ -0,0 +1,90 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # This should be passed as an argument in input. Can we pass arrays as argument ?
+        examples : [ "HelloWorld", "APIGateway" ]
+        # examples: ${{ inputs.examples }}
+        
+        # We are specifying only the major and minor of the docker images to automatically pick up the latest patch release
+        swift:
```
fabianfett, line 61: I honestly think that testing with a single swift version (likely the latest released should be sufficient)

.github/workflows/examples_matrix.yml
```diff
@@ -0,0 +1,90 @@
+name: ExamplesMatrix
+
+on:
+  workflow_call:
+    inputs:
```
fabianfett, line 5: I'm not sure we need an extra workflow for this.

sebsto: @fabianfett This is a matter of separation of concerns. Just like when we create 2 .swift files for two different struct.
In the reporting web view, it doesn't matter if this is a sub workflow or not.

I really don't want to bring this complexity in `pull_request.yml`.  Where would you host this script if not in a separate workflow ?

.github/workflows/examples_matrix.yml
sebsto: done

.github/workflows/examples_matrix.yml
sebsto: I think it's good to test on the 6.x branch and the main branch.
But I will disagree and commit on this.

.github/workflows/examples_matrix.yml
sebsto: image

sebsto:
All green !  Except Swift 6 language mode

- Examples are built on only one version of Swift Linux (the latest one) 
- I kept a separate file to define the matrix job

image

fabianfett (approved):
Love it!

PR #398, opened 2024-10-08
do not force unwrap stdout

sebsto:
Revert a change from #d2bd7f1aae1fc9455b208567ba5d85e42463c82c that caused the plugin to crash

```
building "test" in docker
updating "swift:amazonlinux2" docker image
  amazonlinux2: Pulling from library/swift
  Digest: sha256:089bb14fe09b24ed6ab1fb43379200b42ba9ce52a37889d3dc37cdd1e2d3fbb6
Status: Image is up to date for swift:amazonlinux2
  docker.io/library/swift:amazonlinux2
AWSLambdaPackager/PluginUtils.swift:35: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```


Plugins/AWSLambdaPackager/PluginUtils.swift
```diff
@@ -32,7 +32,7 @@ struct Utils {
 
         let fd = dup(1)
         let stdout = fdopen(fd, "rw")
-        defer { fclose(stdout!) }
```
fabianfett, line 35: we should change ci to use the released swift image here. I think the build issue happens with swift:6.0.0 and swift:6.0.1 but not 6.0-nightly.

sebsto: As much as I understand how it works in Swift-NIO, we can not give the container name as a parameter 
https://github.com/apple/swift-nio/blob/main/.github/workflows/swift_6_language_mode.yml

That would be a change request for the Swift-NIO project

sebsto: The problem with `!` is not in the CI, it crashes the plugin when using it.  We can not force unwrap, there are cases where stdout is nil

fabianfett (approved):

Issue #402, opened 2024-10-09
Create a CI job that tests that we don't like full Foundation


fabianfett:
### Expected behavior

We should have a CI job that validates that we don't link full Foundation and only link FoundationEssentials.

### Actual behavior

N/A

### Steps to reproduce

N/A

### What version of this project (`swift-aws-lambda-runtime`) are you using?

N/A

### Swift version

N/A

sebsto:
Work in progress on the event library 
https://github.com/swift-server/swift-aws-lambda-events/pull/81

sebsto:
Fixed by https://github.com/swift-server/swift-aws-lambda-runtime/pull/441

PR #403, opened 2024-10-10
import foundationEssentials for the archiver plugin

sebsto:
working towards the goal of removing import of Foundation
https://github.com/swift-server/swift-aws-lambda-runtime/issues/402


fabianfett (commented):
Foundation will always be present at build time. We don't want to depend on it at runtime.

PR #406, opened 2024-10-11
[CI] add integration test for the archive plugin

sebsto:
Add a job to test the invocation and the output of the archive plugin


sebsto:
blocked by https://github.com/swift-server/swift-aws-lambda-runtime/issues/409

FranzBusch (approved):

Issue #376, opened 2024-09-27
background example and docc


fabianfett:
### Expected behavior

N/A

### Actual behavior

N/A

### Steps to reproduce

N/A

### What version of this project (`swift-aws-lambda-runtime`) are you using?

N/A

### Swift version

N/A

sebsto:
partially addressed by https://github.com/swift-server/swift-aws-lambda-runtime/pull/418

sebsto:
Missing `docc`. README is OK

PR #418, opened 2024-10-26
Add an example and README for background tasks

sebsto:
Add code example for Lambda Background Tasls 

### Motivation:

Partially addresses https://github.com/swift-server/swift-aws-lambda-runtime/issues/376

### Modifications:

Add a directory `Examples/BackgroundTasks` with code sample and a README

I also slightly modified `Lambda+Codable.swift` to provide default `JSONEncoder()` and `JSONDecoder()` to not require user to create their own if they don't need to.

### Result:

Developers know how to use Lambda background tasks


sebsto:
This code requires to merge https://github.com/swift-server/swift-aws-lambda-runtime/pull/417 to work 
It will compile but no logger debug statement are printed without #417

0xTim (approved):

Issue #369, opened 2024-09-27
Reimplement MockServer for v2


fabianfett:
### Expected behavior

If the env variable `AWS_LAMBDA_RUNTIME_API` is not set, we want to spin up the local mock server directly.

### Actual behavior

Doesn't work

### Steps to reproduce

N/A

### What version of this project (`swift-aws-lambda-runtime`) are you using?

2.0

### Swift version

6.0

PR #425, opened 2024-11-13
re-enable local server for testing

sebsto:
The local server code is the one from v1. 
It is not passing Swift 6 CI test but we decided to let the code as is and re-enable the local testing feature first to optimize the developer experience.

We will fix `LambdaRuntime+LocalServer.swift` for Swift 6 in a second phase

This PR fixes https://github.com/swift-server/swift-aws-lambda-runtime/issues/369


sebsto:
Closing to clean up the commit history

PR #431, opened 2024-12-19
[examples] Add CDK deployment example

sebsto:
added an example showing how to deploy with the CDK
This code is the basis for the upcoming CDK section in the deployment guide


Examples/CDK/Package.swift
```diff
@@ -0,0 +1,62 @@
...
+            dependencies: [
+                .product(name: "AWSLambdaRuntime", package: "swift-aws-lambda-runtime"),
+                .product(name: "AWSLambdaEvents", package: "swift-aws-lambda-events"),
+            ],
+            path: "."
+        )
+    ]
+)
+
+if let localDepsPath = Context.environment["LAMBDA_USE_LOCAL_DEPS"],
```
0xTim, line 37: Do we need all this for the example?

0xTim (approved):
One query around the manifest

Examples/CDK/Package.swift
0xTim: Ok sounds good

PR #433, opened 2024-12-19
Adds default devcontainer configuration

t089:
Adds default devcontainer configuration from SSWG: https://github.com/swift-server/swift-devcontainer-template

### Motivation:

Developing for linux can be challenging on macOS because some APIs are not available on linux. By using VSCode and devcontainer you can essentially develop "on linux".

### Modifications:

Adds devcontainer config.

### Result:

When working in VSCode, you can now easily open the project inside a devcontainer.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,39 @@
+{
+    "name": "Swift",
+    "image": "swift:6.0",
+    "features": {
+        "ghcr.io/devcontainers/features/common-utils:2": {
+            "installZsh": "false",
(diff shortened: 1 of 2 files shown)


sebsto:
As per your comment https://github.com/swift-server/swift-aws-lambda-events/pull/82#issuecomment-2555083759, should we propose a config but leave the directory in the `.gitignore` file ? This would allow users to customize the file without having git interfere.

t089:
That could be an option, indeed.

sebsto (approved):

PR #436, opened 2024-12-20
Explicitly only use `FoundationEssentials` where possible

t089:
We want that the runtime only depends on `FoundationEssentials` where available (ie. on linux) to ensure small binary size. 

### Motivation:

Smaller binary size is good for lambda deployment and cold-start times. The runtime should only depend on `FoundationEssentials`.

### Modifications:

- replace `import Foundation` with `import FoundationEssentials` if `FoundationEssentials` is available.
- I also applied the same treatment to tests to ensure that catch error where tests run on linux and we use API that is only available in `Foundation` which easily happens when you develop on macOS (where always full `Foundation` is available).

### Result:

This should allow builds without linking full `Foundation`.

### DIFF
--- a/Examples/BackgroundTasks/Sources/main.swift
+++ b/Examples/BackgroundTasks/Sources/main.swift
@@ -13,7 +13,11 @@
 //===----------------------------------------------------------------------===//
 
 import AWSLambdaRuntime
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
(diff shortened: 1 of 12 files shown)


sebsto:
Thank you @t089 
Can you run the Swift format script to fix the formatting issue ?

sebsto (approved):
LGTM

PR #438, opened 2024-12-21
[examples] add an example project to show test strategies

sebsto:
This new example project show four testing strategies for Swift Lambda function 

- Unit testing the business logic (not specific to Swift Lambda) 
- Integration testing the handler method 
- Local invocation with the Swift Lambda Runtime 
- Local invocation with SAM 

**[IMPORTANT]**
To allow testing the handler, I had to change visibility of a method in the Runtime project.  This method is clearly marked for testing only, so it should not be a problem. Happy to read feedback and discuss however.


Sources/AWSLambdaRuntimeCore/LambdaContext.swift
```diff
@@ -120,8 +120,10 @@ public struct LambdaContext: CustomDebugStringConvertible, Sendable {
         "\(Self.self)(requestID: \(self.requestID), traceID: \(self.traceID), invokedFunctionARN: \(self.invokedFunctionARN), cognitoIdentity: \(self.cognitoIdentity ?? "nil"), clientContext: \(self.clientContext ?? "nil"), deadline: \(self.deadline))"
     }
 
-    /// This interface is not part of the public API and must not be used by adopters. This API is not part of semver versioning.
-    package static func __forTestsOnly(
+    /// This interface is not part of the public API and must not be used by adopters.
+    /// It is declared public to allow users to unit test their handler.
+    /// This API is not part of semver versioning.
+    public static func __forTestsOnly(
```
0xTim, line 126: Why are we making this public if it can be visible with `@testable`?

0xTim: If we want it public, we might want to consider introducing a `LambdaTesting` target to keep it separate

0xTim (commented):
Some queries around testing infrastructure

Sources/AWSLambdaRuntimeCore/LambdaContext.swift
sebsto: @0xTim 

You're correct, I was importing `@testable import AWSLambdaRuntime` but I needed `AWSLambdaRuntimeCore`
Thank you !

sebsto:
Note that the integration tests fail pending merging of https://github.com/swift-server/swift-aws-lambda-runtime/pull/441

0xTim (approved):

PR #441, opened 2024-12-22
[CI] add test to check libFoundation.so & co are not linked to the bi…

sebsto:
This adds a test in the `integration_test.yml` workflow to verify that `libFoundation.so`,  `libFoundationInternationalization.so`, and `lib_FoundationICU.so` and not linked the the function binary.

This is a follow up from the work make by @t089 to ensure that we don't accidentally link Foundation again in the future.

Remove Foundation : https://github.com/swift-server/swift-aws-lambda-runtime/pull/436
Add test in CI to ensure we don't link to foundation : https://github.com/swift-server/swift-aws-lambda-runtime/issues/402


t089 (approved):
LGTM! Thank you.

.github/workflows/scripts/check-link-foundation.sh
0xTim, line 1: GH is being really weird at the moment and diffs are broken - it's not showing this as a new file, like it's stuck diffing a particular commit or something. This file doesn't exist on main right?

sebsto: @0xTim  I’m sorry I messed things up. I added the file in a separate commit because I initially used the curl-based pattern of the GitHub-workflow project. It required the script to be available before the PR. 

My fault. 

You can see the full script here 
https://github.com/swift-server/swift-aws-lambda-runtime/pull/442/files

0xTim: Ahh makes sense, thanks!

0xTim (approved):

PR #459, opened 2025-01-12
[core] Add cancellation handling for nextInvocation()

adam-fowler:
Allow `LambdaChannelHandler.nextInvocation` to be cancelled.

### Motivation:

If we want to use ServiceLifecycle with the lambda runtime the lambda runtime needs to be cancellable either via a ServiceLifecycle graceful shutdown or via Task cancellation. To avoid bringing in the ServiceLifecycle dependency this PR adds cancellation via Task cancellation handler.

### Modifications:

Add `withTaskCancellationHandler` to nextInvocation which calls close on cancel. 
In `LambdaChannelHandler.channelInactive` resume continuation if state is `waitingForNextInvocation`
Added `LambdaRuntimeClientTests.testCancellation`

### Result:

You can now cancel the runtime while it is waiting for the next invocation.

### DIFF
--- a/Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTests.swift
+++ b/Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTests.swift
@@ -86,4 +86,24 @@ struct LambdaRuntimeClientTests {
             }
         }
     }
+
+    @Test
+    func testCancellation() async throws {
+        try await LambdaRuntimeClient.withRuntimeClient(
+            configuration: .init(ip: "127.0.0.1", port: 7000),
+            eventLoop: NIOSingletons.posixEventLoopGroup.next(),
+            logger: self.logger
+        ) { runtimeClient in
+            try await withThrowingTaskGroup(of: Void.self) { group in
+                group.addTask {
+                    while true {
+                        _ = try await runtimeClient.nextInvocation()
+                    }
+                }
+                // wait a small amount to ensure we are waiting for continuation
+                try await Task.sleep(for: .milliseconds(100))
+                group.cancelAll()
+            }
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


adam-fowler:
This PR does not compile in swift 6 mode and I don't see an easy way around the issues.

fabianfett (commented):
My goodness. two comments have lingered here forever in pending. Just flushing those right now

adam-fowler:
Re-wrote this. It is considerably simpler. Also fixed test to ensure it does check for cancellation

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTests.swift
```diff
@@ -86,4 +86,56 @@ struct LambdaRuntimeClientTests {
...
+                            // Wrap this is a task so cancellation isn't propagated to the write calls
+                            try await Task {
+                                try await writer.write(ByteBuffer(string: "hello"))
+                                try await writer.finish()
+                            }.value
+                        }
+                    }
+                    // wait a small amount to ensure we are waiting for continuation
+                    try await Task.sleep(for: .milliseconds(100))
+                    group.cancelAll()
```
fabianfett, line 136: we should check that the spawned task returns with a cancellation error indeed!

fabianfett (commented):
This looks really good! Mind adding the test check.

Tests/AWSLambdaRuntimeCoreTests/LambdaRuntimeClientTests.swift
adam-fowler: It doesn't return cancellation. Currently returns isInClosedChannel. Not sure how to get it to return the cancellation error as I'm just closing the connection and waiting for it to clean up. I could catch the error and then check Task.isCancelled and throw a Cancellation error if it is cancelled but that seems a bit of a hack.

sebsto:
> This looks really good! Mind adding the test check.

@adam-fowler ?

adam-fowler:
Currently the test check will fail. There isn't an easy way to indicate the connection was closed due to cancellation without some re-structuring.

sebsto:
Looks like we're good to merge then. Thank you for having confirmed

fabianfett (approved):

Issue #377, opened 2024-09-27
Performance tests for the invocation loop


fabianfett:
### Expected behavior

N/A

### Actual behavior

N/A

### Steps to reproduce

N/A

### What version of this project (`swift-aws-lambda-runtime`) are you using?

N/A

### Swift version

N/A

sebsto:
Implement a test for https://maxday.github.io/lambda-perf/
and contact the owner to add the runtime 2 to the list of bench

sebsto:
local performance with swift-benchmark package

sebsto:
Fixed by https://github.com/swift-server/swift-aws-lambda-runtime/pull/542

PR #463, opened 2025-01-13
[test] Update mock server for Swift 6 compliance

sebsto:
Rewrite to MockServer (used only for performance testing and part of a separate target) to comply with Swift 6 language mode and concurrency.

The new MockServer now uses `NIOAsyncChannel` and structured concurrency.

Instead of adding support to [MAX_REQUEST](https://github.com/swift-server/swift-aws-lambda-runtime/blob/11756b4e00ca75894826b41666bdae506b6eb496/Sources/AWSLambdaRuntimeCore/LambdaConfiguration.swift#L53) environment variable like v1 did, we implemented support for `MAX_REQUEST` environment variable in the MockServer itself.  It closes the connection and shutdown the server after servicing MAX_INVOCATIONS Lambda requests). This allow to add the MAX_REQUEST penalty on the MockServer and not on the LambdaRuntimeClient.

However, currently, the LambdaRuntimeClient does not shutdown when the MockServer ends.  I created https://github.com/swift-server/swift-aws-lambda-runtime/issues/465 to track this issue.

See https://github.com/swift-server/swift-aws-lambda-runtime/issues/377


sebsto:
Mock server is now compliant to swift 6 concurrency (thank you @FranzBusch ) 

This PR is now blocked by https://github.com/swift-server/swift-aws-lambda-runtime/issues/465

FranzBusch (approved):

Issue #449, opened 2025-01-02
Building on linux (Ubuntu)... getting permission problems (due to Docker running as root?)


mflint:
### Expected behavior

I'd like to be able to build the lambdas for Arm using a Linux host, because a buying a chonky Raspberry Pi is much cheaper than a new MacBook with Apple Silicon:

```
swift package archive --allow-network-connections docker
```

### Actual behavior

The _archiving_ stage fails with a permission error:

```
  {lots of build output deleted}

  [495/497] Compiling SuggestionGetFunction SuggestionGet.swift
  [495/497] Write Objects.LinkFileList
  /usr/lib/swift_static/linux/libFoundationEssentials.a(FoundationEssentials-1.o):FoundationEssentials-1.o:function $s20FoundationEssentials19createTemporaryFile33_FC9EC52B075D2ACCFF86F1C9F84293BELL2at6inPath6prefix7optionss5Int32V_SStSS_AA0Q5OrURLOSSAA4DataV14WritingOptionsVtKFTf4nnnd_n: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
/usr/lib/swift_static/linux/libFoundationEssentials.a(FoundationEssentials-1.o):FoundationEssentials-1.o:function $s20FoundationEssentials19createTemporaryFile33_FC9EC52B075D2ACCFF86F1C9F84293BELL2at6inPath6prefix7optionss5Int32V_SStSS_AA0Q5OrURLOSSAA4DataV14WritingOptionsVtKFTf4nnnd_n: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
/usr/lib/swift_static/linux/libFoundationEssentials.a(FoundationEssentials-1.o):FoundationEssentials-1.o:function $s20FoundationEssentials19createTemporaryFile33_FC9EC52B075D2ACCFF86F1C9F84293BELL2at6inPath6prefix7optionss5Int32V_SStSS_AA0Q5OrURLOSSAA4DataV14WritingOptionsVtKFTf4nnnd_n: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
[...]

PR #467, opened 2025-01-18
[plugin] support for resources packaging on ubuntu

sebsto:
When including resources in the package and packaging on Ubuntu, `FileManager` throws a FilePermission error. The docker daemon runs as root and files to be copied are owned by `root` while the archiver runs as the current user (`ubuntu`  on EC2 Ubuntu). The `FileManager` manages to copy the files but throws an error after the copy. We suspect the `FileManager` to perform some kind of operation after the copy and it fails because of the `root` permission of the files. 

See https://github.com/swift-server/swift-aws-lambda-runtime/issues/449#issuecomment-2595978246 for a description of the problem.

This PR contains code to reproduce the problem, a very simple workaround, and an integration test.
The workaround consists of 
- trapping all errors
- verify if the error is the permission error (Code = 513)
- verify if the files have been copied or not 
- if the two above conditions are met, ignore the error, otherwise re-throw it 

I would rather prefer a solution that solves the root cause rather than just ignoring the error.
We're still investigating the root cause (see [this thread](https://forums.swift.org/t/filemanager-copyitem-on-linux-fails-after-copying-the-files/77282) on the Swift Forum and this issue on Swift Foundation https://github.com/swiftlang/swift-foundation/issues/1125

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -19,6 +19,11 @@ on:
         type: boolean
         description: "Boolean to enable the compilation of examples. Defaults to true."
         default: true
+      archive_plugin_examples:
+        type: string
+        description: "The list of examples to run through the archive plugin test. Pass a String with a valid JSON array such as \"[ 'HelloWorld', 'APIGateway' ]\""
+        required: true
+        default: ""
       archive_plugin_enabled:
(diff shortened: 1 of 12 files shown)


0xTim (approved):
LGTM

PR #477, opened 2025-02-11
Add `S3EventNotifier` example

ptoffy:
Add `S3EventNotifier` example

### Motivation:

There is currently no example regarding an AWS event triggered lambda, such as a lambda that gets invoked on an S3 object upload. I've had to look for a while to figure out that the `AWSLambdaEvents` exists and that it can be used for that (had to find out via a Java example! 😜) so I think this could be useful and a somewhat common use case

### Modifications:

Added an example of a lambda that gets triggered when an object gets uploaded to an S3 bucket.

I have not described how to set up the actual S3 event in AWS because I figured if you needed such a lambda you'd know, but I can describe that too in the README if needed

### DIFF
--- a/Examples/S3EventNotifier/Package.swift
+++ b/Examples/S3EventNotifier/Package.swift
@@ -1,0 +1,52 @@
+// swift-tools-version: 6.0
+import PackageDescription
+
+// needed for CI to test the local version of the library
+import struct Foundation.URL
+
+let package = Package(
+    name: "CSVUploadAPINotificationLambda",
+    platforms: [.macOS(.v15)],
+    dependencies: [
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", branch: "main"),
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-events", branch: "main"),
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.24.0"),
+    ],
+    targets: [
+        .executableTarget(
+            name: "CSVUploadAPINotificationLambda",
+            dependencies: [
+                .product(name: "AWSLambdaRuntime", package: "swift-aws-lambda-runtime"),
+                .product(name: "AWSLambdaEvents", package: "swift-aws-lambda-events"),
+                .product(name: "AsyncHTTPClient", package: "async-http-client"),
+            ]
+        )
+    ]
+)
+
+if let localDepsPath = Context.environment["LAMBDA_USE_LOCAL_DEPS"],
+    localDepsPath != "",
+    let v = try? URL(fileURLWithPath: localDepsPath).resourceValues(forKeys: [.isDirectoryKey]),
+    v.isDirectory == true
+{
+    // when we use the local runtime as deps, let's remove the dependency added above
+    let indexToRemove = package.dependencies.firstIndex { dependency in
+        if case .sourceControl(
+            name: _,
+            location: "https://github.com/swift-server/swift-aws-lambda-runtime.git",
+            requirement: _
+        ) = dependency.kind {
+            return true
+        }
+        return false
+    }
+    if let indexToRemove {
+        package.dependencies.remove(at: indexToRemove)
+    }
+
+    // then we add the dependency on LAMBDA_USE_LOCAL_DEPS' path (typically ../..)
+    print("[INFO] Compiling against swift-aws-lambda-runtime located at \(localDepsPath)")
+    package.dependencies += [
+        .package(name: "swift-aws-lambda-runtime", path: localDepsPath)
+    ]
+}
--- a/Examples/S3EventNotifier/Package.swift
+++ b/Examples/S3EventNotifier/Package.swift
@@ -5,20 +5,18 @@ import PackageDescription
 import struct Foundation.URL
 
 let package = Package(
-    name: "CSVUploadAPINotificationLambda",
+    name: "S3EventNotifier",
     platforms: [.macOS(.v15)],
     dependencies: [
         .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", branch: "main"),
         .package(url: "https://github.com/swift-server/swift-aws-lambda-events", branch: "main"),
-        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.24.0"),
     ],
     targets: [
         .executableTarget(
-            name: "CSVUploadAPINotificationLambda",
+            name: "S3EventNotifier",
             dependencies: [
                 .product(name: "AWSLambdaRuntime", package: "swift-aws-lambda-runtime"),
                 .product(name: "AWSLambdaEvents", package: "swift-aws-lambda-events"),
-                .product(name: "AsyncHTTPClient", package: "async-http-client"),
             ]
         )
     ]
--- a/Examples/S3EventNotifier/README.md
+++ b/Examples/S3EventNotifier/README.md
@@ -4,7 +4,7 @@ This example demonstrates how to write a Lambda that is invoked by an event orig
 
 ## Code
 
-In this example the lambda function receives an `S3Event` object defined in the `AWSLambdaEvents` library as input object. The `S3Event` object contains all the information about the S3 event that triggered the function, but what we are interested in is the bucket name and the object key, which are inside of a notification `Record`. The object contains an array of records, however since the lambda is triggered by a single event, we can safely assume that there is only one record in the array: the first one. Inside of this record, we can find the bucket name and the object key:
+In this example the Lambda function receives an `S3Event` object defined in the `AWSLambdaEvents` library as input object. The `S3Event` object contains all the information about the S3 event that triggered the function, but what we are interested in is the bucket name and the object key, which are inside of a notification `Record`. The object contains an array of records, however since the Lambda function is triggered by a single event, we can safely assume that there is only one record in the array: the first one. Inside of this record, we can find the bucket name and the object key:
 
 ```swift
 guard let s3NotificationRecord = event.records.first else {
@@ -46,13 +46,13 @@ The `--architectures` flag is only required when you build the binary on an Appl
 
 Be sure to replace `<YOUR_ACCOUNT_ID>` with your actual AWS account ID (for example: 012345678901).
 
-Besides deploying the lambda function you also need to create the S3 bucket and configure it to send events to the lambda function. You can do this using the following commands:
+Besides deploying the Lambda function you also need to create the S3 bucket and configure it to send events to the Lambda function. You can do this using the following commands:
 
 ```bash
 aws s3api create-bucket --bucket my-test-bucket --region eu-west-1 --create-bucket-configuration LocationConstraint=eu-west-1
 
 aws lambda add-permission \
-    --function-name ProcessS3Upload \
+    --function-name S3EventNotifier \
     --statement-id S3InvokeFunction \
     --action lambda:InvokeFunction \
     --principal s3.amazonaws.com \
@@ -62,7 +62,7 @@ aws s3api put-bucket-notification-configuration \
     --bucket my-test-bucket \
     --notification-configuration '{
         "LambdaFunctionConfigurations": [{
-            "LambdaFunctionArn": "arn:aws:lambda:<REGION>:<YOUR_ACCOUNT_ID>:function:ProcessS3Upload",
+            "LambdaFunctionArn": "arn:aws:lambda:<REGION>:<YOUR_ACCOUNT_ID>:function:S3EventNotifier",
             "Events": ["s3:ObjectCreated:*"]
         }]
     }'
@@ -72,8 +72,8 @@ aws s3 cp testfile.txt s3://my-test-bucket/
 
 This will:
  - create a bucket named `my-test-bucket` in the `eu-west-1` region;
- - add a permission to the lambda function to be invoked by the bucket;
- - configure the bucket to send `s3:ObjectCreated:*` events to the lambda function named `ProcessS3Upload`;
+ - add a permission to the Lambda function to be invoked by Amazon S3;
+ - configure the bucket to send `s3:ObjectCreated:*` events to the Lambda function named `S3EventNotifier`;
  - upload a file named `testfile.txt` to the bucket.
 
-Replace `<REGION>` with the region where you deployed the lambda function and `<YOUR_ACCOUNT_ID>` with your actual AWS account ID.
+Replace `<REGION>` with the region where you deployed the Lambda function and `<YOUR_ACCOUNT_ID>` with your actual AWS account ID.
--- a/Examples/S3EventNotifier/README.md
+++ b/Examples/S3EventNotifier/README.md
@@ -30,8 +30,8 @@ If there are no errors, a ZIP file should be ready to deploy, located at `.build
 
 ## Deploy
 
-[!IMPORTANT]
-The Lambda function and the S3 bucket must be located in the same AWS Region. In the code below, we use `eu-west-1` (Ireland). 
+> [!IMPORTANT]
+> The Lambda function and the S3 bucket must be located in the same AWS Region. In the code below, we use `eu-west-1` (Ireland). 
 
 To deploy the Lambda function, you can use the `aws` command line:
 
@@ -90,5 +90,5 @@ This will:
 
 Replace `my-test-bucket` with your bucket name (bucket names are unique globaly and this one is already taken). Also replace `REGION` environment variable with the AWS Region where you deployed the Lambda function and `<YOUR_ACCOUNT_ID>` with your actual AWS account ID.
 
-[!IMPORTANT]
-The Lambda function and the S3 bucket must be located in the same AWS Region. Adjust the code above according to your closest AWS Region.
+> [!IMPORTANT]
+> The Lambda function and the S3 bucket must be located in the same AWS Region. Adjust the code above according to your closest AWS Region.
--- a/Examples/S3EventNotifier/README.md
+++ b/Examples/S3EventNotifier/README.md
@@ -1,0 +1,52 @@
+# S3 Event Notifier
+
+This example demonstrates how to create a Lambda that notifies an API of an S3 event in a bucket.
+
+## Code
+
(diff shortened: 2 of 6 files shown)


sebsto:
Thank you for your contribution. This is much appreciated.  I will add some comments as a code review.
Can you also 

- add this example in the CICD tool : https://github.com/swift-server/swift-aws-lambda-runtime/blob/bd0ec626464f1c1232e7baf2cabf4138e1f770a7/.github/workflows/pull_request.yml#L39
- add this example in the general example readme file : https://github.com/swift-server/swift-aws-lambda-runtime/blob/main/Examples/README.md

Examples/S3EventNotifier/Package.swift
```diff
@@ -0,0 +1,52 @@
+// swift-tools-version: 6.0
+import PackageDescription
+
+// needed for CI to test the local version of the library
+import struct Foundation.URL
+
+let package = Package(
+    name: "CSVUploadAPINotificationLambda",
```
sebsto, line 8: I would use the same name as the directory : S3EventNotifier

Examples/S3EventNotifier/Package.swift
```diff
@@ -0,0 +1,52 @@
...
+    platforms: [.macOS(.v15)],
+    dependencies: [
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", branch: "main"),
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-events", branch: "main"),
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.24.0"),
+    ],
+    targets: [
+        .executableTarget(
+            name: "CSVUploadAPINotificationLambda",
```
sebsto, line 17: same

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,52 @@
+# S3 Event Notifier
+
+This example demonstrates how to create a Lambda that notifies an API of an S3 event in a bucket.
```
sebsto, line 3: This example demonstrates how to write a Lambda function that is invoked by an event originating from Amazon S3, such as a file has been created in an S3 bucket.

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Code
+
+In this example the lambda function receives an `S3Event` object from the `AWSLambdaEvents` library as input object instead of a `APIGatewayV2Request`. The `S3Event` object contains all the information about the S3 event that triggered the lambda, but what we are interested in is the bucket name and the object key, which are inside of a notification `Record`. The object contains an array of records, however since the lambda is triggered by a single event, we can safely assume that there is only one record in the array: the first one. Inside of this record, we can find the bucket name and the object key:
```
sebsto, line 7: The S3Event is defined in the Lambda event library but the event is sent by S3.
I would not refer to APIGatewayRequest - this is not bringing new value in the context of this example.

"the lambda" -> "the Lambda function" or just "the function" (2x)

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,52 @@
...
+    throw LambdaError.noNotificationRecord
+}
+
+let bucket = s3NotificationRecord.s3.bucket.name
+let key = s3NotificationRecord.s3.object.key.replacingOccurrences(of: "+", with: " ")
+```
+
+The key is URL encoded, so we replace the `+` with a space.
+
+Once the event is decoded, the lambda sends a POST request to an API endpoint with the bucket name and the object key as parameters. The API URL is set as an environment variable.
```
sebsto, line 20: Is this necessary for this example? It adds complexity and the user would beed an API endpoint to validate the example.

Can we replace this with just a `context.log.info` to print the name of the bucket and key?

Examples/S3EventNotifier/Package.swift
```diff
@@ -0,0 +1,52 @@
...
+// needed for CI to test the local version of the library
+import struct Foundation.URL
+
+let package = Package(
+    name: "CSVUploadAPINotificationLambda",
+    platforms: [.macOS(.v15)],
+    dependencies: [
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", branch: "main"),
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-events", branch: "main"),
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.24.0"),
```
sebsto, line 13: Maybe can be removed if we remove the POST API as discussed below

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,52 @@
...
+--architectures arm64 \
+--role arn:aws:iam::<YOUR_ACCOUNT_ID>:role/lambda_basic_execution
+```
+
+The `--architectures` flag is only required when you build the binary on an Apple Silicon machine (Apple M1 or more recent). It defaults to `x64`.
+
+Be sure to replace <YOUR_ACCOUNT_ID> with your actual AWS account ID (for example: 012345678901).
+
+> [!WARNING]
+> You will have to set up an S3 bucket and configure it to send events to the lambda function. This is not covered in this example.
```
sebsto, line 52: I would add minimum command line instructions to do so 

1. create a bucket 
2. configure the event 
3. upload a file to show it works

Something like this (not tested) 

```
aws s3api create-bucket --bucket my-test-bucket --region eu-west-1 --create-bucket-configuration LocationConstraint=eu-west-1

aws lambda add-permission \
    --function-name ProcessS3Upload \
    --statement-id S3InvokeFunction \
    --action lambda:InvokeFunction \
    --principal s3.amazonaws.com \
    --source-arn arn:aws:s3:::my-test-bucket

aws s3api put-bucket-notification-configuration \
    --bucket my-test-bucket \
    --notification-configuration '{
        "LambdaFunctionConfigurations": [{
            "LambdaFunctionArn": "arn:aws:lambda:<REGION>:<YOUR-ACCOUNT-ID>:function:ProcessS3Upload",
            "Events": ["s3:ObjectCreated:*"]
        }]
    }'

aws s3 cp testfile.txt s3://my-test-bucket/
```

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
```
sebsto, line 5: 2025

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        let body = """
+            {
+                "bucket": "\(bucket)",
+                "key": "\(key)"
+            }
+            """
+
+        context.logger.debug("Sending request to \(apiURL) with body \(body)")
```
sebsto, line 63: I would simplify and not send a POST request to an API.
That obliges developers willing to test this function to setup an HTTPS API endpoint, which is unnecessary to understand this example.

I propose to just log relevant informations from the event.  The specific handling of these events are application specific, we can't assume a dev will have to call a REST API. I think it just adds complexity (and maybe confusion) to the example.

You can add comments to suggest what can be done after receiving the event

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+        switch self {
+        case .noNotificationRecord:
+            "No notification record in S3 event"
+        case .missingEnvVar(let name):
+            "Missing env var named \(name)"
+        }
+    }
+}
+
+let runtime = LambdaRuntime { (event: S3Event, context: LambdaContext) async throws -> APIGatewayV2Response in
```
sebsto, line 41: S3 Notifications are fire and forget (asynchronous Lambda invocations). We don't need to return a value.  `APIGatewayV2Response` should be `Void`

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        var request = HTTPClientRequest(url: "\(apiURL)/upload-complete/")
+        request.method = .POST
+        request.headers = [
+            "Content-Type": "application/json"
+        ]
+        request.body = .bytes(.init(string: body))
+
+        let response = try await httpClient.execute(request, timeout: .seconds(30))
+        return APIGatewayV2Response(
```
sebsto, line 73: no return value necessary

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+            statusCode: .ok,
+            body: "Lambda terminated successfully. API responded with: Status: \(response.status), Body: \(response.body)"
+        )
+    } catch let error as LambdaError {
+        context.logger.error("\(error.description)")
+        return APIGatewayV2Response(statusCode: .internalServerError, body: "[ERROR] \(error.description)")
```
sebsto, line 79: no return value necessary

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+    } catch {
+        context.logger.error("\(error)")
+        return APIGatewayV2Response(statusCode: .internalServerError, body: "[ERROR] \(error)")
```
sebsto, line 82: no return value necessary

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        guard let s3NotificationRecord = event.records.first else {
+            throw LambdaError.noNotificationRecord
+        }
+
+        let bucket = s3NotificationRecord.s3.bucket.name
+        let key = s3NotificationRecord.s3.object.key.replacingOccurrences(of: "+", with: " ")
+
+        guard let apiURL = ProcessInfo.processInfo.environment["API_URL"] else {
+            throw LambdaError.missingEnvVar(name: "API_URL")
```
sebsto, line 53: The error is throw and catch in the same function.
I think it would be easier to read if the core of the `guard` statement would just log the error and return

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+}
+
+let runtime = LambdaRuntime { (event: S3Event, context: LambdaContext) async throws -> APIGatewayV2Response in
+    do {
+        context.logger.debug("Received S3 event: \(event)")
+
+        guard let s3NotificationRecord = event.records.first else {
+            throw LambdaError.noNotificationRecord
```
sebsto, line 46: The error is throw and catch in the same function.
I think it would be easier to read if the core of the guard statement would just log the error and return

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,86 @@
...
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
+let httpClient = HTTPClient.shared
+
+enum LambdaError: Error {
```
sebsto, line 27: If you implement the error handling as suggested below, this whole struct can be removed

sebsto (changes requested):
Thank you for this great contribution !  I think you're on the right path. I did propose a couple of simplifications. The primary objective of the example code is readability and understandability.  The code must be simple to read and understand and distraction-free.

I think there are a bit of improvement possible with this objective in mind.

I'm happy to discuss alternatives.

Examples/S3EventNotifier/Package.swift
ptoffy: Oops, forgot to change this

Examples/S3EventNotifier/README.md
ptoffy: Yep no problem

sebsto:
Thank you for the quick turnaround. I will provide additional comments on the README.
Fo the AWS CLI, I will test them before merging the PR, no worries.

The example does not compile on the CI chain.
See https://github.com/swift-server/swift-aws-lambda-runtime/actions/runs/13264375954

This is because it compiles on Linux and `replacingOccurence(of:with:)` is based on `NSString`. I guess it is not available in Foundations Essentials.

This triggers the question : did you package and deploy this code on Lambda to test it ?

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,79 @@
+# S3 Event Notifier
+
+This example demonstrates how to write a Lambda that is invoked by an event originating from Amazon S3, such as a new object being uploaded to a bucket.
+
+## Code
+
+In this example the lambda function receives an `S3Event` object defined in the `AWSLambdaEvents` library as input object. The `S3Event` object contains all the information about the S3 event that triggered the function, but what we are interested in is the bucket name and the object key, which are inside of a notification `Record`. The object contains an array of records, however since the lambda is triggered by a single event, we can safely assume that there is only one record in the array: the first one. Inside of this record, we can find the bucket name and the object key:
```
sebsto, line 7: the second occurence of 'lambda' must be replaced by 'Lambda function' (capital L and the words "Lambda function" together, not just Lambda)

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,79 @@
...
+    --handler provided  \
+    --architectures arm64 \
+    --role arn:aws:iam::<YOUR_ACCOUNT_ID>:role/lambda_basic_execution
+```
+
+The `--architectures` flag is only required when you build the binary on an Apple Silicon machine (Apple M1 or more recent). It defaults to `x64`.
+
+Be sure to replace `<YOUR_ACCOUNT_ID>` with your actual AWS account ID (for example: 012345678901).
+
+Besides deploying the lambda function you also need to create the S3 bucket and configure it to send events to the lambda function. You can do this using the following commands:
```
sebsto, line 49: Lambda function (capital L)

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,79 @@
...
+            "Events": ["s3:ObjectCreated:*"]
+        }]
+    }'
+
+aws s3 cp testfile.txt s3://my-test-bucket/
+```
+
+This will:
+ - create a bucket named `my-test-bucket` in the `eu-west-1` region;
+ - add a permission to the lambda function to be invoked by the bucket;
```
sebsto, line 75: to be invoked by Amazon S3

Examples/S3EventNotifier/README.md
```diff
@@ -0,0 +1,79 @@
...
+ - configure the bucket to send `s3:ObjectCreated:*` events to the lambda function named `ProcessS3Upload`;
```
sebsto, line 76: U[date function name to match the code

Examples/S3EventNotifier/Sources/main.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import AWSLambdaEvents
+import AWSLambdaRuntime
+
+let runtime = LambdaRuntime { (event: S3Event, context: LambdaContext) async throws in
+    context.logger.debug("Received S3 event: \(event)")
```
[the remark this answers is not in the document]
ptoffy: I had it set up because the original code was doing a bunch of more stuff so diverse logging would help debug but I'm happy to remove it

ptoffy:
Yep I have the lambda (with some additions such as the POST request) running in production just fine. I guess either AHC or some other dependency which I'm using there and we trimmed here use Foundation and it's falling back to that. I'll update this to use Foundation

sebsto:
Thank you for all these changes. It looks good to me now.  I will check the CLI command before merging

Examples/S3EventNotifier/README.md
```diff
@@ -76,4 +82,7 @@ This will:
  - configure the bucket to send `s3:ObjectCreated:*` events to the Lambda function named `S3EventNotifier`;
  - upload a file named `testfile.txt` to the bucket.
 
-Replace `<REGION>` with the region where you deployed the Lambda function and `<YOUR_ACCOUNT_ID>` with your actual AWS account ID.
+Replace `my-test-bucket` with your bucket name (bucket names are unique globaly and this one is already taken). Also replace `<REGION>` with the region where you deployed the Lambda function and `<YOUR_ACCOUNT_ID>` with your actual AWS account ID.
+
+[!IMPORTANT]
+The Lambda function and the S3 bucket must be located in the same AWS Region. Adjust the code above according to your closest AWS Region.
```
ptoffy, line 94: This may be unnecessary if we have the one above already? But I'm not precious about it. If you do end up leaving it, this should use block quotes too

Examples/S3EventNotifier/README.md
```diff
@@ -30,10 +30,14 @@ If there are no errors, a ZIP file should be ready to deploy, located at `.build
 
 ## Deploy
 
+[!IMPORTANT]
+The Lambda function and the S3 bucket must be located in the same AWS Region. In the code below, we use `eu-west-1` (Ireland).
```
ptoffy, line 34: I think GH markdown requires this to be in block quotes, so
```
> [!IMPORTANT]
> The Lambda function...
```

Examples/S3EventNotifier/README.md
sebsto: https://github.com/orgs/community/discussions/16925

Examples/S3EventNotifier/README.md
sebsto: It's a GitHub specific tag https://github.com/orgs/community/discussions/16925

Examples/S3EventNotifier/README.md
sebsto: I made it on purpose - it so important and I'm sure so many will not read it :-)

Examples/S3EventNotifier/README.md
sebsto: You're correct - thank you

PR #479, opened 2025-02-16
Fix data race in LocalServer's invocation pool

sebsto:
Use a single Mutex to store both the Invocation buffer and the continuation to avoid a possible data race condition as described here.

https://forums.swift.org/t/how-to-use-withcheckedcontinuation-with-an-async-closure/77852/8

### DIFF
--- a/Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
+++ b/Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
@@ -349,23 +349,25 @@ private struct LambdaHttpServer {
     private final class Pool<T>: AsyncSequence, AsyncIteratorProtocol, Sendable where T: Sendable {
         typealias Element = T
 
-        private let _buffer = Mutex<CircularBuffer<T>>(.init())
-        private let _continuation = Mutex<CheckedContinuation<T, any Error>?>(nil)
+        private let mutex = Mutex<(CircularBuffer<T>, CheckedContinuation<T, any Error>?)>((.init(), nil))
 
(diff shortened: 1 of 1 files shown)


Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
```diff
@@ -376,15 +378,17 @@ private struct LambdaHttpServer {
...
-                return try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<T, any Error>) in
-                    // store the continuation for later, when an element is enqueued
-                    self._continuation.withLock {
-                        $0 = continuation
+                // so, wait for an element to be enqueued
+                return try await withCheckedThrowingContinuation {
+                    (continuation: CheckedContinuation<T, any Error>) in
+                    self.mutex.withLock { mutexContent in
+                        // store the continuation for later, when an element is enqueued
+                        mutexContent.1 = continuation
```
t089, line 391: This is still racy: between "popFirst" returning nil and storing the continuation, somebody could have already put sth in the buffer. Really, to implement this correctly, I think you need to acquire the lock, check your invariants, apply the necessary state change, release the lock and then run side-effects.

t089 (changes requested):

Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift
t089: The invariant here is: There can only be a stored continuation if the buffer is empty. If the buffer is non-empty, there cannot be a stored continuation.

PR #484, opened 2025-02-24
Instructions for customizing the output path are incorrect

mattcorey:
Instructions for customizing the output path are incorrect

### Motivation:

`displayHelpMessage` was incorrect - it claimed that the output path was customizable via a `--output-directory` flag, but the code actually looks for `--output-path`

### Modifications:

Changing the help message here, instead of the code, in case existing projects already depend on `--output-path`

### Result:

`displayHelpMessage` will show the correct instructions


sebsto (approved):
LGTM

sebsto:
Thank you for having spot this !

PR #487, opened 2025-02-27
Replace usage of `NIOLockedValueBox` with `Mutex`

fabianfett:


sebsto (approved):
LGTM in terms of code, but the many tests fail now. The compiler crashes with a stack trace

Issue #488, opened 2025-02-27
[core] LocalServer does not report errors correctly


sebsto:
## Expected behavior

The LocalLambdaServer used for testing does not report errors to the caller correctly.

- the server should capture the error POSTed by the `LambdaRuntime` and forward it to the client 
- the server should close the TCP connection

## Actual behavior

1. The server does not report the error back to the client
2. The server does not close the connection and the `curl` client stays blocked.

## Steps to reproduce

```
cd Examples/APIGateway
LOG_LEVEL=trace swift run 
```

In another terminal, send an invalid payload:

```
curl --header 'Content-Type: application/json' --data "{}" --verbose http://127.0.0.1:7000/invoke 
```

## If possible, minimal yet complete reproducer code (or URL to code)

APIGateway example

## What version of this project (`swift-aws-lambda-runtime`) are you using?

main

### Swift version

n/a

## Amazon Linux 2 docker image version

n/a

## Diagnostic 

**Diagnostic** 

Server trace:
```
2025-02-27T17:36:12+0100 debug LambdaRuntime : [AWSLambdaRuntimeCore] LambdaRuntime initialized
2025-02-27T17:36:12+0100 info LocalServer : host="127.0.0.1" port=7000 [AWSLambdaRuntimeCore] Server started and listening
2025-02-27T17:36:12+0100 trace LambdaRuntime : lambda_ip=127.0.0.1 lambda_port=7000 [AWSLambdaRuntimeCore] Connection to control plane created
2025-02-27T17:36:12+0100 trace LocalServer : [AWSLambdaRuntimeCore] Handling a new connection
2025-02-27T17:36:12+0100 trace LocalServer : URI=GET /2018-06-01/runtime/invocation/next [AWSLambdaRuntimeCore] Processing request
2025-02-27T17:36:12+0100 trace LocalServer : [AWSLambdaRuntimeCore] /next waiting for /invoke
2025-02-27T17:36:38+0100 trace LocalServer : [AWSLambdaRuntimeCore] Handling a new connection
2025-02-27T17:36:38+0100 trace LocalServer : Body={} URI=POST /invoke [AWSLambdaRuntimeCore] Processing request
2025-02-27T17:36:38+0100 trace LocalServer : requestID=544640702971583 [AWSLambdaRuntimeCore] /invoke received invocation
[...]

PR #489, opened 2025-03-01
[TestServer] Fix 488 and add simple test lambda

fabianfett:
Fixes: #488


sebsto:
Thank you @fabianfett for this patch. I take the responsibility of the original error :-) 

I just tested it with the APIGateway example (as per "Steps to reproduce" in the issue description). I confirm this fixes the error.

Can you delete from this PR the `HelloWorldLambda` and revert back the `Package.swift` ?

On a side not, the CI fails to compile `swift-nio` on the Swift toolchain nightly build.  But this is a separate issue and will not block the merging of this change.

Package.swift
```diff
@@ -85,6 +85,12 @@ let package = Package(
                 .byName(name: "AWSLambdaTesting")
             ]
         ),
+        .executableTarget(
```
sebsto, line 88: I think this was for your own test. I think we can remove this

Sources/HelloWorldLambda/main.swift
```diff
@@ -0,0 +1,29 @@
+//===----------------------------------------------------------------------===//
```
sebsto, line 1: This file can be remove for the PR too,

sebsto (commented):
With the exception of the example code, LGTM

Issue #494, opened 2025-03-06
[test] Amazon S3 with AWS SDK fails to compile


sebsto:
Check 
https://github.com/swift-server/swift-aws-lambda-runtime/actions/runs/13701827010/job/38317369346

sebsto:
Will be fixed by https://github.com/swift-server/swift-aws-lambda-runtime/pull/496

sebsto:
fix by https://github.com/swift-server/swift-aws-lambda-runtime/pull/496

PR #496, opened 2025-03-06
Fix CI error on Amazon S3 examples

sebsto:
When we include Traits, we changed the CI container from Swift 6.0 Amazon Linux to Swift 6.1 Nightly on Ubuntu.
This broke the CI because AWS SDK used in the `Examples/S3_AWSSDK` example depends on `libssl-devel` which is not installed by default on Jammy images.

This PR reverts back the container used to test all examples to an Amazon Linux 2, but this time the 6.1-nightly version to allow to compile the package traits.

This should address https://github.com/swift-server/swift-aws-lambda-runtime/issues/494


fabianfett (approved):
Love it!

PR #500, opened 2025-03-07
[test] add a unit test for the LambdaHTTPServer Pool

sebsto:


sebsto:
@fabianfett any feedback or concerns about that one

Issue #481, opened 2025-02-19
Local Debugging: LambdaHttpServer Drops All But Last Inbound Body Chunk


macserv:
### Expected behavior

When handling the connection, all body data chunks should be collected into the `requestBody` buffer from the inbound stream for processing.

### Actual behavior

The `requestBody` buffer is repeatedly reset to the last chunk received, so that only the last chunk is sent for processing.

### Steps to reproduce

1. Create a sample `StreamingLambdaHandler` and run it locally.
2. Feed a decent amount of data to the `invoke` endpoint (my test data is around 260KB, but I don't think it needs to be anywhere near that large to get split into chunks).
3. Log the event data.
4. Observe that only the last chunk of body data is processed.

## Sample `StreamingLambdaHandler`

```
@main
struct RawInputHandler: StreamingLambdaHandler
{
    func handle(_ event: ByteBuffer, responseWriter: some LambdaResponseStreamWriter, context: LambdaContext) async throws
    {
        context.logger.info("Raw JSON:\n\( String(data: Data(buffer: event), encoding: .utf8)! )")
        try await responseWriter.finish()
    }

    static func main() async throws { try await LambdaRuntime(handler: Self()).run() }
}
```

## Invocation with `curl`

`cat ./TestData.json | curl --json '@-' 'http://localhost:7000/invoke'`

## Root Cause

The issue seems to reside in the `handleConnection()` function in [/Sources/AWSLambdaRuntimeCore/Lambda+LocalServer.swift:163–204](https://github.com/swift-server/swift-aws-lambda-runtime/blob/40e2291532fdc0cf2e54f285bf33dc37b740646f/Sources/AWSLambdaRuntimeCore/Lambda%2BLocalServer.swift#L163C1-L204):

```
    /// This method handles individual TCP connections
    private func handleConnection(
        channel: NIOAsyncChannel<HTTPServerRequestPart, HTTPServerResponsePart>
    ) async {

        var requestHead: HTTPRequestHead!
        var requestBody: ByteBuffer?

        // Note that this method is non-throwing and we are catching any error.
        // We do this since we don't want to tear down the whole server when a single connection
        // encounters an error.
        do {
            try await channel.executeThenClose { inbound, outbound in
                for try await inboundData in inbound {
                    if case .head(let head) = inboundData {
                        requestHead = head
                    }
                    if case .body(let body) = inboundData {
                        requestBody = body
                    }
                    if case .end = inboundData {
[...]

PR #504, opened 2025-03-12
[core] Fix LocalServer failure to handle large body request (Issue #481)

sebsto:
Fix https://github.com/swift-server/swift-aws-lambda-runtime/issues/481


fabianfett (changes requested):
We should fix this by adding streaming support here.

sebsto:
@fabianfett Let's separate concerns. This fix handles the incoming data to the local server (what the client sends us 
Let's open a separate issue and PR to fix the outgoing data, i.e. the streaming response

sebsto:
@fabianfett We need to take a decision. 
I understand your preference is to revert back to non NIOAsync implementation of the local server. 
But while waiting for this, can we merge this one line change that will unblock https://github.com/swift-server/swift-aws-lambda-runtime/issues/481 ?

Then we can plan for a proper revert to v1 implementation.

sebsto:
@fabianfett as this is a blocker for several persons, I decided to move forward by merging this. We can resume the discussion about NIOAsync in a separate issue / PR.

@Andrea-Scuderi heard loud and clear - merged today

PR #508, opened 2025-03-14
[core] Only one `LambdaRuntime.run()` can be called at a time  (fix #507)

### LINKED ISSUE #507: [core] user can create multiple instances of the Lambda Runtime
Now that user controls the `main` function with the runtime v2, nothing prevents users to create multiple runtimes.

I tested the code below on Lambda and it works. Any of the two runtimes can pickup the invocation and  respond.  In my exemple, they're booth taking the same input parameter and there is not too much consequences. But if the user codes the two handler with different input types, the Lambda function will crash when the "incorrect" runtime picks up the event.

We should enforce the runtime as a singleton.

**demo code**
```
import AWSLambdaRuntime

let runtime1 = LambdaRuntime {
    (event: String, context: LambdaContext) in
    "Hello1 \(event)"
}

let runtime2 = LambdaRuntime {
    (event: String, context: LambdaContext) in
    "Hello2 \(event)"
}

try await withThrowingTaskGroup(of: Void.self) { group in
    group.addTask {
        try await runtime1.run()
    }
    group.addTask {
        try await runtime2.run()
    }

    try await group.waitForAll()
}

```

sebsto:
This is a proposal to fix issue #507 

**changes**
- `LambdaRuntime.init()` uses a `Mutex<Bool>` to make sure only one instance is created
- `LambdaRuntime.init()` can now throw an error in case an instance already exists (I did not use `fatalError()` to make it easier to test)
- All `convenience init()` methods catch possible errors instead of re-throwing it to a void breaking the user-facing API 
- Renamed existing `LambdaRuntimeError` to `LambdaRuntimeClientError`
- Introduced a new type `LambdaRuntimeError` to represent the double initialization error

### DIFF
--- a/Sources/AWSLambdaRuntime/LambdaRuntime.swift
+++ b/Sources/AWSLambdaRuntime/LambdaRuntime.swift
@@ -29,7 +29,7 @@ private let _isRunning = Atomic<Bool>(false)
 public final class LambdaRuntime<Handler>: Sendable where Handler: StreamingLambdaHandler {
     @usableFromInline
     /// we protect the handler behind a Mutex to ensure that we only ever have one copy of it
-    let handlerMutex: SendingStorage<Handler>
+    let handlerStorage: SendingStorage<Handler>
     @usableFromInline
     let logger: Logger
     @usableFromInline
@@ -80,7 +80,7 @@ public final class LambdaRuntime<Handler>: Sendable where Handler: StreamingLamb
         }
 
         // The handler can be non-sendable, we want to ensure we only ever have one copy of it
-        let handler = try? self.handlerMutex.get()
+        let handler = try? self.handlerStorage.get()
         guard let handler else {
             throw LambdaRuntimeError(code: .runtimeCanOnlyBeStartedOnce)
         }


sebsto:
CI is now green, except API Breakage, which is expected.

Examples/BackgroundTasks/Sources/main.swift
```diff
@@ -53,5 +53,5 @@ struct BackgroundProcessingHandler: LambdaWithBackgroundProcessingHandler {
 }
 
 let adapter = LambdaCodableAdapter(handler: BackgroundProcessingHandler())
-let runtime = LambdaRuntime.init(handler: adapter)
+let runtime = try LambdaRuntime.init(handler: adapter)
```
fabianfett, line 56: I think creating multiple handlers should be fine. We should throw in the `run` method, if another lambda has already started.

Sources/AWSLambdaRuntime/LambdaRuntime.swift
```diff
@@ -15,16 +15,24 @@
...
 #if canImport(FoundationEssentials)
 import FoundationEssentials
 #else
 import Foundation
 #endif
 
+// This is our gardian to ensure only one LambdaRuntime is initialized
+// We use a Mutex here to ensure thread safety
+// We use Bool instead of LambdaRuntime<Handler> as the type here, as we don't know the concrete type that will be used
+private let _singleton = Mutex<Bool>(false)
```
fabianfett, line 29: We can use an Atomic here instead. Singleton is fine for this.

Sources/AWSLambdaRuntime/LambdaRuntimeClientError.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-package struct LambdaRuntimeError: Error {
+package struct LambdaRuntimeClientError: Error {
```
fabianfett, line 15: Why the rename? I don't think the `Client` adds more info here.

fabianfett (changes requested):
Great spot! Thanks for tackling this one.

Sources/AWSLambdaRuntime/LambdaRuntimeClientError.swift
sebsto: @fabianfett  because this class only deals with errors from the HTTP Client (`LambdaRuntimeClient`) and is internal to the library.

The new `LambdaRuntimeError` is public, exposed to users, and represent a user error when using the Runtime, in this case, not respecting the singleton

Examples/BackgroundTasks/Sources/main.swift
sebsto: Good idea, l see the benefit of this approach. Let me work on this

Sources/AWSLambdaRuntime/LambdaRuntime.swift
sebsto: OK, will try to use Atomic instead of a Mutex

sebsto:
@fabianfett 
I changed 
- `Mutex` for `Atomic` 
- Test is now on `run()` and not on `init()`
-  `LamabdaRuntimeError` returned and is now public + one additional error case 
-  unit test is adjusted and pass

Tests/AWSLambdaRuntimeTests/LambdaRuntimeTests.swift
```diff
@@ -0,0 +1,83 @@
...
+
+            let nonReturningTask = Task.detached(priority: .userInitiated) {
+                try await runtime2.run()
+            }
+
+            // Set timeout and cancel the runtime 2
+            try await Task.sleep(for: .seconds(2))
+            nonReturningTask.cancel()
+
+        }
```
fabianfett, line 71: Again we should use structured concurrency here
```suggestion
        // Running the second runtime should work now
        try await withThrowingTaskGroup(of: Void.self) { taskGroup in
            taskGroup.addTask {
                await #expect(throws: Never.self) { try await runtime2.run() }
            }

            // Set timeout and cancel the runtime 2
            try await Task.sleep(for: .seconds(2))
            taskGroup.cancelAll()
        }
```

fabianfett (dismissed):
some golden rules:
1. we should make sure we never add enums that we want to extend after the release to our public API
2. use structured concurrency when possible

Tests/AWSLambdaRuntimeTests/LambdaRuntimeTests.swift
```diff
@@ -0,0 +1,83 @@
...
+
+        // cancel runtime 1 / task 1
+        print("--- cancelling ---")
+        task1.cancel()
+
+        // Running the second runtime should work now
+        await #expect(throws: Never.self) {
+
+            let nonReturningTask = Task.detached(priority: .userInitiated) {
+                try await runtime2.run()
```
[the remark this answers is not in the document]
sebsto: There is a problem in the current code, `run()` never returns, even when cancelled.

This call blocks 
```
            await #expect(throws: CancellationError.self) {
               _ = try await nonReturningTask.value
            }
```

sebsto:
Now that managing cancellation is working (thank you @adam https://github.com/swift-server/swift-aws-lambda-runtime/commit/b2811a5e1a96a98033ad857d9f93ff086dbba142)

This PR is ready to review before merged.
API Breakage is expected as we are adding an error case.

Sources/AWSLambdaRuntime/LambdaRuntime.swift
```diff
@@ -15,16 +15,21 @@
...
 import NIOCore
+import Synchronization
 
 #if canImport(FoundationEssentials)
 import FoundationEssentials
 #else
 import Foundation
 #endif
 
+// This is our gardian to ensure only one LambdaRuntime is running at the time
```
adam-fowler:
```suggestion
// This is our guardian to ensure only one LambdaRuntime is running at the time
```

Sources/AWSLambdaRuntime/LambdaRuntime.swift
```diff
@@ -58,8 +63,22 @@ public final class LambdaRuntime<Handler>: @unchecked Sendable where Handler: St
...
+
+        // if the original value was already true, run() is already running
+        if original {
+            throw LambdaRuntimeError(code: .moreThanOneLambdaRuntimeInstance)
+        }
+
+        defer {
+            _isRunning.store(false, ordering: .releasing)
+        }
+
```
adam-fowler, line 81: Do we need the check for multiple versions of `LambdaRuntime` below this code if we are also checking that only one `LambdaRuntime` is running at any point?

sebsto: I don't understand this comment. There is no check that multiple LambdaRuntime are created. The only check is at lines 72-77.

The defer just reset the global state to false to allow another `run()`

Fabian suggested to move the check from `init()` to `run()`. The guardian is a global variable, not an instance one. So the user can create multiple instances but only launch one `run()` at the time.

What do you think must be added or removed ?

Sources/AWSLambdaRuntime/Lambda+LocalServer.swift
```diff
@@ -202,12 +201,18 @@ internal struct LambdaHTTPServer {
...
 
             case .serverReturned(let result):
-                logger.error(
-                    "Server shutdown before closure completed",
-                    metadata: [
-                        "error": "\(result.maybeError != nil ? "\(result.maybeError!)" : "none")"
-                    ]
-                )
+
+                if (result.maybeError as? CancellationError) != nil {
```
adam-fowler:
```suggestion
                if result.maybeError is CancellationError {
```

Sources/AWSLambdaRuntime/LambdaRuntime.swift
```diff
@@ -39,7 +40,7 @@ public final class LambdaRuntime<Handler>: @unchecked Sendable where Handler: St
         eventLoop: EventLoop = Lambda.defaultEventLoop,
         logger: Logger = Logger(label: "LambdaRuntime")
     ) {
-        self.handlerMutex = NIOLockedValueBox(handler)
+        self.handlerMutex = SendingStorage(handler)
```
adam-fowler: Its not really a Mutex now
```suggestion
        self.handlerStorage = SendingStorage(handler)
```

adam-fowler (approved):
Looks good, only a couple of minor issues

PR #519, opened 2025-06-29
Add support for local server graceful shutdown

adam-fowler:
- Add ServiceLifecycle version of `LambdaRuntime.run` that wraps internal `_run` call in `cancelOnGracefulShutdown`
- Add cancellation handlers for shutting down existing connections in Local lambda
- Added test for lambda graceful shutdown

### Motivation:

Ensure local lambda supports graceful shutdown


sebsto (approved):
Thank you !

Issue #520, opened 2025-06-29
[core] Runtime does not support AWS Lambda Streaming specification


sebsto:
### Expected behavior

The current version of the runtime allows user handlers functions to return data as a stream. But this is a welcome side effect of the design of the swift runtime. The runtime does not follow the Lambda Runtime API specifications.

As a consequence, it is not possible to return user HTTP headers to a Lambda function URL. Also, it is not possible to interrupt the stream with an error that would be reported to the caller.

Specifically, the Lambda Runtime API specification mandates that a custom runtime return specifics HTTP headers and then use HTTP chunks to return the content 
See https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html#runtimes-custom-response-streaming

See Rust's implementation here 
https://github.com/awslabs/aws-lambda-rust-runtime/blob/62a36b04ddfd67000d407d06d0da60f394ca4b36/lambda-runtime/src/requests.rs#L103

### Actual behavior

It is not possible to return a custom HTTP header to a function URL or to interrupt the stream in case an error occur

### Steps to reproduce

n/a

### If possible, minimal yet complete reproducer code (or URL to code)

n/a

### What version of this project (`swift-aws-lambda-runtime`) are you using?

main

### Swift version

6.x

sebsto:
This was reported by @jpenwith at https://github.com/swift-server/swift-aws-lambda-runtime/issues/515

sebsto:
Example of client code in JS 
https://ravensmove.com/aws-lambda-function-response-streaming
https://www.presidio.com/introduction-to-response-streaming-with-aws-lambda/

sebsto:
Blog from Julian, with more hints about content chunks
https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/
https://github.com/aws-samples/serverless-patterns/tree/main/lambda-streaming-sdk-sam

sebsto:
This is the NodeJS Lambda runtime implemntation 

HTTP response Stream
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/main/src/HttpResponseStream.js

ResponseStream
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/main/src/ResponseStream.js

jpenwith:
Thanks, those examples are really useful! I didn’t realize it was an AWS issue

sebsto:
[...]

PR #521, opened 2025-06-29
[core] Implement Lambda streaming with custom HTTP headers

sebsto:
Fix https://github.com/swift-server/swift-aws-lambda-runtime/issues/520

### DIFF
--- a/Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
+++ b/Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
@@ -44,7 +44,7 @@ public struct StreamingLambdaStatusAndHeadersResponse: Codable, Sendable {
     public init(
         statusCode: Int,
         headers: [String: String]? = nil,
-        multiValueHeaders: [String: [String]]? = nil,
+        multiValueHeaders: [String: [String]]? = nil
     ) {
         self.statusCode = statusCode
         self.headers = headers
--- a/Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
+++ b/Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
@@ -44,7 +44,7 @@ public struct StreamingLambdaStatusAndHeadersResponse: Codable, Sendable {
     public init(
         statusCode: Int,
         headers: [String: String]? = nil,
-        multiValueHeaders: [String: [String]]? = nil
+        multiValueHeaders: [String: [String]]? = nil,
     ) {
         self.statusCode = statusCode
         self.headers = headers
@@ -67,17 +67,15 @@ extension LambdaResponseStreamWriter {
         encoder: Encoder
     ) async throws where Encoder.Output == StreamingLambdaStatusAndHeadersResponse {
 
-        // Convert Data to ByteBuffer
+        // Convert JSON headers to an array of bytes in a ByteBuffer
         var buffer = ByteBuffer()
         try encoder.encode(response, into: &buffer)
 
-        // Write the JSON data
-        try await write(buffer)
-
         // Write eight null bytes as separator
-        var separatorBuffer = ByteBuffer()
-        separatorBuffer.writeBytes([0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
-        try await write(separatorBuffer)
+        buffer.writeBytes([0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
+
+        // Write the JSON data and the separator
+        try await writeCustomHeader(buffer)
     }
 }
 
@@ -95,6 +93,7 @@ extension LambdaResponseStreamWriter {
         _ response: StreamingLambdaStatusAndHeadersResponse,
         encoder: JSONEncoder = JSONEncoder()
     ) async throws {
+        encoder.outputFormatting = .withoutEscapingSlashes
         try await self.writeStatusAndHeaders(response, encoder: LambdaJSONOutputEncoder(encoder))
     }
 }
--- a/Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
+++ b/Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
@@ -1,0 +1,100 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2017-2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
+/// A response structure specifically designed for streaming Lambda responses that contains
+/// HTTP status code and headers without body content.
+///
+/// This structure is used with `LambdaResponseStreamWriter.writeStatusAndHeaders(_:)` to send
+/// HTTP response metadata before streaming the response body.
+public struct StreamingLambdaStatusAndHeadersResponse: Codable, Sendable {
+    /// The HTTP status code for the response (e.g., 200, 404, 500)
+    public let statusCode: Int
+
+    /// Dictionary of single-value HTTP headers
+    public let headers: [String: String]?
+
+    /// Dictionary of multi-value HTTP headers (e.g., Set-Cookie headers)
+    public let multiValueHeaders: [String: [String]]?
+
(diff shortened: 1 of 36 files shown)


Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
```diff
@@ -0,0 +1,99 @@
...
+/// HTTP status code and headers without body content.
+///
+/// This structure is used with `LambdaResponseStreamWriter.writeStatusAndHeaders(_:)` to send
+/// HTTP response metadata before streaming the response body.
+public struct StreamingLambdaStatusAndHeadersResponse: Codable, Sendable {
+    /// The HTTP status code for the response (e.g., 200, 404, 500)
+    public let statusCode: Int
+
+    /// Dictionary of single-value HTTP headers
+    public let headers: [String: String]?
```
adam-fowler, line 33: Any reason we aren't use swift-http-types here?

Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
```diff
@@ -0,0 +1,99 @@
...
+        self.headers = headers
+        self.multiValueHeaders = multiValueHeaders
+    }
+}
+
+extension LambdaResponseStreamWriter {
+    /// Writes the HTTP status code and headers to the response stream.
+    ///
+    /// This method serializes the status and headers as JSON and writes them to the stream,
+    /// followed by eight null bytes as a separator before the response body.
```
adam-fowler, line 59: Is there any documentation anywhere we can link to that provides a reason for writing 8 null bytes

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
```diff
@@ -425,13 +443,17 @@ extension LambdaRuntimeClient: LambdaChannelHandlerDelegate {
 
             }
         }
+    }
 
+    func hasStreamingCustomHeaders(isolation: isolated (any Actor)? = #isolation) async -> Bool {
```
adam-fowler: Do we need the isolation parameter here?

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
```diff
@@ -188,6 +197,10 @@ final actor LambdaRuntimeClient: LambdaRuntimeClientProtocol {
         }
     }
 
+    private func writeCustomHeader(_ buffer: NIOCore.ByteBuffer) async throws {
+        _hasStreamingCustomHeaders = true
```
adam-fowler: Any reason you don't pass `_hasStreamingCustomHeaders` as a parameter through all the functions.
Otherwise if you are going to add some state perhaps you should add it to the `LambdaRuntimeClient.state`.

adam-fowler (commented):
Some comments

Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
sebsto: Because there is no dependency on HTTPTypes in this library and I did not want to pull a new dependency just for this. The data format is well controlled. It's between the Swift runtime and the Lambda data plane.

Sources/AWSLambdaRuntime/LambdaResponseStreamWriter+Headers.swift
sebsto: This is undocumented at the moment I received the information from an SDM in the Lambda service team and I have confirmed by looking at the NodeJS runtime implementation.
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/main/src/HttpResponseStream.js

The value of the "magic" header is here 
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/a5ae1c2a92708e81c9df4949c60fd9e1e6e46bed/src/HttpResponseStream.js#L17

The 8 x 0 bytes are defined here 
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/a5ae1c2a92708e81c9df4949c60fd9e1e6e46bed/src/HttpResponseStream.js#L26

BTW, at the moment, the NodeJS runtime is the only runtime supporting this capability. When we will release the Swift runtime, it will be second to offer this possibility, before all the AWS managed runtime

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
sebsto: Yes, because the user handler only has access to the `Writer` API and I need to pass that information to the actual code that sends the bytes to the Lambda Data plane. This happens in the `LambdaChannelHandler` which is created by actor-isolated `LambdaRuntimeClient`

The chain is 

1. user invoke a special `write()` method on the writer 
2. the writer passes the call to actor-isolated `LambdaRuntime` 
3. the `LambdaRuntime` creates a `LambdaChannelHandler` to manage the connection 

Because `LambdaRuntime` is actor-isolated, Fabian choose to pass the isolation context to functions in `LambdaChannelhandler`. See [`nextInvocation()`](https://github.com/swift-server/swift-aws-lambda-runtime/blob/db7e7897eba75b0eb82f9135bf9f18a4a0737a24/Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift#L493) as an example.

I used the same technique allowing to share a mutable state from the `LambdaRuntime` from its `LambdaChannelhandler`. Because it is read only in the `LambdaChannelHandler` there is no risk of data race.

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
sebsto: That's a good point.  The `_hasStreamingCustomHeaders` is indeed a state of the current response (not the connection, as it can be reused for multiple next() and their responses)

I'll check if I can move this flag to one of the many states struct present in the `LambdaRuntimeClient` :-)

sebsto: @adam-fowler
Regardless or where I store this value (in an existing state, if appropriate, or in a Bool) ...
What is best in terms of API? Adding that flag (with a default value) to all `write()` methods on the protocol ? Or add a new fun to the protocol ?

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
sebsto: When I do not pass the isolation parameter from the caller, I receive this compiler error 

```
swift-aws-lambda-runtime/Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift:672:36: error: sending 'self.delegate' risks causing data races
670 | 
671 |             var headers = self.streamingHeaders
672 |             if await self.delegate.hasStreamingCustomHeaders() {
    |                                    |- error: sending 'self.delegate' risks causing data races
    |                                    `- note: sending 'isolation'-isolated 'self.delegate' to nonisolated instance method 'hasStreamingCustomHeaders()' risks causing data races between nonisolated and 'isolation'-isolated uses
673 |                 // this header is required by Function URL when the user sends custom status code or headers
674 |                 headers.add(name: "Content-Type", value: "application/vnd.awslambda.http-integration-response")
[9/9] Compiling AWSLambdaRuntime Lambda+JSON.swift
```

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
sebsto: @adam-fowler after having reviewed the existing code, I conclude that there is an easier way to pass that flag down to the channel handler.  Look at the new version. Thank you for having challenged me on this!

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
sebsto: This is now removed and replaced by a much simpler version where I pass the `hasStreamingHeaders` directly to the `write()` functions which propagates the value down the the actual `ChannelHandler`

sebsto:
@adam-fowler ready to give it another review

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
```diff
@@ -596,18 +582,29 @@ private final class LambdaChannelHandler<Delegate: LambdaChannelHandlerDelegate>
...
+            try await self.sendResponseBodyPart(
+                byteBuffer,
+                sendHeadWithRequestID: requestID,
+                context: context,
+                hasCustomHeaders: hasCustomHeaders
+            )
 
         case .connected(let context, .sendingResponse):
-            try await self.sendResponseBodyPart(byteBuffer, sendHeadWithRequestID: nil, context: context)
+            try await self.sendResponseBodyPart(
```
adam-fowler, line 604: It is invalid to be have custom headers here, perhaps you should add a precondition for that, or throw an error

adam-fowler (commented):
One minor thing

Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift
sebsto: Good point. I choose `precondition` because if this happens, it is a programming error in the library, not a user or environmental error

PR #523, opened 2025-06-30
[core] Give user the possibility to pass their logger to the lambda runtime

sebsto:
Allow user to give their logger to the LambdaRuntime.

### Motivation:

Overloaded versions of `LambdaRuntime.init` don't allow to pass a logger

### Modifications:

Add a `logger` parameter to overloaded versions of `LambdaRuntime.init`

### Result:

It is now possible to write 

```
        let runtime = LambdaRuntime(logger: Logger(label: "MyLogger"), body: handler)
```

### DIFF
--- a/Sources/AWSLambdaRuntime/LambdaHandlers.swift
+++ b/Sources/AWSLambdaRuntime/LambdaHandlers.swift
@@ -191,6 +191,7 @@ extension LambdaRuntime {
     /// - Parameters:
     ///   - encoder: The encoder object that will be used to encode the generic `Output` into a `ByteBuffer`.
     ///   - decoder: The decoder object that will be used to decode the incoming `ByteBuffer` event into the generic `Event` type.
+    ///   - logger: The logger to use for the runtime. Defaults to a logger with label "LambdaRuntime".
     ///   - body: The handler in the form of a closure.
     public convenience init<
         Event: Decodable,
@@ -200,6 +201,7 @@ extension LambdaRuntime {
     >(
         encoder: sending Encoder,
         decoder: sending Decoder,
+        logger: Logger = Logger(label: "LambdaRuntime"),
         body: sending @escaping (Event, LambdaContext) async throws -> Output
     )
     where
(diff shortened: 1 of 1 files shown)


Sources/AWSLambdaRuntime/FoundationSupport/Lambda+JSON.swift
```diff
@@ -85,10 +87,12 @@ extension LambdaCodableAdapter {
 extension LambdaRuntime {
     /// Initialize an instance with a `LambdaHandler` defined in the form of a closure **with a non-`Void` return type**.
     /// - Parameters:
+    ///   - logger: The logger to use for the runtime. Defaults to a logger with label "LambdaRuntime".
     ///   - decoder: The decoder object that will be used to decode the incoming `ByteBuffer` event into the generic `Event` type. `JSONDecoder()` used as default.
     ///   - encoder: The encoder object that will be used to encode the generic `Output` into a `ByteBuffer`. `JSONEncoder()` used as default.
     ///   - body: The handler in the form of a closure.
     public convenience init<Event: Decodable, Output>(
+        logger: Logger = Logger(label: "LambdaRuntime"),
```
adam-fowler: It is common for the logger parameter to be the last parameter (or at least the last before any optional parameters and closures).

Sources/AWSLambdaRuntime/LambdaHandlers.swift
```diff
@@ -175,11 +176,15 @@ public struct ClosureHandler<Event: Decodable, Output>: LambdaHandler {
 
 extension LambdaRuntime {
     /// Initialize an instance with a ``StreamingLambdaHandler`` in the form of a closure.
-    /// - Parameter body: The handler in the form of a closure.
+    /// - Parameter
+    ///   - logger: The logger to use for the runtime. Defaults to a logger with label "LambdaRuntime".
+    ///   - body: The handler in the form of a closure.
     public convenience init(
+        logger: Logger = Logger(label: "LambdaRuntime"),
```
adam-fowler, line 183: `LambdaRuntime.init(handler:eventLoop:logger)` forces the log level to be `info` if there is no `LOG_LEVEL` environment variable. Perhaps it should just use the supplied Logger logLevel if the environment variable is not there

Sources/AWSLambdaRuntime/FoundationSupport/Lambda+JSON.swift
sebsto: OK

Sources/AWSLambdaRuntime/LambdaHandlers.swift
sebsto: excellent idea. I wonder why we didn't do that earlier

adam-fowler (approved):

PR #525, opened 2025-07-01
[example] Group all examples under one project

sebsto:
All the examples are now under one Swift project (one `Package.swift` file)

### Motivation:

Building all the examples on my laptop took > 1h because each example downloaded and compiled the Lambda Runtime and its dependency (NIO)

### Modifications:

I remove each individual example's Package.swift and consolidated all the targets into a single file.

I adjusted the README for each example (instruct user to use `swift build --target xxx` insead of `swift build`, and the same for the `swift run` and `swift package archive commands`

I adjusted the CI script 

### Result:

Examples build in 4 minutes on my laptop instead of ~60 min


sebsto:
Closing this as this increases the build time during CI from ~8 min to ~20 min

PR #526, opened 2025-07-02
[WIP] DO NOT MERGE - [core] Remove `@unchecked Sendable` from Lambda Runtime (depends on PR#508)

sebsto:
This PR depends on https://github.com/swift-server/swift-aws-lambda-runtime/pull/508

When removing the `NIOLockedBox`, we wanted to remove the last `@unchecked Sendable`  in the code.

### Motivation:

There is no good reason to bypass compiler verifications with `@unchecked Sendable` in this scenario. We would like to bring the extra verifications' benefit to the user of the library.

### Modifications:

Lambda handlers and all depending structs and functions are tagged @Sendable

### Result:

This adds a requirement on user defined handlers (the public API of this library).
Code like 

```
let handler:
    (Event, LambdaContext) async throws -> Response = {
        (request: Event, context: LambdaContext) in
    ...
    return Response() 
}
let runtime = LambdaRuntime(body: handler) 
try await runtime().run() 
```

is no longer accepted.  The `handler` must marked `@Sendable` 

```
let handler:
    @Sendable (Event, LambdaContext) async throws -> Response = {
        (request: Event, context: LambdaContext) in
    ...
    return Response() 
}
```

Fortunately, thanks to the compiler type inference, this does not affect handlers passed as a closure, such as 

```
let runtime = LambdaRuntime {
    (event: HelloRequest, context: LambdaContext) in

    HelloResponse(
        greetings: "Hello \(event.name). You look \(event.age > 30 ? "younger" : "older") than your age."
    )
}
```

As we expect this is the preferred way to write a handler, we think the change in the public API is acceptable. Only more advanced users will have to explicitly add the `@Sendable` conformance to their standalone handler functions.

### DIFF
--- a/Examples/APIGateway+LambdaAuthorizer/Sources/AuthorizerLambda/main.swift
+++ b/Examples/APIGateway+LambdaAuthorizer/Sources/AuthorizerLambda/main.swift
@@ -22,7 +22,7 @@ import AWSLambdaRuntime
 // This code is shown for the example only and is not used in this demo.
 // This code doesn't perform any type of token validation. It should be used as a reference only.
 let policyAuthorizerHandler:
-    @Sendable (APIGatewayLambdaAuthorizerRequest, LambdaContext) async throws ->
+    (APIGatewayLambdaAuthorizerRequest, LambdaContext) async throws ->
         APIGatewayLambdaAuthorizerPolicyResponse = {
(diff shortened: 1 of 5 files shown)


sebsto:
We decided to not force `Sendable` on Handler. This commit should allow that
https://github.com/swift-server/swift-aws-lambda-runtime/pull/508/commits/7b501671a0e0dbf57fa3033397823d0ec0b6dc49

PR #529, opened 2025-07-07
[test] Add a test for cancellable

sebsto:
- Add a test on `LambdaRuntime` for cancellable.
- Move the service life cycle test to its own file

### DIFF
--- a/Tests/AWSLambdaRuntimeTests/LambdaRuntimeTests.swift
+++ b/Tests/AWSLambdaRuntimeTests/LambdaRuntimeTests.swift
@@ -50,7 +50,7 @@ struct LambdaRuntimeTests {
             }
 
             // wait a small amount to ensure runtime1 task is started
-            try await Task.sleep(for: .seconds(0.5))
+            try await Task.sleep(for: .seconds(1))
 
             // Running the second runtime should trigger LambdaRuntimeError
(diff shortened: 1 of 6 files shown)


Tests/AWSLambdaRuntimeTests/LambdaRuntimeTests.swift
```diff
@@ -50,17 +50,24 @@ struct LambdaRuntimeTests {
             }
 
             // wait a small amount to ensure runtime1 task is started
-            try await Task.sleep(for: .seconds(0.5))
```
adam-fowler, line 53: When I run all the tests in parallel this test fails quite regularly as the second task gets run before the first. I don't think you can rely on a sleep to ensure the two tasks are run in order. Maybe you should care which order they run in, just that one throws a `LambdaRuntimeError`. Do the error check before the `taskGroup.next()`
```swift
await #expect(throws: LambdaRuntimeError.self) {
    try await taskGroup.next()
}
```
I'm pretty sure you can get rid of the test that you can run the second runtime after the failure. Is it really valid to have two separate runtimes, run in the the same lambda?

sebsto: Thank you for the suggestion.  Fixed now.

adam-fowler (approved):

PR #532, opened 2025-07-15
[core] Add user-facing API for Streaming Lambda functions that receive JSON events

sebsto:
Add user-facing API for Streaming Lambda functions that receives JSON events

### Motivation:

Streaming Lambda functions developed by developers had no choice but to implement a handler that receives incoming data as a `ByteBuffer`. While this is useful for low-level development, I assume most developers will want to receive a JSON event to trigger their streaming Lambda function.

Going efficiently from a `ByteBuffer` to a Swift struct requires some code implemented in the `JSON+ByteBuffer.swift` file of the librray. We propose to further help developers by providing them with a new `handler()` function that directly receives their `Decodable` type.

### Modifications:

This PR adds a public facing API (+ unit test + updated README) allowing developers to write a handler method accepting any `Decodable` struct as input.

```swift
import AWSLambdaRuntime
import NIOCore

// Define your input event structure
struct StreamingRequest: Decodable {
    let count: Int
    let message: String
    let delayMs: Int?
}

// Use the new streaming handler with JSON decoding
let runtime = LambdaRuntime { (event: StreamingRequest, responseWriter, context: LambdaContext) in
    context.logger.info("Received request to send \(event.count) messages")
    
    // Stream the messages
    for i in 1...event.count {
        let response = "Message \(i)/\(event.count): \(event.message)\n"
        try await responseWriter.write(ByteBuffer(string: response))
        
        // Optional delay between messages
        if let delay = event.delayMs, delay > 0 {
            try await Task.sleep(for: .milliseconds(delay))
        }
    }
    
    // Finish the stream
    try await responseWriter.finish()
    
    // Optional: Execute background work after response is sent
    context.logger.info("Background work: processing completed")
}

try await runtime.run()
```

This interface provides:
- **Type-safe JSON input**: Automatic decoding of JSON events into Swift structs
- **Streaming responses**: Full control over when and how to stream data back to clients  
- **Background work support**: Ability to execute code after the response stream is finished
- **Familiar API**: Uses the same closure-based pattern as regular Lambda handlers

Because streaming Lambda functions can be invoked either directly through the API or through Lambda Function URL, this PR adds the decoding logic to support both types, shielding developers from working with Function URL requests and base64 encoding. 

We understand these choice will have an impact on the raw performance for event handling. Those advanced users that want to get the maximum might use the existing `handler(_ event: ByteBuffer, writer: LambaStreamingWriter)` function to implement their own custom decoding logic.

This PR provides a balance between ease of use for 80% of the users vs ultimate performance, without closing the door for the 20% who need it. 

### Result:

Lambda function developers can now use arbitrary `Decodable` Swift struct or Lambda events to trigger their streaming functions. 🎉

### DIFF
--- a/Examples/Streaming/README.md
+++ b/Examples/Streaming/README.md
@@ -245,11 +245,6 @@ The content of the input `ByteBuffer` depends on how you invoke the function:
 Here is an example of Lambda function URL payload:
 
 ```json
-// This is an example of payload received when
-// the function is invoked by a Lambda function URL.
-// You can use the `FunctionURL` structure provided by the Lambda Event library to decode this
-// See, https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift
-
 /* 
 {
     "version": "2.0",
--- a/Examples/Streaming/README.md
+++ b/Examples/Streaming/README.md
@@ -239,15 +239,15 @@ sam delete
 
 The content of the input `ByteBuffer` depends on how you invoke the function:
 
-- when you invoke the function with the [`InvokeWithresponseStream` API](https://docs.aws.amazon.com/lambda/latest/api/API_InvokeWithResponseStream.html) to invoke the function, the function incoming payload is what you pass to the API. You can decode the `ByteBuffer` with a [`JSONDecoder.decode()`](https://developer.apple.com/documentation/foundation/jsondecoder) function call.
-- when you invoke the function through a [Lambda function URL](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html), the incoming `ByteBuffer` contains a payload that gives developer access to the underlying HTTP call. The payload contains information about the HTTP verb used, the headers received, the authentication method and so on. The [AWS documentation contains the details](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html) of the payload. The [Swift lambda Event library](https://github.com/swift-server/swift-aws-lambda-events) contains a [FunctionURL Swift struct definition](https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift) ready to use in your projects.
+- when you use [`InvokeWithResponseStream` API](https://docs.aws.amazon.com/lambda/latest/api/API_InvokeWithResponseStream.html) to invoke the function, the function incoming payload is what you pass to the API. You can decode the `ByteBuffer` with a [`JSONDecoder.decode()`](https://developer.apple.com/documentation/foundation/jsondecoder) function call.
+- when you invoke the function through a [Lambda function URL](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html), the incoming `ByteBuffer` contains a payload that gives developer access to the underlying HTTP call. The payload contains information about the HTTP verb used, the headers received, the authentication method and so on. The [AWS documentation contains the details](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html) of the payload. The [Swift Lambda Event library](https://github.com/swift-server/swift-aws-lambda-events) contains a [`FunctionURL` type](https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift) ready to use in your projects.
 
 Here is an example of Lambda function URL payload:
 
-```
+```json
 // This is an example of payload received when
 // the function is invoked by a Lambda function URL.
-// You can use the `FunctionURL`` structure provided by the Lambda Event library to decode this
+// You can use the `FunctionURL` structure provided by the Lambda Event library to decode this
 // See, https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift
 
 /* 
--- a/Examples/Streaming/README.md
+++ b/Examples/Streaming/README.md
@@ -220,3 +220,69 @@ When done testing, you can delete the infrastructure with this command.
 ```bash
 sam delete 
 ```
+
+## Payload decoding
+
+The content of the input `ByteBuffer` depends on how you invoke the function:
+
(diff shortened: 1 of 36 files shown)


Examples/Streaming/README.md
```diff
@@ -220,3 +220,69 @@ When done testing, you can delete the infrastructure with this command.
 ```bash
 sam delete 
 ```
+
+## Payload decoding
+
+The content of the input `ByteBuffer` depends on how you invoke the function:
+
+- when you invoke the function with the [`InvokeWithresponseStream` API](https://docs.aws.amazon.com/lambda/latest/api/API_InvokeWithResponseStream.html) to invoke the function, the function incoming payload is what you pass to the API. You can decode the `ByteBuffer` with a [`JSONDecoder.decode()`](https://developer.apple.com/documentation/foundation/jsondecoder) function call.
```
0xTim:
```suggestion
- when you use [`InvokeWithResponseStream` API](https://docs.aws.amazon.com/lambda/latest/api/API_InvokeWithResponseStream.html) to invoke the function, the function incoming payload is what you pass to the API. You can decode the `ByteBuffer` with a [`JSONDecoder.decode()`](https://developer.apple.com/documentation/foundation/jsondecoder) function call.
```

Examples/Streaming/README.md
```diff
@@ -220,3 +220,69 @@ When done testing, you can delete the infrastructure with this command.
...
+- when you invoke the function through a [Lambda function URL](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html), the incoming `ByteBuffer` contains a payload that gives developer access to the underlying HTTP call. The payload contains information about the HTTP verb used, the headers received, the authentication method and so on. The [AWS documentation contains the details](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html) of the payload. The [Swift lambda Event library](https://github.com/swift-server/swift-aws-lambda-events) contains a [FunctionURL Swift struct definition](https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift) ready to use in your projects.
```
0xTim:
```suggestion
- when you invoke the function through a [Lambda function URL](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html), the incoming `ByteBuffer` contains a payload that gives developer access to the underlying HTTP call. The payload contains information about the HTTP verb used, the headers received, the authentication method and so on. The [AWS documentation contains the details](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html) of the payload. The [Swift Lambda Event library](https://github.com/swift-server/swift-aws-lambda-events) contains a [`FunctionURL` type](https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift) ready to use in your projects.
```

Examples/Streaming/README.md
```diff
@@ -220,3 +220,69 @@ When done testing, you can delete the infrastructure with this command.
...
+
+Here is an example of Lambda function URL payload:
+
+```
+// This is an example of payload received when
+// the function is invoked by a Lambda function URL.
+// You can use the `FunctionURL`` structure provided by the Lambda Event library to decode this
```
0xTim:
```suggestion
// You can use the `FunctionURL` structure provided by the Lambda Event library to decode this
```

Examples/Streaming/README.md
```diff
@@ -220,3 +220,69 @@ When done testing, you can delete the infrastructure with this command.
...
+- when you invoke the function through a [Lambda function URL](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html), the incoming `ByteBuffer` contains a payload that gives developer access to the underlying HTTP call. The payload contains information about the HTTP verb used, the headers received, the authentication method and so on. The [AWS documentation contains the details](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html) of the payload. The [Swift lambda Event library](https://github.com/swift-server/swift-aws-lambda-events) contains a [FunctionURL Swift struct definition](https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift) ready to use in your projects.
+
+Here is an example of Lambda function URL payload:
+
+```
```
0xTim: Should we mark this as JSON?

```suggestion
```json
```

Examples/Streaming/README.md
```diff
@@ -220,3 +220,69 @@ When done testing, you can delete the infrastructure with this command.
...
+// This is an example of payload received when
+// the function is invoked by a Lambda function URL.
+// You can use the `FunctionURL`` structure provided by the Lambda Event library to decode this
+// See, https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift
+
+/*
```
0xTim: Remove?

Examples/StreamingFromEvent/Package.swift
```diff
@@ -0,0 +1,50 @@
...
+        .executableTarget(
+            name: "StreamingFromEvent",
+            dependencies: [
+                .product(name: "AWSLambdaRuntime", package: "swift-aws-lambda-runtime")
+            ]
+        )
+    ]
+)
+
+if let localDepsPath = Context.environment["LAMBDA_USE_LOCAL_DEPS"],
```
0xTim, line 25: Something for a potential future PR - it would be good to simplify these example manifest and remove all of this in favour of a `swift package edit` in CI

sebsto: OK, tracked here https://github.com/swift-server/swift-aws-lambda-runtime/issues/536
However we should also consider development of the library itself and the examples. It's convenient to build with `AS_LAMBDA_LOCAL_DEPS=../..`

sebsto:
@0xTim Thank you for your review and suggestions. All have been addressed

Examples/Streaming/README.md
```diff
@@ -239,15 +239,15 @@ sam delete
...
 
-```
+```json
 // This is an example of payload received when
 // the function is invoked by a Lambda function URL.
-// You can use the `FunctionURL`` structure provided by the Lambda Event library to decode this
+// You can use the `FunctionURL` structure provided by the Lambda Event library to decode this
 // See, https://github.com/swift-server/swift-aws-lambda-events/blob/main/Sources/AWSLambdaEvents/FunctionURL.swift
 
 /*
```
0xTim: Do we want to remove this opening comment? There's nothing to close it, and it doesn't make sense if we're displaying this as JSON

sebsto: I though I did !  Let me check again

sebsto: @0xTim fixed, apologies for the miss

Issue #404, opened 2024-10-10
[core] Make sure TRACE log level exists in runtime v2


sebsto:
Verify if this change has been kept in v2 
https://github.com/swift-server/swift-aws-lambda-runtime/commit/221978224cf82027f46c0aa54def83476fb74412

PR #534, opened 2025-07-21
fix [core] trace level shows the first kb of the payload before being decoded

sebsto:
Add a log statement in the Lambda loop, before calling the user's handler to show the raw payload before any attempt to decode it.

This was available in Runtime v1 and is now ported to v2.
This fixes https://github.com/swift-server/swift-aws-lambda-runtime/issues/404

### Motivation:

This is useful when handling custom event and there is a Decoding error. It allows to see the exact payload received by the handler before any attempt to decode it.

### Modifications:

Add a log.trace statement with metatadata. Metadata are computed only when the log level is trace or below. 

### Result:

```
2025-07-21T08:58:33+0200 trace LambdaRuntime : Event's first bytes={"name": "me", "age": 50} aws-request-id=769127502334125 [AWSLambdaRuntime] sending invocation event to lambda handler
```


ktoso (approved):
Looks ok to me :)

PR #539, opened 2025-07-31
Use a struct for ClientContext (fix #169)

### LINKED ISSUE #169: ClientContext should be a type, not String
### Expected behavior
The `ClientContext` is often of well known shape:

It should handle the typing more gracefully I think, see docs: 
https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html

See also: https://github.com/swift-server/swift-aws-lambda-runtime/pull/167#discussion_r489236665 for discussion on this

### Actual behavior

It's a String today:

```
            let clientContext: String?
```

sebsto:
Do not use a String for Lambdacontext.ClientContext, use a struct instead
fix for https://github.com/swift-server/swift-aws-lambda-runtime/issues/169

Note: this PR introduces an API change that will break function using `LambdaContext`, we should integrate this change during the beta otherwise it will require a major version bump.

### Motivation:

Let the compiler detect type errors for us

### Modifications:

- Create a struct for ClientContext and it's embedded ClientApplication
- add three unit test to validate the struct 

### Result:

No more String?

### DIFF
--- a/Tests/AWSLambdaRuntimeTests/LambdaContextTests.swift
+++ b/Tests/AWSLambdaRuntimeTests/LambdaContextTests.swift
@@ -44,7 +44,7 @@ struct LambdaContextTests {
         let decodedClientContext = try decoder.decode(ClientContext.self, from: clientContextData)
 
         let decodedClient = try #require(decodedClientContext.client)
-        let originalClient = try #require(clientContext.client) 
+        let originalClient = try #require(clientContext.client)
 
         #expect(decodedClient.installationID == originalClient.installationID)
         #expect(decodedClient.appTitle == originalClient.appTitle)


sebsto:
Hello @ktoso 
This is the oldest open issue on this project 
https://github.com/swift-server/swift-aws-lambda-runtime/issues/169

You opened it 5 years ago :-)  
Can you review the change, and provide feedback or approve it ?

Thanks

Sources/AWSLambdaRuntime/LambdaContext.swift
```diff
@@ -16,6 +16,70 @@ import Dispatch
 import Logging
 import NIOCore
 
+// MARK: - Client Context
+
+/// AWS Mobile SDK client fields.
+public struct ClientApplication: Codable, Sendable {
+    /// The mobile app installation id
+    public let installationId: String?
```
ktoso: Nitpick but Swift style for such is `ID`, and since we're fixing the coding keys already we can just change the property name

```suggestion
    public let installationID: String?
```

PR #540, opened 2025-07-31
Remove dependency on DispatchWallTime (fix #384)

### LINKED ISSUE #384: Remove use of DispatchWallTime in LambdaContext
### Expected behavior

We should use another type for the deadline in LambdaContext than DispatchWallTime, which we currently use.

### Actual behavior

N/A

### Steps to reproduce

N/A

### If possible, minimal yet complete reproducer code (or URL to code)

_No response_

### What version of this project (`swift-aws-lambda-runtime`) are you using?

N/A

### Swift version

N/A

### Amazon Linux 2 docker image version

_No response_

sebsto:
Fix [#384](https://github.com/swift-server/swift-aws-lambda-runtime/issues/384)

Note: this PR introduces an API change that will break Lambda functions using `LambdaContext`, we should integrate this change during the beta otherwise it will require a major version bump.

### Motivation:

`DispatchWallTime` has no public API to extract the time in milliseconds, making it a dead end.
Previous implementation used the internal representation of time inside `DispatchWallTime` to extract the value, creating a risk if its implementation will change in the future.
Moreover, the use of `DispatchWallTime` obliges users to import the `Dispatch` library or `Foundation`.

Old Code:
```
extension DispatchWallTime {
    @usableFromInline
    init(millisSinceEpoch: Int64) {
        let nanoSinceEpoch = UInt64(millisSinceEpoch) * 1_000_000
        let seconds = UInt64(nanoSinceEpoch / 1_000_000_000)
        let nanoseconds = nanoSinceEpoch - (seconds * 1_000_000_000)
        self.init(timespec: timespec(tv_sec: Int(seconds), tv_nsec: Int(nanoseconds)))
    }

    var millisSinceEpoch: Int64 {
        Int64(bitPattern: self.rawValue) / -1_000_000
    }
}
```

Issue [#384](https://github.com/swift-server/swift-aws-lambda-runtime/issues/384) has a long discussion about possible replacements, including creating a brand new `UTCClock`, which I think is an overkill for this project.

Instead, I propose this simple implementation, based on two assumptions: 

- AWS always sends the time in milliseconds since Unix Epoch (1st Jan 1970) ([Lambda Runtime API documentation](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next))
- AWS always uses UTC time (not only for Lambda, this is a general rule for all AWS APIs)  ([TZ=UTC on Lambda](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html))

Therefore, this library just needs to store and make math on milliseconds since epoch, without having to care about timezone.

I had two possibilities to implement the storage and the math on milliseconds since Unix Epoch: either I could use an `UInt64` (as does [the Rust implementation](https://github.com/awslabs/aws-lambda-rust-runtime/blob/aff8d883c62997ef2615714dce9f7ddfd557147d/lambda-runtime/src/types.rs#L70)) or I could use a standard Swift type, such as `Duration`.

`Duration` is a good candidate for this because 1/ the time we receive from the Lambda Service API is indeed a duration between 1/1/1970 and the execution deadline for the Lambda function, expressed in milliseconds, 2/ it gives a strong type that can be verified by the compiler, and 3/ it is possible to do basic arithmetic operations and compare two values.

As an additional benefit, it allows library users to not import `Dispatch` or `Foundation`

### Modifications:

I made two changes:

1. I extend the `Duration` type to provide us with simple unix epoch time manipulation functions and values.

```swift
extension Duration {
    /// Returns the time in milliseconds since the Unix epoch.
    @usableFromInline
    static var millisSinceEpoch: Duration {
        var ts = timespec()
        clock_gettime(CLOCK_REALTIME, &ts)
        return .milliseconds(Int64(ts.tv_sec) * 1000 + Int64(ts.tv_nsec) / 1_000_000)
    }

    /// Returns a Duration between Unix epoch and the distant future
    @usableFromInline
    static var distantFuture: Duration {
        // Use a very large value to represent the distant future
        millisSinceEpoch + Duration.seconds(.greatestFiniteMagnitude)
    }

    /// Returns the Duration in milliseconds
    @usableFromInline
    func milliseconds() -> Int64 {
        Int64(self / .milliseconds(1))
    }

    /// Create a Duration from milliseconds since Unix Epoch
    @usableFromInline
    init(millisSinceEpoch: Int64) {
        self = .milliseconds(millisSinceEpoch)
    }
}
```

3. I replaced all references to `DispatchWallTime` by `Duration` 

### Result:

No more `DispatchWallTime`
No dependencies on Foundation, as I use `clock_gettime()` to get the epoch from the system clock.

### DIFF
--- a/Sources/MockServer/MockHTTPServer.swift
+++ b/Sources/MockServer/MockHTTPServer.swift
@@ -35,7 +35,7 @@ struct HttpServer {
     private let eventLoopGroup: MultiThreadedEventLoopGroup
     /// the mode. Are we mocking a server for a Lambda function that expects a String or a JSON document? (default: string)
     private let mode: Mode
-    /// the number of connections this server must accept before shutting down (default: 1)
+    /// the number of invocations this server must accept before shutting down (default: 1)
     private let maxInvocations: Int
     /// the logger (control verbosity with LOG_LEVEL environment variable)
     private let logger: Logger
@@ -91,10 +91,6 @@ struct HttpServer {
             ]
         )
 
-        // This counter is used to track the number of incoming connections.
-        // This mock servers accepts n TCP connection then shutdowns
-        let connectionCounter = SharedCounter(maxValue: self.maxInvocations)
-
         // We are handling each incoming connection in a separate child task. It is important
         // to use a discarding task group here which automatically discards finished child tasks.
         // A normal task group retains all child tasks and their outputs in memory until they are
@@ -105,21 +101,15 @@ struct HttpServer {
             try await channel.executeThenClose { inbound in
                 for try await connectionChannel in inbound {
 
-                    let counter = connectionCounter.current()
-                    logger.trace("Handling new connection", metadata: ["connectionNumber": "\(counter)"])
-
                     group.addTask {
-                        await self.handleConnection(channel: connectionChannel)
-                        logger.trace("Done handling connection", metadata: ["connectionNumber": "\(counter)"])
+                        await self.handleConnection(channel: connectionChannel, maxInvocations: self.maxInvocations)
+                        logger.trace("Done handling connection")
                     }
 
-                    if connectionCounter.increment() {
-                        logger.info(
-                            "Maximum number of connections reached, shutting down after current connection",
-                            metadata: ["maxConnections": "\(self.maxInvocations)"]
-                        )
-                        break  // this causes the server to shutdown after handling the connection
-                    }
+                    // This mock server only accepts one connection
+                    // the Lambda Function Runtime will send multiple requests on that single connection
+                    // This Mock Server closes the connection when MAX_INVOCATION is reached
+                    break
                 }
             }
         }
@@ -131,17 +121,19 @@ struct HttpServer {
     /// It handles two requests: one for the next invocation and one for the response.
     /// when the maximum number of requests is reached, it closes the connection.
     private func handleConnection(
-        channel: NIOAsyncChannel<HTTPServerRequestPart, HTTPServerResponsePart>
+        channel: NIOAsyncChannel<HTTPServerRequestPart, HTTPServerResponsePart>,
+        maxInvocations: Int
     ) async {
 
         var requestHead: HTTPRequestHead!
         var requestBody: ByteBuffer?
 
-        // each Lambda invocation results in TWO HTTP requests (next and response)
-        let requestCount = SharedCounter(maxValue: 2)
+        // each Lambda invocation results in TWO HTTP requests (GET /next and POST /response)
+        let maxRequests = maxInvocations * 2
+        let requestCount = SharedCounter(maxValue: maxRequests)
 
         // Note that this method is non-throwing and we are catching any error.
-        // We do this since we don't want to tear down the whole server when a single connection
+        // We do this since we don't want to tear down the whole server when a single request
         // encounters an error.
         do {
             try await channel.executeThenClose { inbound, outbound in
@@ -178,7 +170,7 @@ struct HttpServer {
                         if requestCount.increment() {
                             logger.info(
                                 "Maximum number of requests reached, closing this connection",
-                                metadata: ["maxRequest": "2"]
+                                metadata: ["maxRequest": "\(maxRequests)"]
                             )
                             break  // this finishes handiling request on this connection
                         }
--- a/scripts/performance_test.sh
+++ b/scripts/performance_test.sh
@@ -3,7 +3,7 @@
 ##
 ## This source file is part of the SwiftAWSLambdaRuntime open source project
 ##
-## Copyright (c) 2017-2022 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+## Copyright (c) 2017-2025 Apple Inc. and the SwiftAWSLambdaRuntime project authors
 ## Licensed under Apache License v2.0
 ##
 ## See LICENSE.txt for license information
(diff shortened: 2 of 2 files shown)


adam-fowler:
Using Duration for Instances seems wrong. Perhaps you could use a Clock in this form
```swift
public struct LambdaClock: Clock {
    public typealias Duration = Swift.Duration

    public struct Instant: InstantProtocol {
        public let date: Date
        public static var now: Self { .init(date: .now) }

        public typealias Duration = Swift.Duration

        public func advanced(by duration: Duration) -> Instant {
            .init(date: self.date.addingTimeInterval(duration / .seconds(1)))
        }

        public func duration(to other: Instant) -> Duration {
            .seconds(other.date.timeIntervalSince(self.date))
        }

        public static func < (lhs: Instant, rhs: Instant) -> Bool {
            lhs.date < rhs.date
        }
    }

    public var now: Instant { .init(date: Date.now) }

    public var minimumResolution: Duration {
        .nanoseconds(1)
    }

    public static var systemEpoch: Instant {
        .init(date: Date(timeIntervalSince1970: 0))
    }

    public func sleep(until deadline: Instant, tolerance: Instant.Duration?) async throws {
        try await ContinuousClock().sleep(
            until: .now.advanced(by: .seconds(deadline.date.timeIntervalSinceNow)),
            tolerance: tolerance
        )
    }
}
```
This is very similar to the UTCClock that is being suggested for Foundation, except it is 1970 based. Instead of using `Duration` for deadlines you can then use `LambdaClock.Instant`.

sebsto:
Thank you @adam-fowler for the suggestion.  I recognize the name duration is not great but the concept is exactly what I need: a period of time between two Instants.

I propose to not use your suggested code, because it uses `Date()` and I want to avoid relying on Foundation as much as possible. I think it has a precision of seconds and we need milliseconds here.

But I got your point and I will wrap `Duration` in a very simple struct that just implement what we need to support Lambda Deadlines

adam-fowler:
> Thank you @adam-fowler for the suggestion. I recognize the name duration is not great but the concept is exactly what I need: a period of time between two Instants.
> 

You are using `Duration` to define an instance in time, which is incorrect

> I propose to not use your suggested code, because it uses `Date()` and I want to avoid relying on Foundation as much as possible. I think it has a precision of seconds and we need milliseconds here.
> 

`Date` uses a `Double` for precision, so it can do millisecond precision. You are already using `FoundationEssentials` so using `Date` won't hurt

Buy yeah if you want to contain using Duration you could use

```
struct LambdaDeadline {
    let durationSince1970: Duration
}
```

or something similar

Sources/AWSLambdaRuntime/LambdaContext.swift
```diff
@@ -25,7 +25,7 @@ public struct LambdaContext: CustomDebugStringConvertible, Sendable {
         let requestID: String
         let traceID: String
         let invokedFunctionARN: String
-        let deadline: DispatchWallTime
+        let deadline: Duration
```
adam-fowler: You need to remove the Dispatch import at the top of the file

sebsto:
@adam-fowler Here it is again, this time with a `LambdaClock` that implements `Clock` and `Instant` and based on Unix epoch time, in milliseconds. It doesn't use Foundation's Date.

You're correct about the precision of `Date`. I was confused by its initializer `Date(timeIntervalSince1970:)` that [explicitly mentions that the parameter is expressed in seconds](https://developer.apple.com/documentation/foundation/date/init(timeintervalsince1970:)).

Sources/AWSLambdaRuntime/LambdaContext.swift
```diff
@@ -173,30 +172,28 @@ public struct LambdaContext: CustomDebugStringConvertible, Sendable {
     }
 
     public func getRemainingTime() -> Duration {
-        let deadline = self.deadline.millisSinceEpoch
-        let now = DispatchWallTime.now().millisSinceEpoch
-
-        let remaining = deadline - now
-        return .milliseconds(remaining)
+        let deadline = self.deadline
+        return deadline.duration(to: LambdaClock().now)
```
adam-fowler: Shouldn't this be `LambdaClock().now.duration(to: deadline)`

Sources/AWSLambdaRuntime/LambdaClock.swift
```diff
@@ -0,0 +1,184 @@
...
+///
+/// This clock uses `clock_gettime(CLOCK_REALTIME)` on Unix systems for
+/// high-precision wall-clock time measurement with millisecond resolution.
+///
+/// ## TimeZone Handling
+///
+/// The Lambda execution environment uses UTC as a timezone,
+/// `LambdaClock` operates in UTC and does not account for time zones.
+/// see: TZ in https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
+public struct LambdaClock: Clock {
```
adam-fowler, line 54: 👍 I did think of using c std lib functions, but couldn't be arsed working them out. This is great

sebsto: My C is a bit rusty, I got some help from Amazon Q

Sources/AWSLambdaRuntime/LambdaContext.swift
sebsto: You're correct, otherwise it would return a negative duration. I added a unit test to verify this

adam-fowler (approved):

sebsto:
Thank you !

Issue #465, opened 2025-01-15
[core] LambdaRuntimeClient do nothing when server closes the connection


sebsto:
### Expected behavior

When the server closes the connection, the client should try to reconnect and eventually shutdown if it can't.

We should implement a retry strategy to re-establish the connection (ideally with an exponential backoff) and gracefully shut down the client if it can't reconnect after n attempts.

### Actual behavior

The LambdaRunTimeClient change its internal state but does not attempt to reconnect nor exit.
See this line of code https://github.com/swift-server/swift-aws-lambda-runtime/blob/b4f016422ebd5e9c2e6aa97b8c74e924dad2f4c9/Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift#L267

### Steps to reproduce

1. checkout this branch to use the new MockServer 
https://github.com/sebsto/swift-aws-lambda-runtime/tree/sebsto/mockserver

2. Start the Mock server 

```
LOG_LEVEL=trace  swift run

2025-01-15T11:43:07+0100 info MockServer : host="127.0.0.1" maxInvocations=1 port=7000 [MockServer] Server started and listening
```

3. From another terminal, start the HelloWorld example in DEBUG mode.

```
cd Examples/HelloWorld
export AWS_LAMBDA_RUNTIME_API=127.0.0.1:7000
LAMBDA_USE_LOCAL_DEPS=../.. LOG_LEVEL=trace swift run
```

The mock server closes the connection after 2 HTTP request on 1 connection.
The runtime will send the `/next` and the `/response` requests and then it will stay up and running without retrying the connection.

Adding a `print()` statement [at line 267](https://github.com/swift-server/swift-aws-lambda-runtime/blob/b4f016422ebd5e9c2e6aa97b8c74e924dad2f4c9/Sources/AWSLambdaRuntimeCore/LambdaRuntimeClient.swift#L267) shows that the connection close even is trapped, but no action is taken.

### If possible, minimal yet complete reproducer code (or URL to code)

n/a

### What version of this project (`swift-aws-lambda-runtime`) are you using?

main

### Swift version

swift-driver version: 1.115.1 Apple Swift version 6.0.3 (swiftlang-6.0.3.1.10 clang-1600.0.30.1)
Target: arm64-apple-macosx15.0

### Amazon Linux 2 docker image version

n/a

sebsto:
[...]

PR #542, opened 2025-08-01
Performance Test the invocation loop

sebsto:
re-implement MAX_INVOCATIONS and fix shell script 
Fix https://github.com/swift-server/swift-aws-lambda-runtime/issues/377

### Motivation:

In v1, there was a script measuring the performance of the invocation loop.
Re-instate this script to allow user and develoeprs to measure the performance impact of their changes.

### Modifications:

I re-implemented MAX_INVOCATIONS, to avoid teh client looping against the Mock Server. But this time, MAX_INVOCATIONS is handled on the server, not on the client.

I slightly modified the script to work with v2 and the new MockServer.

### Result:

The script works.

This PR has a dependency on https://github.com/swift-server/swift-aws-lambda-runtime/issues/465


sebsto:
proceeding and merging without a review as this does not touch the library itself, just a mock server used for perf testing and the test script.

PR #543, opened 2025-08-03
Fix link (Response Streaming) in readme.md

natanrolnik:
When reading about the V2 changes , I noticed that the link for Response Streaming in the AWS docs was leading to a 404.

A trailing `]` character was the reason - this PR fixes it.


sebsto (approved):
Thank you Nathan for having spot and fixed this.

Issue #546, opened 2025-08-04
Lambda Streaming now support 200Mb payload


sebsto:
Check thorough the code for the old 20Mb limit 

https://aws.amazon.com/about-aws/whats-new/2025/07/aws-lambda-response-streaming-200-mb-payloads/

dsplatonov:
Hello, just make pull request for this issue https://github.com/swift-server/swift-aws-lambda-runtime/pull/548 . 
I found 20 mb mention only in documentation and fixed it. Please take a look. Thanks in advance

sebsto:
Thank you @dsplatonov 
Merged.

PR #548, opened 2025-08-05
[doc] Updated streaming response limit to 200 MB

dsplatonov:
### Motivation:
AWS Lambda response streaming now supports a default maximum response payload size of 200 MB (increased from 20 MB). This update aligns the documentation with the new AWS limits.

### Modifications:
- Updated readme.md: Changed "soft limit of 20 MB" to "default maximum response payload size of 200 MB"
- Updated Examples/Streaming/README.md: Same change as above
- Updated terminology to match AWS official documentation

### Result:
Documentation now accurately reflects the current AWS Lambda streaming response limits, enabling users to understand they can stream up to 200 MB payloads.

Reference: https://aws.amazon.com/about-aws/whats-new/2025/07/aws-lambda-response-streaming-200-mb-payloads/

### DIFF
--- a/Examples/Streaming/README.md
+++ b/Examples/Streaming/README.md
@@ -1,6 +1,6 @@
 # Streaming Lambda function
 
-You can configure your Lambda function to stream response payloads back to clients. Response streaming can benefit latency sensitive applications by improving time to first byte (TTFB) performance. This is because you can send partial responses back to the client as they become available. Additionally, you can use response streaming to build functions that return larger payloads. Response stream payloads have a soft limit of 20 MB as compared to the 6 MB limit for buffered responses. Streaming a response also means that your function doesn’t need to fit the entire response in memory. For very large responses, this can reduce the amount of memory you need to configure for your function.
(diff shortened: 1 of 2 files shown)


sebsto:
Thank you @dsplatonov !
Your change will fix https://github.com/swift-server/swift-aws-lambda-runtime/issues/546

But I think there is a bit more than the doc to check. I know the [LambdaRuntimeClient](https://github.com/swift-server/swift-aws-lambda-runtime/blob/main/Sources/AWSLambdaRuntime/LambdaRuntimeClient.swift) struct has several references to the 6Mb payload limit for non-streaming operations. We need to make sure there is no 20Mb limit hardcoded here and there.

dsplatonov:
Yea, I’ve checked all files, found 20 mb only in documentation, 6mb limit is not changed, so I didn’t change it in code:)

sebsto (approved):
LGTM. I also checked for hard-code 20Mb limit in the code and could not find any.

Issue #551, opened 2025-08-24
[core] regression - the new `LambdaClock` and `Instant` breaks HTTP headers sent by LocalServer


sebsto:
The change recently introduced by #540 (remove dependency on DispatchWallTime) introduced a regression.

The Lambda runtime client crashes because of "missing required headers" (`AWSLambdaRuntime.LambdaRuntimeError.Code.nextInvocationMissingHeaderDeadline`) when calling the Lambda Local Server for testing.

When the runtime client calls `HTTP GET /next`, it checks for presence of mandatory server headers. One of them is `Lambda-Runtime-Deadline-Ms` which is expected to be  an EPOCH number (see https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next)

The newly introduced `LambdaClock.Instant` doesn't generate the correct value, probably a lack of `CustomStringConvertible`. The object is rendered as string instead of the EPOCH value.

```
Lambda-Runtime-Aws-Request-Id: 21235745116708; Lambda-Runtime-Invoked-Function-Arn: arn:aws:lambda:us-east-1:-6648:function:custom-runtime; Lambda-Runtime-Trace-Id: Root=1-68aaeb5d-86cd1c934bb9178e8c7e6ec0;Sampled=1; Lambda-Runtime-Deadline-Ms: Instant(instant: 1756032737492); Content-Length: 26
```

`Instant` is not rendered correctly. 
```
Instant(instant: 1756032737492)
```
instead of 
```
1756032737492
```

sebsto:
Fixed by https://github.com/swift-server/swift-aws-lambda-runtime/pull/552

PR #552, opened 2025-08-24
fix: Fix deadline header with correct instant value (#551)

sebsto:
Instant's value now correctly prints as an EPOCH number

### Motivation:

A regression was introduced by https://github.com/swift-server/swift-aws-lambda-runtime/pull/540. The HTTP headers returned by `LocalServer` contained an invalid representation of the Lamba Deadline.

See https://github.com/swift-server/swift-aws-lambda-runtime/issues/551

### Modifications:

- Add `CustomStringConvertible` to `LambdaClock.Instant` to just print the `Int64` value 
- add a unit test 

### Result:

The runtime works correctly with the new `LambdaClock`


sebsto:
As all CI check pass, and this is a trivial bug fix, I will merge without further review.

Issue #556, opened 2025-09-01
Can we allow customisation of local server port


adam-fowler:
### Expected behavior

Add environment variable `LOCAL_LAMBDA_PORT` to customise port local server binds to.

### Actual behavior

Airplay can use port 7000

### Steps to reproduce

Run local lambda on Mac with setting `General -> AirDrop & Handoff -> AirPlay receiver` enabled
Check server response using `curl -I localhost:7000`

### What version of this project (`swift-aws-lambda-runtime`) are you using?

2.0.0-beta.2

### Swift version

Apple Swift version 6.1 (swift-6.1-RELEASE)
Target: arm64-apple-macosx15.0
Darwin Adams-Mac-mini.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6041 arm64

sebsto:
I know TCP port conflicts with Airplay.

However, it never really caused problems. Somehow the runtime binds TCP port 7000 even when AirPlay is configured. I had problems when the server stops and the runtime continues trying polling events, which causes a `nextInvocationMissingHeaders` error (because Airplay doesn't send expected headers)

I'm curious to learn in which context it prevents running.

Anyway, I think it's better to address this once for all and allow users to select their port.

Can you please review https://github.com/swift-server/swift-aws-lambda-runtime/pull/557 ?

sebsto:
fixed by https://github.com/swift-server/swift-aws-lambda-runtime/pull/557

PR #557, opened 2025-09-01
add support for LOCAL_LAMBDA_PORT

sebsto:
Allows users to define on which port the Local server listens to, using the `LOCAL_LAMBDA_PORT` environment variable.

While being at it, I also added `LOCAL_LAMBDA_HOST` if the user wants to bind on a specific IP address.

I renamed `LOCAL_LAMBDA_SERVER_INVOCATION_ENDPOINT` to `LOCAL_LAMBDA_INVOCATION_ENDPOINT` for consistency.

### Motivation:

Addresses https://github.com/swift-server/swift-aws-lambda-runtime/issues/556

### Modifications:

- When run outside of the Lambda execution environment, check for the value of `LOCAL_LAMBDA_PORT` and passes it down to the Lambda HTTP Local Server and runtime client.

- Add a unit test 

### Result:

```
LAMBDA_USE_LOCAL_DEPS=../.. LOCAL_LAMBDA_PORT=8888 swift run                  

2025-09-01T21:55:22+0200 info LambdaRuntime: host="127.0.0.1" port=8888 [AWSLambdaRuntime] Server started and listening
```

### DIFF
--- a/Tests/AWSLambdaRuntimeTests/LambdaLocalServerTests.swift
+++ b/Tests/AWSLambdaRuntimeTests/LambdaLocalServerTests.swift
@@ -1,0 +1,95 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Logging
+import NIOCore
+import NIOPosix
+import Testing
+
+@testable import AWSLambdaRuntime
+
+@Suite("Lambda Local Server Tests")
+struct LambdaLocalServerTests {
+
+    @Test("Local server respects LOCAL_LAMBDA_PORT environment variable")
+    func testLocalServerCustomPort() async throws {
+        let customPort = 8080
+
+        // Set environment variable
+        setenv("LOCAL_LAMBDA_PORT", "\(customPort)", 1)
+        defer { unsetenv("LOCAL_LAMBDA_PORT") }
+
+        let result = try? await withThrowingTaskGroup(of: Bool.self) { group in
+
+            // start a local lambda + local server on custom port
+            group.addTask {
+                // Create a simple handler
+                struct TestHandler: StreamingLambdaHandler {
+                    func handle(
+                        _ event: ByteBuffer,
+                        responseWriter: some LambdaResponseStreamWriter,
+                        context: LambdaContext
+                    ) async throws {
+                        try await responseWriter.write(ByteBuffer(string: "test"))
+                        try await responseWriter.finish()
+                    }
+                }
+
+                // create the Lambda Runtime
+                let runtime = LambdaRuntime(
+                    handler: TestHandler(),
+                    logger: Logger(label: "test", factory: { _ in SwiftLogNoOpLogHandler() })
+                )
+
+                // Start runtime
+                try await runtime._run()
+
+                // we reach this line when the group is cancelled
+                return false
+            }
+
+            // start a client to check if tsomething responds on the custom port
+            group.addTask {
+                // Give server time to start
+                try await Task.sleep(for: .milliseconds(100))
+
+                // Verify server is listening on custom port
+                return try await isPortResponding(host: "127.0.0.1", port: customPort)
+            }
+
+            let first = try await group.next()
+            group.cancelAll()
+            return first ?? false
+
+        }
+
+        #expect(result == true)
+    }
+
+    private func isPortResponding(host: String, port: Int) async throws -> Bool {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+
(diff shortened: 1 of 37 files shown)


readme.md
```diff
@@ -464,16 +464,22 @@ curl -v --header "Content-Type:\ application/json" --data @events/create-session
...
+
+- Set `LOCAL_LAMBDA_HOST` to configure the local server to listen on a different TCP address.
+- Set `LOCAL_LAMBDA_PORT` to configure the local server to listen on a different TCP port.
+- Set `LOCAL_LAMBDA_INVOCATION_ENDPOINT` to force the local server to listen on a different endpoint.
 
 Example:
 
 ```sh
-LOCAL_LAMBDA_SERVER_INVOCATION_ENDPOINT=/2015-03-31/functions/function/invocations swift run
+LOCAL_LAMBDA_PORT=8080 LOCAL_LAMBDA_INVOCATION_ENDPOINT=/2015-03-31/functions/function/invocations swift run
```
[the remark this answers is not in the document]
sebsto: This is incorrect

Tests/AWSLambdaRuntimeTests/LambdaLocalServerTests.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        let bootstrap = ClientBootstrap(group: group)
+
+        do {
+            let channel = try await bootstrap.connect(host: host, port: port).get()
+            try await channel.close().get()
+            try await group.shutdownGracefully()
+            return true
+        } catch {
+            try await group.shutdownGracefully()
```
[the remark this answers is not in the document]
sebsto: This is incorrect `defer { try? await group.shutdownGracefully() }` doesn't compile (defer doesn't allow async code)

Tests/AWSLambdaRuntimeTests/LambdaLocalServerTests.swift
```diff
@@ -0,0 +1,95 @@
...
+@Suite("Lambda Local Server Tests")
+struct LambdaLocalServerTests {
+
+    @Test("Local server respects LOCAL_LAMBDA_PORT environment variable")
+    func testLocalServerCustomPort() async throws {
+        let customPort = 8080
+
+        // Set environment variable
+        setenv("LOCAL_LAMBDA_PORT", "\(customPort)", 1)
+        defer { unsetenv("LOCAL_LAMBDA_PORT") }
```
adam-fowler, line 30: Given these tests are running in parallel, there is a chance the LOCAL_LAMBDA_PORT environment var setting will affect other tests.

adam-fowler (commented):
One minor issue in the tests. I think you are going to have to group together all the tests that use LambdaRuntime and mark them with the `.serialized` test trait so they don't run on parallel

Tests/AWSLambdaRuntimeTests/LambdaLocalServerTests.swift
sebsto: @adam-fowler good point. I aggressively added `.serialized` on all tests using the server or the runtime client

adam-fowler: I think that might only run tests serially within each suite so two suites can still run in parallel, even though the tests internally run serially. Might be worth asking swift-testing folks

sebsto: @adam-fowler During the tests, we let the MockServer bind on any available port (port = 0) and share its actual port number with the runtime client. There is no risk of conflict. 

See https://github.com/swift-server/swift-aws-lambda-runtime/blob/ec28c9669684bbbe22e447b24c201add4480e8b0/Tests/AWSLambdaRuntimeTests/MockLambdaServer.swift#L28

I kept the `.serialized` only on the new test in case someone adds another test to this suite later on.

adam-fowler: The problem isn't the fact you might have multiple servers binding to the same address (although that is a problem and am surprised it hasn't reared it head before). The problem is the environment variable is being set when other tests could be running and could affect those tests. 

We can leave as is, but I think we will get the occasional test failure as multiple tests inherit the environment variable and bind to the same address.

sebsto: @adam-fowler 
>am surprised it hasn't reared it head before 

In general, the unit tests don't use the local server. They use a [MockServer](https://github.com/swift-server/swift-aws-lambda-runtime/blob/ec28c9669684bbbe22e447b24c201add4480e8b0/Tests/AWSLambdaRuntimeTests/MockLambdaServer.swift#L28) that uses a random port number to avoid conflicts.

> I think we will get the occasional test failure as multiple tests inherit the environment variable and bind to the same address

There are three groups of tests that use the real server : 
- `LambdaRuntimeTests.swift`
- `LambdaRuntime+ServiceLifecycle.swift`
- `LambdaLocalServerTests.swift` (this PR) 

I will add a commit to group them in the same `.serialized` suite.

sebsto: @adam-fowler wdyt ?  https://github.com/swift-server/swift-aws-lambda-runtime/pull/557/commits/2216a5cf9bd2a317832ebef29ce8f3bd99c73118

adam-fowler (approved):
Looks good

sebsto:
fixes https://github.com/swift-server/swift-aws-lambda-runtime/issues/556

PR #564, opened 2025-09-23
Update toolchain and doc for 6.2

sebsto:
In preparation for the 2.0.0 GA release,

- Update `.swift-version`, `Package.swift` and all examples' `package.swift` to Swift 6.2
- Update all references to `2.0.0-beta.3` to `2.0.0`. This  includes the doc and readme, but also the dependencies in the examples `Package.swift`.  This will temporary break the build of the examples, until we tag v2.0.0.  Note the CI will not be affected as its consumes the local version of the library
- [CI] Use Swift-6.2-noble for all testing tasks
- Reinstate the script to generate the contributors list and update the list


.github/workflows/integration_tests.yml
```diff
@@ -123,14 +127,37 @@ jobs:
...
     if: ${{ inputs.check_foundation_enabled }}
     runs-on: ubuntu-latest
+    container:
+      image: ${{ inputs.matrix_linux_swift_container_image }}
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-        with:
-          persist-credentials: false
+      # GitHub checkout action has a dep on NodeJS 20 which is not running on Amazonlinux2
```
0xTim, line 133: What version of node is installed? The v2 action requires node 24 or higher

Examples/HelloWorld/Package.swift
```diff
@@ -1,4 +1,9 @@
-// swift-tools-version:6.0
+// swift-tools-version:6.1
+// This example has to be in Swift 6.1 because it is used in the test archive plugin CI job
+// That job runs on GitHub's ubuntu-latest environment that only supports Swift 6.1
+// https://github.com/actions/runner-images?tab=readme-ov-file
+// https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
+// We can update to Swift 6.2 when GitHUb hosts will have Swift 6.2
```
0xTim, line 6: Tracking here https://github.com/actions/runner-images/pull/13068

.github/workflows/integration_tests.yml
sebsto: Amazon Linux doesn't have Node 24, this is the "normal" dance when using Amazon Linux 2.
I hope to get an Amazon Linux 2023 Swift image soon (at least a nightly build), I will upgrade to ALI2023 when available.

0xTim: Ah I got mixed up between AL2 and AL2023 😅

PR #569, opened 2025-09-26
[Example] Add APIGatewayV1 example

everyplace:
This PR adds an APIGateway V1 example, which differs slightly from the existing [V2 example](https://github.com/swift-server/swift-aws-lambda-runtime/tree/main/Examples/APIGateway).

### Motivation:

While APIGatewayV2 has existed for a while, there are still some scenarios where V1 is preferable (see [Choose between REST APIs and HTTP APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html) for more details). This PR adds a working example for the `Api` endpoint type.

### Modifications:

Sources/main.swift
- Swapped APIGatewayRequestV2 and APIGatewayResponseV2 for their unversioned counterparts (aka V1).

template.yaml
- Changed the Events name and corresponding `Type: Api` values.
- Added necessary additional properties to define a `GET` endpoint for an `Api` endpoint.

Readme.md
- Updated the request and response example outputs.

### Result:

Now, an APIGateway and APIGatewayV1 example exist. 

> [!NOTE]
> It is the author's opinion that this new example should technically be called `APIGateway` and the existing one should be renamed to `APIGatewayV2` to keep consistent with the request and response type naming conventions, but this is omitted from this request to be less disruptive.

### DIFF
--- a/Examples/APIGatewayV1/Sources/main.swift
+++ b/Examples/APIGatewayV1/Sources/main.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftAWSLambdaRuntime open source project
 //
-// Copyright (c) 2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Copyright (c) 2025 Apple Inc. and the SwiftAWSLambdaRuntime project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
--- a/Examples/APIGatewayV1/Sources/main.swift
+++ b/Examples/APIGatewayV1/Sources/main.swift
@@ -1,0 +1,30 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftAWSLambdaRuntime project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import AWSLambdaEvents
+import AWSLambdaRuntime
+
+let runtime = LambdaRuntime {
+    (event: APIGatewayRequest, context: LambdaContext) -> APIGatewayResponse in
+
+    var header = HTTPHeaders()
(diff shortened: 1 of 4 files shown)


Examples/APIGatewayV1/Sources/main.swift
```diff
@@ -0,0 +1,30 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftAWSLambdaRuntime open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftAWSLambdaRuntime project authors
```
sebsto: I would update this to 2025

Examples/APIGatewayV1/template.yaml
```diff
@@ -0,0 +1,34 @@
...
+Resources:
+  # Lambda function
+  APIGatewayLambda:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: .build/plugins/AWSLambdaPackager/outputs/AWSLambdaPackager/APIGatewayLambda/APIGatewayLambda.zip
+      Timeout: 60
+      Handler: swift.bootstrap  # ignored by the Swift runtime
+      Runtime: provided.al2
+      MemorySize: 512
```
sebsto: I know it's coming from the example you copied, but I would use 128Mb here. This is enough to run the function and it reduces costs for customers.
I will add a GH issue to remember doing the same on all SAM examples.

sebsto: https://github.com/swift-server/swift-aws-lambda-runtime/issues/570

sebsto:
Thank you for this PR.
Congrats for submitting a PR that passes all the tests at first attempt  :-) 

I suggested two minor changes in the review.

Can you also add this test to the CI ([this line](https://github.com/swift-server/swift-aws-lambda-runtime/blob/a1ab8df708285b34a0a63335ac7802dfd3b4fe04/.github/workflows/pull_request.yml#L39))?

I like your suggestion to rename to existing APIGateway example to `...V2`. Let's keep this one as `...V1`, so, at the end, we will have `Examples/APIGatewayV1` and `Examples/APIGatewayV2` and `Examples/APIGatewayV2+LambdaAuthorizer`

If you have time, can you submit a PR for this change too (don't forget to change the CI too :-) )

sebsto (changes requested):
LGTM, just two tiny changes

sebsto (approved):
Thank you !

Issue #580, opened 2025-10-06
Implement support for throwing errors that will result in different ErrorType's


benrosen78:
I make use of AWS Step Functions to orchestrate my Lambdas, in some cases routing Retry/Catch logic based on specific types of errors that Lambdas throw. AWS documents this behavior [here](https://docs.aws.amazon.com/step-functions/latest/dg/tutorial-handling-error-conditions.html) with the following example in NodeJS:
```
export const handler = async () => {
    function CustomError(message) {
        this.name = 'CustomError';
        this.message = message;
    }
    CustomError.prototype = new Error();

    throw new CustomError('This is a custom error!');
};
```

Similarly, in Python, the following can be done:
```
class CustomError(Exception):
    pass

def lambda_handler(event, context):
    raise CustomError("Test Error")
 ```

This will result in the following output:
```
{
  "errorMessage": "Test Error",
  "errorType": "CustomError"
}
```

The AWS Step Function allows Catch/Retry behaviors based on the `errorType` that is outputted in the Lambda. 

In Swift, the `errorType` is hardcoded [here](https://github.com/swift-server/swift-aws-lambda-runtime/blob/eeb354986bea30e674b50a3f4237b294bfb47c55/Sources/AWSLambdaRuntime/LambdaRuntimeClient%2BChannelHandler.swift#L337) to output "FunctionError". This makes it impossible to Catch/Retry in a step function based off of the type of error that was thrown. It is possible to accomplish similar goals in step function by using string comparisons, but this would increase state transitions, and therefore cost.

To solve this issue for my own case, I forked v1 and added the capability in [this commit](https://github.com/SongShift/swift-aws-lambda-runtime/commit/a12974f4cdaf516e54212c028e4016d27b6807f3). This fork does accomplish the desired goal, but it is most likely not the most elegant way to do so. I would like to start a discussion on how this behavior could be adopted in the runtime.

sebsto:
Thank you @benrosen78 for having taken the time to submit this.  I will have a look asap

PR #587, opened 2025-10-14
Make Lambda handler output for failed requests output errorType as the type of an error rather than FunctionError constant

benrosen78:
### Motivation:

Fix for Issue [#580](https://github.com/swift-server/swift-aws-lambda-runtime/issues/580), by making it so that the `errorType` in failed requests will be the type of the error entity, rather than a hardcoded string of `FunctionError`. This allows orchestration within step functions that perform retry/catch logic based on different error output types.

### Modifications:

At a high level, the issue is that swift-aws-lambda-runtime, when an error is thrown, outputs the errorType as hardcoded to FunctionError. You can see that [here](https://github.com/swift-server/swift-aws-lambda-runtime/blob/main/Sources/AWSLambdaRuntime/LambdaRuntimeClient%2BChannelHandler.swift#L337):

```
let errorResponse = ErrorResponse(errorType: Consts.functionError, errorMessage: "\(error)")
```

This PR changes this for all cases to output the type of the error, rather than the hardcoded string:
```
let errorResponse = ErrorResponse(errorType: "\(type(of: error))", errorMessage: "\(error)")
```
Now, I will show 2 examples with this solution:
```
let runtime = LambdaRuntime {
    (event: Input, context: LambdaContext) in

    enum MyTestErrorType: Error {
        case testError
    }
    
    throw MyTestErrorType.testError
}

// outputs {"errorType":"MyTestErrorType","errorMessage":"testError"}
```
```
let dynamoDB: DynamoDB = DynamoDB(client: .init())

let runtime = LambdaRuntime {
    (event: Input, context: LambdaContext) in

    let _ = try await dynamoDB.putItem(DynamoDB.PutItemInput(item: [:], tableName: ""))
    
    return Output()
}

// outputs {"errorType":"AWSClientError","errorMessage":"ValidationError: Length of PutItemInput.tableName (0) is less than minimum allowed value 1."}
```


Tests/AWSLambdaRuntimeTests/LambdaRuntimeClientTests.swift
```diff
@@ -330,4 +330,78 @@ struct LambdaRuntimeClientTests {
             }
         }
     }
+
+    @Test(
```
sebsto, line 334: Great !

PR #581, opened 2025-10-09
Add LambdaRuntime initializer with LambdaHandler directly with Codable support

natanrolnik:
_Add convenience initializer for `LambdaRuntime` to accept `LambdaHandler` instances directly_

### Motivation:

When using the Swift AWS Lambda Runtime with custom handler types that conform to `LambdaHandler`, developers previously had two options to initialize `LambdaRuntime`:

1. Manually wrap their handler with `LambdaCodableAdapter` and `LambdaHandlerAdapter`:
   ```swift
   let lambdaHandler = MyHandler()
   let handler = LambdaCodableAdapter(
       encoder: JSONEncoder(),
       decoder: JSONDecoder(),
       handler: LambdaHandlerAdapter(handler: lambdaHandler)
   )
   let runtime = LambdaRuntime(handler: handler)
   ```

2. Use a closure-based initializer that indirectly calls the handler:
   ```swift
   let lambdaHandler = MyHandler()
   let runtime = LambdaRuntime { event, context in
       try await lambdaHandler.handle(event, context: context)
   }
   ```

Both approaches are verbose and don't provide a clean, ergonomic API for the common case of initializing `LambdaRuntime` with a custom `LambdaHandler` instance. The closure approach also creates an unnecessary indirection layer, wrapping the handler in a `ClosureHandler` before adapting it, or using `LambdaCodableAdapter`.

### Modifications:

Added a new convenience initializer to `LambdaRuntime` in `Lambda+JSON.swift` that accepts a `LambdaHandler` instance directly:

```swift
public convenience init<Event: Decodable, Output, LHandler: LambdaHandler>(
    decoder: JSONDecoder = JSONDecoder(),
    encoder: JSONEncoder = JSONEncoder(),
    logger: Logger = Logger(label: "LambdaRuntime"),
    lambdaHandler: sending LHandler
)
where
    Handler == LambdaCodableAdapter<
        LambdaHandlerAdapter<Event, Output, LHandler>,
        Event,
        Output,
        LambdaJSONEventDecoder,
        LambdaJSONOutputEncoder<Output>
    >,
    LHandler.Event == Event,
    LHandler.Output == Output
```

This initializer handles the wrapping of the `LambdaHandler` with the necessary adapters internally, matching the pattern already established for closure-based handlers.

### Result:

Developers can now initialize `LambdaRuntime` with a `LambdaHandler` instance using a clean, direct API:

```swift
let lambdaHandler = MyHandler()
[...]

PR #589, opened 2025-10-16
Add initializer with LambdaHandler directly with Decodable event and Void output

natanrolnik:
_Add convenience initializer for `LambdaRuntime` to accept `LambdaHandler` instances with `Void` output_

### Motivation:

Following PR #581, which added convenience initializers for `LambdaRuntime` to accept `LambdaHandler` instances with `Encodable` outputs, there was still a gap for handlers that return `Void`. This is useful, for example, for AWS event sources like SQS, where the `AWSLambdaEvents` package provides `SQSEvent` but there's no corresponding output type - handlers simply process messages and return `Void`.

Without this initializer, developers had to manually wrap their handlers with `LambdaCodableAdapter` and `LambdaHandlerAdapter`, which was verbose and inconsistent with the new API (similar to what is described in #581)

### Modifications:

1. Added a new `init(decoder:handler:)` convenience initializer to `LambdaCodableAdapter` in `Lambda+JSON.swift` that accepts a `JSONDecoder` for handlers with `Void` output.

2. Added a new convenience initializer to `LambdaRuntime` that accepts a `LambdaHandler` instance with `Void` output directly, handling the wrapping internally.

3. Updated documentation comments to distinguish between handlers with `Void` and `Encodable` outputs.

### Result:

Developers can now initialize `LambdaRuntime` with handlers that return `Void` using a clean, direct API. This is especially useful for event sources like SQS:

```swift
import AWSLambdaEvents

struct MySQSHandler: LambdaHandler {
    func handle(_ event: SQSEvent, context: LambdaContext) async throws {
        // Process SQS messages
    }
}

let runtime = LambdaRuntime(lambdaHandler: MySQSHandler())
```

This provides API completeness, matching the convenience initializers for handlers with `Encodable` outputs, and delivers better ergonomics for common serverless patterns.

### DIFF
--- a/Sources/AWSLambdaRuntime/FoundationSupport/Lambda+JSON.swift
+++ b/Sources/AWSLambdaRuntime/FoundationSupport/Lambda+JSON.swift
@@ -83,7 +83,28 @@ extension LambdaCodableAdapter {
             handler: handler
         )
     }
+
+    /// Initializes an instance given a decoder, and a handler with a `Void` output.
+    ///   - Parameters:
+    ///   - decoder: The decoder object that will be used to decode the received `ByteBuffer` event into the generic `Event` type served to the `handler`. By default, a JSONDecoder is used.
+    ///   - handler: The handler object.
+    public init(
+        decoder: JSONDecoder = JSONDecoder(),
+        handler: sending Handler
+    )
+    where
+        Output == Void,
+        Handler.Output == Void,
+        Decoder == LambdaJSONEventDecoder,
+        Encoder == VoidEncoder
+    {
+        self.init(
+            decoder: LambdaJSONEventDecoder(decoder),
+            handler: handler
+        )
+    }
 }
+
 @available(LambdaSwift 2.0, *)
 extension LambdaResponseStreamWriter {
     /// Writes the HTTP status code and headers to the response stream.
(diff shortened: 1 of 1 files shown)


Sources/AWSLambdaRuntime/FoundationSupport/Lambda+JSON.swift
```diff
@@ -83,7 +83,28 @@ extension LambdaCodableAdapter {
             handler: handler
         )
     }
+
+    /// Initializes an instance given a decoder, and a handler with a `Void` output.
+    ///   - Parameters:
+    ///   - decoder: The decoder object that will be used to decode the received `ByteBuffer` event into the generic `Event` type served to the `handler`. By default, a JSONDecoder is used.
+    ///   - handler: The handler object.
+    public init(
```
sebsto, line 91: Just a question. Should this be a convenience init ?

natanrolnik: I mostly copied the init a few lines above this one and changed the generics definitions. When asking Claude Sonnet, I got this answer:

> This initializer doesn't need to be a convenience initializer because:
> * It's calling a designated initializer (the one at line 114 in Lambda+Codable.swift) - it's not delegating to another initializer within the same type using self.init
> * Extensions can only have convenience initializers in classes, but LambdaCodableAdapter is a struct, and struct initializers in extensions are automatically treated as convenience-like (they must delegate to an existing initializer)

sebsto: That makes sense.  Thank you

sebsto (approved):
LGTM, thank you for improving the API

PR #593, opened 2025-10-20
fix ci : check libFoundation on HelloWorld example

sebsto:
Fix errors in the CI
The script that checks the presence of `libFoundation` in the binary started to fail. 
I can't think about a recent change that would cause this.

This PR change the test script to use the `HelloWorld` example instead of `APIGAtewayV2`

### DIFF
--- a/.github/workflows/scripts/check-link-foundation.sh
+++ b/.github/workflows/scripts/check-link-foundation.sh
@@ -17,23 +17,23 @@ log() { printf -- "** %s\n" "$*" >&2; }
 error() { printf -- "** ERROR: %s\n" "$*" >&2; }
 fatal() { error "$@"; exit 1; }
 
-EXAMPLE=APIGatewayV2
+EXAMPLE=HelloWorld
 OUTPUT_DIR=.build/release
-OUTPUT_FILE=${OUTPUT_DIR}/APIGatewayLambda
+OUTPUT_FILE=${OUTPUT_DIR}/MyLambda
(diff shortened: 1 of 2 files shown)


sebsto:
The `HelloWorld` example doesn't link with Foundation, which makes me think the dependency might be imported by `swift-lambda-events` library, although I can't find any trace of `libFoundation` in its binary.

```
bash-4.2# nm Examples/APIGatewayV2/.build/release/APIGatewayLambda | grep -i NSURL
                 U $s10Foundation5NSURLC6string10relativeToACSgSS_0A10Essentials3URLVSgtcfc
                 U $s10Foundation5NSURLCMa
bash-4.2# nm Examples/APIGatewayV2/.build/release/APIGatewayLambda | grep -i foundation | grep -v -i essentials
                 U $s10Foundation5NSURLCMa
```

I don't know if this is the CI test that never worked and start to work and produce these error or if something has changed.

I can't find a recent code change that caused this, neither in the runtime or on the event library. I even tried the event library v1, before your change for  Codable)

When testing for `libFoundation` on the `helloworld `example, it is negative. I think the link is caused by something in the event library.
But I can't find anything on the event library

```
bash-4.2# swift build -c release
Building for production...
[7/7] Compiling AWSLambdaEvents ALB.swift
Build complete! (13.20s)
bash-4.2# nm .build/release/AWSLambdaEvents.build/*.o  | grep -i foundation | grep -v -i essentials
bash-4.2# nm .build/release/AWSLambdaEvents.build/*.o  | grep -i nsurl                             
bash-4.2#
```

sebsto:
Root cause is https://github.com/apple/swift-http-types/issues/107

PR #598, opened 2025-10-30
Add Multi-Source API Example

sebsto:
This PR adds a new example demonstrating how to build a Lambda function that handles requests from multiple sources (Application Load Balancer and API Gateway V2) using a single handler.

### What's New

**New Example: `Examples/MultiSourceAPI`**

A Lambda function that:
- Implements `StreamingLambdaHandler` to accept raw `ByteBuffer` events
- Dynamically decodes events as either `ALBTargetGroupRequest` or `APIGatewayV2Request`
- Returns appropriate responses based on the detected event source
- Demonstrates handling multiple AWS service integrations with a single function

### Key Features

- **Type-safe event detection**: Uses Swift's `Decodable` to identify the event source at runtime
- **Streaming response**: Implements `StreamingLambdaHandler` for efficient response handling
- **Complete deployment**: Includes SAM template with both ALB and API Gateway V2 infrastructure
- **Production-ready**: Full VPC setup with subnets, security groups, and load balancer configuration

### Files Added

- `Examples/MultiSourceAPI/Sources/main.swift` - Lambda handler implementation
- `Examples/MultiSourceAPI/Package.swift` - Swift package configuration
- `Examples/MultiSourceAPI/template.yaml` - SAM deployment template with ALB and API Gateway V2
- `Examples/MultiSourceAPI/README.md` - Documentation with build, deploy, and test instructions
- Updated CI to include the new example

### Use Case

This pattern is useful when you want to:
- Expose the same Lambda function through multiple AWS services
- Reduce code duplication by handling similar requests from different sources
- Maintain a single codebase for multi-channel APIs

### DIFF
--- a/Examples/MultiSourceAPI/Package.swift
+++ b/Examples/MultiSourceAPI/Package.swift
@@ -34,7 +34,11 @@ if let localDepsPath = Context.environment["LAMBDA_USE_LOCAL_DEPS"],
 {
     let indexToRemove = package.dependencies.firstIndex { dependency in
         switch dependency.kind {
-        case .sourceControl(name: _, location: "https://github.com/awslabs/swift-aws-lambda-runtime.git", requirement: _):
+        case .sourceControl(
+            name: _,
+            location: "https://github.com/awslabs/swift-aws-lambda-runtime.git",
+            requirement: _
+        ):
             return true
         default:
             return false


Examples/MultiSourceAPI/Package.swift
```diff
@@ -0,0 +1,53 @@
+// swift-tools-version:6.2
+
+import PackageDescription
+
+// needed for CI to test the local version of the library
+import struct Foundation.URL
+
+let package = Package(
+    name: "MultiSourceAPI",
```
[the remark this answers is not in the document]
sebsto: will not fix

Examples/MultiSourceAPI/Sources/main.swift
```diff
@@ -0,0 +1,66 @@
+import AWSLambdaEvents
+import AWSLambdaRuntime
+import HTTPTypes
```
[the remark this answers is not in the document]
sebsto: fixed

Examples/MultiSourceAPI/Sources/main.swift
```diff
@@ -0,0 +1,66 @@
...
+import NIOCore
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
+struct MultiSourceHandler: StreamingLambdaHandler {
```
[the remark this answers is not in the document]
sebsto: fixed

Examples/MultiSourceAPI/Package.swift
```diff
@@ -0,0 +1,53 @@
...
+{
+    let indexToRemove = package.dependencies.firstIndex { dependency in
+        if case .sourceControl(
+            name: _,
+            location: "https://github.com/awslabs/swift-aws-lambda-runtime.git",
+            requirement: _
+        ) = dependency.kind {
+            return true
+        }
+        return false
```
jbelkins: nit: consider converting the `if` above to a `switch` for clarity

jbelkins:
```suggestion
        switch dependency.kind {
        case .sourceControl(name: _, location: "https://github.com/awslabs/swift-aws-lambda-runtime.git", requirement: _):
            return true
        default:
            return false
        }
```

Examples/MultiSourceAPI/Package.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    if let indexToRemove {
+        package.dependencies.remove(at: indexToRemove)
+    }
+
+    print("[INFO] Compiling against swift-aws-lambda-runtime located at \(localDepsPath)")
+    package.dependencies += [
+        .package(name: "swift-aws-lambda-runtime", path: localDepsPath)
+    ]
```
jbelkins, line 54: Does modifying the `package` after it is set raise Swift 6 language mode warnings?

If so, rearrange this logic to do this work when the package is initialized.

.github/workflows/pull_request.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       # We pass the list of examples here, but we can't pass an array as argument
       # Instead, we pass a String with a valid JSON array.
       # The workaround is mentioned here https://github.com/orgs/community/discussions/11692
-      examples: "[ 'APIGatewayV1', 'APIGatewayV2', 'APIGatewayV2+LambdaAuthorizer', 'BackgroundTasks', 'HelloJSON', 'HelloWorld', 'HelloWorldNoTraits', 'HummingbirdLambda', 'ResourcesPackaging', 'S3EventNotifier', 'S3_AWSSDK', 'S3_Soto', 'Streaming', 'Streaming+Codable', 'ServiceLifecycle+Postgres', 'Testing', 'Tutorial' ]"
+      examples: "[ 'APIGatewayV1', 'APIGatewayV2', 'APIGatewayV2+LambdaAuthorizer', 'BackgroundTasks', 'HelloJSON', 'HelloWorld', 'HelloWorldNoTraits', 'HummingbirdLambda', 'MultiSourceAPI', 'ResourcesPackaging', 'S3EventNotifier', 'S3_AWSSDK', 'S3_Soto', 'Streaming', 'Streaming+Codable', 'ServiceLifecycle+Postgres', 'Testing', 'Tutorial' ]"
```
jbelkins, line 42: Thanks for the comment above, until I read it I was going to suggest that this should be a yaml array

Examples/MultiSourceAPI/Package.swift
```diff
@@ -0,0 +1,53 @@
...
+        ) = dependency.kind {
+            return true
+        }
+        return false
+    }
+    if let indexToRemove {
+        package.dependencies.remove(at: indexToRemove)
+    }
+
+    print("[INFO] Compiling against swift-aws-lambda-runtime located at \(localDepsPath)")
```
jbelkins, line 51: Was this print meant for debugging?  Not sure a print during package resolution shows up anywhere you can see

Examples/MultiSourceAPI/Package.swift
sebsto: I don't think so. We're using the same mechanism for the library itself and for the all examples.

Examples/MultiSourceAPI/Package.swift
sebsto: Yes, it shows on `stdio`. During local development of the library or the example itself, it allows to verify that we picked up the local version of the code and not the released version.

PR #639, opened 2026-02-12
Fix for swift test : _Concurrency/Executor.swift:357: Fatal error

sebsto:
Fix the issue described at https://github.com/awslabs/swift-aws-lambda-runtime/issues/640

Here is the proposed fix:

I added a new function `assumeIsolatedOnEventLoop` — a nonisolated method that:
1. Calls `self.eventLoop.preconditionInEventLoop()` to verify we're on the correct event loop (NIO's own thread-identity check, which always works)
2. Uses `unsafeBitCast` to strip the isolated annotation, the same pattern NIO uses internally and that I found on the Swift Forums.

See: https://github.com/swiftlang/swift/blob/main/stdlib/public/Concurrency/ExecutorAssertions.swift#L348

See: https://forums.swift.org/t/actor-assumeisolated-erroneously-crashes-when-using-a-dispatch-queue-as-the-underlying-executor/72434/3


sebsto:
@adam-fowler if time permits, can you have a look at this and tell me if it makes sense? I'd like to check if my understanding is correct and if the fix, borrowed from NIO's project, is safe. Thank you

PR #641, opened 2026-02-12
Fix crash on Linux when testing the archive plugin

sebsto:
Fix a rare crash in CI on the archive plugin.

The crash is a race condition in `PluginUtils.swift`'s execute method. The code set a `terminationHandler` on the `Process` that called `readToEnd()` on the pipe, while simultaneously the `readabilityHandler` could still be firing on the same pipe's file handle. On Linux (x86_64, Ubuntu 24.04 in CI), this race corrupts memory during Swift runtime metadata resolution (`swift_conformsToProtocol, _swift_getGenericMetadata`), which manifests as the `SIGSEGV` we're seeing in `_dispatch_event_loop_drain`.

The fix:

I removed the `terminationHandler` entirely. Since `waitUntilExit()` is already called synchronously, we know the process is done.
After `waitUntilExit()`, we set `readabilityHandler = nil` to stop the async reads, then do one final `readToEnd()` on the output queue to drain any remaining data.
This eliminates the race between the readability handler and the termination handler competing over the same file handle.

### DIFF
--- a/Sources/AWSLambdaRuntime/HTTPClient/LambdaRuntimeClient.swift
+++ b/Sources/AWSLambdaRuntime/HTTPClient/LambdaRuntimeClient.swift
@@ -133,6 +133,32 @@ final actor LambdaRuntimeClient: LambdaRuntimeClientProtocol {
         self.logger = logger
     }
 
+    /// Assume that the current context is isolated to this actor's event loop and execute the closure.
+    ///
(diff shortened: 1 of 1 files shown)


sebsto:
@0xTim When time permits, can you check this change on the plugin ? It fixes a race condition that causes the plugin to crash from time to time.

PR #643, opened 2026-02-16
fix: Local Server should pass HTTP headers down to the Lambda Runtime

manojmahapatra:
## Issue \#

https://github.com/awslabs/swift-aws-lambda-runtime/issues/607

## Description of changes

The local HTTP server was not forwarding user‑provided headers to the runtime’s response. It passes all headers through to the runtime. This it makes local behavior match the Lambda runtime API contract and allows developers to opt into metadata by sending the appropriate runtime headers.

N/A

## Conventional Commits

### DIFF
--- a/Plugins/Documentation.docc/Proposals/0001-v2-plugins.md
+++ b/Plugins/Documentation.docc/Proposals/0001-v2-plugins.md
@@ -11,7 +11,7 @@ This document describes a proposal for the v2 plugins for `swift-aws-lambda-runt
 Versions:
 
 * v1 (2024-12-25): Initial version
-* v2 (2025-03-13): 
+* v2 (2025-03-13):
 - Include [comments from the community](https://forums.swift.org/t/lambda-plugins-for-v2/76859).
 - [init] Add the templates for `main.swift`
 - [build] Add the section **Cross-compiling options**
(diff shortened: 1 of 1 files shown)


sebsto:
Thank you @manojmahapatra, this is a really good first contribution. I added a tiny suggestion in the code.

May I suggest to also update the `Examples/MultiTenant`

1/ add an example that doesn't depend on the API Gateway. 
2/ update the README file and add a section about Local testing?
We need to show user it's actually possible to make an invocation like this 

(the APIGateway case is already covered, users can list the forwarded headers in the payload) 

```
  # Test multi-tenant function locally
  curl -X POST http://127.0.0.1:7000/invoke \
    -H "Content-Type: application/json" \
    -H "Lambda-Runtime-Aws-Tenant-Id: tenant-123" \
    -d '{"message" :  "hello"}'
```

Sources/AWSLambdaRuntime/HTTPServer/Lambda+LocalServer.swift
```diff
@@ -383,7 +383,10 @@ internal struct LambdaHTTPServer {
             logger[metadataKey: "requestId"] = "\(requestId)"
 
             logger.trace("/invoke received invocation, pushing it to the pool and wait for a lambda response")
-            self.invocationPool.push(LocalServerInvocation(requestId: requestId, request: body))
+            let forwardedHeaders = head.headers
```
sebsto: This variable is only used once. Maybe you can just use `head.headers` in the `LocalServerInvocation` initializer below

manojmahapatra:
Hi @sebsto, thanks for taking a look. As suggested I've also added an example that does not depend on the API gateway, and updated relevant docs/README.  Tested locally, working as expected.

sebsto:
Looks good to me. I made a small change to pass the swift format check.

PR #647, opened 2026-02-22
feat: Added support for apple container

manojmahapatra:
## Issue \#

https://github.com/awslabs/swift-aws-lambda-runtime/issues/644

## Description of changes

- Added `--container-cli` option to the archive SwiftPM plugin so users can choose Docker (default) or Apple container
- Implemented Apple container support by switching pull/run commands to `container image pull` and `container run`
- Documented the Apple container path in readme.md
- swift test passed

N/A

## Conventional Commits

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,7 +35,7 @@ Swift AWS Lambda Runtime was designed to make building Lambda functions in Swift
 
 - When developing on macOS, be sure you use macOS 15 (Sequoia) or a more recent macOS version.
 
-- To build and archive your Lambda function, you need to [install docker](https://docs.docker.com/desktop/install/mac-install/).
+- To build and archive your Lambda function, you need to install [docker](https://docs.docker.com/desktop/install/mac-install/) or Apple [container](https://github.com/apple/container).
 
 - To deploy the Lambda function and invoke it, you must have [an AWS account](https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-creating.html) and [install and configure the `aws` command line](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html).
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -137,6 +137,12 @@ The runtime comes with a plugin to compile on Amazon Linux and create a ZIP arch
 swift package archive --allow-network-connections docker
 ```
 
+Apple container (requires disabling the SwiftPM sandbox):
+
+```bash
+swift package --disable-sandbox --allow-network-connections docker archive --container-cli container
```
sebsto: The `--allow-network-connections docker` is not necessary when we use `--disable-sandbox`

manojmahapatra: Thanks for calling this out, but this was intentional. Without `--allow-network-connections docker`, the command will fail in non-interactive mode.

manojmahapatra: interactive version
```
swift package --disable-sandbox archive --container-cli container

Plugin ‘AWSLambdaPackager’ wants permission to allow docker unix domain socket connections.
Stated reason: “This plugin uses Docker to create the AWS Lambda ZIP package.”.
Allow this plugin to allow docker unix domain socket connections? (yes/no)
```
vs non-interactive version
```
swift package --disable-sandbox --allow-network-connections docker archive --container-cli container

building "swift-aws-lambda-runtime" in container
updating "swift:amazonlinux2" image
  Warning! Running debug build. Performance may be degraded.
building "MockServer"
  Warning! Running debug build. Performance may be degraded.
  Building for production...
  [0/2] Write swift-version-24593BA9C3E375BF.txt
  Build of product 'MockServer' complete! (0.82s)
archiving "MockServer"
1 archive created
```

sebsto: According to the SPM team, it sounds like a bug.  `--disable-sandbox` should overrule them all and not trigger the interactive prompt. See https://swift-open-source.slack.com/archives/C04MQHWCQKW/p1771858011429749?thread_ts=1771486919.778029&cid=C04MQHWCQKW

sebsto:
I filed a bug against SPM 
https://github.com/swiftlang/swift-package-manager/issues/9763

sebsto:
Question: should we rename the function `buildInDocker` to something else ? `buildInContainer` ? (with the generic usage of the word container) - but this is confusing because it's also the name of the container CLI.

What do you think?

readme.md
manojmahapatra: Ahh..I see. Now it makes sense, I had been wondering what's going on with disable sandbox.

manojmahapatra:
> Question: should we rename the function `buildInDocker` to something else ? `buildInContainer` ? (with the generic usage of the word container) - but this is confusing because it's also the name of the container CLI.
> 
> What do you think?

That's good point. we can perhaps rename it something akin to `buidInContainerEnvironment` to make it sound like generic container based path? Plus the function is private, I'll be not worried too much.

sebsto:
> Plus the function is private, I'll be not worried too much.

Correct. That's something I can change in the new plugin implementation too

sebsto:
Wow - the SPM fix is waiting to be merged. https://github.com/swiftlang/swift-package-manager/pull/9764
But it won't land until a future Swift release.

manojmahapatra:
Oh wow...that was super quick :)

PR #652, opened 2026-02-25
Correctly propagate the trace ID

### LINKED ISSUE #633: [core] Implement X-Ray Trace ID Propagation via Environment Variable
The Swift AWS Lambda Runtime currently receives the X-Ray trace ID from the Lambda Runtime API via the `Lambda-Runtime-Trace-Id` header and stores it in the `LambdaContext.traceID` property. However, according to the [AWS Lambda Runtime API documentation](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html), the runtime should also set the `_X_AMZN_TRACE_ID` environment variable with this value to enable the X-Ray SDK to automatically trace requests.

> "The runtime should set the _X_AMZN_TRACE_ID with the value of the header. The X-Ray SDK reads this to get the IDs and determine whether to trace the request."

This feature is currently missing from the Swift runtime, preventing seamless integration with X-Ray tracing SDKs.

## Current Behavior

1. The runtime receives the `Lambda-Runtime-Trace-Id` header from the Lambda Runtime API
2. The trace ID is stored in `LambdaContext.traceID` 
3. The `_X_AMZN_TRACE_ID` environment variable is **not** set
4. X-Ray SDKs cannot automatically discover the trace context

## Expected Behavior

1. The runtime receives the `Lambda-Runtime-Trace-Id` header
2. The trace ID is stored in `LambdaContext.traceID`
3. The `_X_AMZN_TRACE_ID` environment variable is set with the trace ID value
4. X-Ray SDKs can automatically read the trace context from the environment variable

### Standard Lambda Functions

For standard Lambda functions, setting the environment variable is straightforward using `setenv()` or similar mechanisms.

### Lambda Managed Instances (LMI)

Lambda Managed Instances introduce complexity because:
- Multiple concurrent invocations may be processed by the same runtime instance
- Environment variables are process-global and shared across all invocations
- Setting `_X_AMZN_TRACE_ID` globally would cause trace ID conflicts between concurrent invocations

### How Other Runtimes Handle This

**Python Runtime:**
- Uses `os.environ['_X_AMZN_TRACE_ID']` to set the trace ID before each handler invocation
- Reference: [bootstrap.py#L434](https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/main/awslambdaric/bootstrap.py#L434)
- Boto3 SDK reads from environment variable: [botocore handlers.py#L110](https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L110)

**Java Runtime:**
- Uses thread-local storage via `LambdaSegmentContext` to avoid environment variable conflicts
- Reference: [AWSLambda.java#L241](https://github.com/aws/aws-lambda-java-libs/blob/5a0f8ad9e9b8ff61ba626e71c4b5de62134e82e4/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambda.java#L241)
- X-Ray SDK uses thread-local context: [LambdaSegmentContext.java#L46](https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java#L46C15-L46C33)

**Node.js Runtime:**
- Sets `process.env._X_AMZN_TRACE_ID` before each invocation
- Reference: [InvokeContext.js#L115](https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/a5ae1c2a92708e81c9df4949c60fd9e1e6e46bed/src/InvokeContext.js#L115)
- AWS SDK middleware injects trace ID: [middleware-recursion-detection index.ts#L41](https://github.com/aws/aws-sdk-js-v3/blob/bfe38d0075ff6e707852b00437fbc3bd4540cdf5/packages/middleware-recursion-detection/src/index.ts#L41)

### Solution: Task-Local Values

Swift's task-local values provide the ideal solution for managing per-invocation trace IDs in Lambda Managed Instances. This approach:
- Isolates trace IDs to individual invocation tasks
- Avoids global environment variable conflicts
- Aligns with Swift's structured concurrency model
- Mirrors Java's thread-local approach but leverages Swift's modern concurrency primitives

### 1. Define Task-Local Trace ID Storage

Create a task-local value to store the X-Ray trace ID for the current invocation:

```swift
// In Sources/AWSLambdaRuntime/LambdaContext.swift or new file

@TaskLocal
public static var xrayTraceID: String?
```

### 2. Set Task-Local Value During Invocation

Modify the runtime to set the task-local value when processing each invocation:

**File: `Sources/AWSLambdaRuntime/Runtime/LambdaRuntime.swift`**

In the invocation handling code, wrap the handler execution with the task-local value:

```swift
await $xrayTraceID.withValue(context.traceID) {
    // Execute handler with trace ID available in task-local storage
    try await handler(event, context)
}
```

### 3. Provide Public API for X-Ray SDK Integration

Add a public accessor that X-Ray SDKs can use to retrieve the current trace ID:

**File: `Sources/AWSLambdaRuntime/LambdaContext.swift`**

```swift
extension LambdaContext {
    /// Returns the X-Ray trace ID for the current invocation.
    /// This value is available via task-local storage and is safe to use in concurrent environments.
    public static var currentTraceID: String? {
        xrayTraceID
    }
}
```

### 4. Maintain Backward Compatibility with Environment Variable

For standard Lambda functions (non-LMI), also set the environment variable to maintain compatibility with existing X-Ray SDK implementations:

**File: `Sources/AWSLambdaRuntime/Runtime/LambdaRuntime.swift`**

```swift
// Set environment variable for backward compatibility
// This is safe in standard Lambda but will be overridden per-invocation in LMI
setenv("_X_AMZN_TRACE_ID", context.traceID, 1)
```

### Files to Modify

1. **`Sources/AWSLambdaRuntime/LambdaContext.swift`**
   - Add `@TaskLocal static var xrayTraceID: String?`
   - Add `public static var currentTraceID: String?` accessor

2. **`Sources/AWSLambdaRuntime/Runtime/LambdaRuntime.swift`**
   - Set task-local value before handler execution
   - Set environment variable for backward compatibility

3. **Documentation**
   - Update documentation to explain X-Ray integration
   - Document the task-local API for X-Ray SDK developers
   - Add example showing how to access trace ID from within handlers

## Testing Requirements

- [ ] Unit tests for task-local trace ID storage and retrieval
- [ ] Integration tests with X-Ray SDK using task-local API
- [ ] Tests for concurrent invocations (LMI scenario) verifying trace ID isolation
- [ ] Verify no trace ID leakage between concurrent invocations
- [ ] Test environment variable setting for backward compatibility
- [ ] Performance tests to ensure minimal overhead from task-local storage

## References

- [AWS Lambda Runtime API Documentation](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html)
- [X-Ray Tracing Header Documentation](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader)
- [Swift Task-Local Values Documentation](https://github.com/apple/swift-evolution/blob/main/proposals/0311-task-locals.md)

### Other Runtime Implementations

- Python Runtime: [bootstrap.py#L434](https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/main/awslambdaric/bootstrap.py#L434)
- Java Runtime: [LambdaSegmentContext.java](https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java)
- Node.js Runtime: [InvokeContext.js#L115](https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/a5ae1c2a92708e81c9df4949c60fd9e1e6e46bed/src/InvokeContext.js#L115)

## Labels

- `enhancement`
- `x-ray`
- `tracing`
- `observability`
- `lambda-managed-instances`

sebsto:
## Issue #
https://github.com/awslabs/swift-aws-lambda-runtime/issues/633

## Description of changes

(read original PR description below) 

Updated from the initial PR — based on review feedback, the trace ID propagation has been reworked to use `ServiceContext` from [swift-service-context](https://github.com/apple/swift-service-context) instead of a Lambda-specific `@TaskLocal`.

The trace ID from AWS wasn't being propagated in a way that's accessible to the full async task tree. Tracing libraries like OpenTelemetry need the trace ID without requiring an explicit `LambdaContext` reference, and the old approach of only setting the `_X_AMZN_TRACE_ID` environment variable doesn't work safely in multi-concurrency mode since env vars are process-global.

This change propagates the trace ID via `ServiceContext.current`, the standard context propagation mechanism in the Swift server ecosystem. The runtime sets `ServiceContext.current?.traceID` before calling the handler, making it automatically available to all code in the handler's async task tree — including downstream libraries that have no dependency on `AWSLambdaRuntime`.

A `ServiceContextKey` (`LambdaTraceIDKey`) is defined with a `public traceID` accessor on `ServiceContext`, so any library depending only on swift-service-context can read the trace ID:

```swift
import ServiceContextModule

if let traceID = ServiceContext.current?.traceID {
    // propagate to outgoing requests
}
```

In single-concurrency mode, the runtime still sets and clears the `_X_AMZN_TRACE_ID` env var for backward compatibility with legacy tooling. In multi-concurrency mode, only `ServiceContext` is used to avoid races between concurrent invocations.

The `runLoop` function accepts an `isSingleConcurrencyMode` flag so it knows which strategy to use. Call sites in `LambdaRuntime` and `LambdaManagedRuntime` pass the appropriate value.

A "Trace ID Propagation" section has been added to the README documenting both access patterns (from a handler via `context.traceID`, and from downstream libraries via `ServiceContext)`.

### New/existing dependencies

Added swift-service-context (1.3.0+) as a new dependency of the `AWSLambdaRuntime` target. This is a lightweight package from Apple that provides the standard `ServiceContext` type used across the Swift server ecosystem (swift-distributed-tracing, swift-service-lifecycle, etc.). It has no transitive dependencies.

### Conventional Commits
feat: propagate trace ID via ServiceContext for ecosystem-wide access

---
Original PR description for history
---

The trace ID from AWS wasn't being propagated in a way that's accessible to the full async task tree. Tracing libraries like OpenTelemetry need the trace ID without requiring an explicit `LambdaContext` reference, and the old approach of only setting the `_X_AMZN_TRACE_ID` environment variable doesn't work safely in multi-concurrency mode since env vars are process-global.

This change adds a `@TaskLocal` static property `LambdaContext.currentTraceID` that the runtime sets before calling the handler. It's automatically available to all code in the handler's async task tree, including child tasks and background work.

In single-concurrency mode, the runtime still sets and clears the `_X_AMZN_TRACE_ID` env var for backward compatibility with legacy tooling. In multi-concurrency mode, only the TaskLocal is used to avoid races between concurrent invocations.

The `runLoop` function now accepts an `isSingleConcurrencyMode` flag so it knows which strategy to use. Call sites in `LambdaRuntime` and `LambdaManagedRuntime` pass the appropriate value.

Tests cover TaskLocal scoping, child task inheritance, concurrent isolation between tasks, env var lifecycle in both modes, and coexistence of the static TaskLocal with the existing instance `traceID` property.

No new dependencies were added to this change.

## Conventional Commits

`feat: propagate trace ID via TaskLocal for async task tree access`

### DIFF
--- a/Sources/AWSLambdaRuntime/Lambda.swift
+++ b/Sources/AWSLambdaRuntime/Lambda.swift
@@ -68,12 +68,13 @@ public enum Lambda {
 
                 logger.trace("Waiting for next invocation")
                 let (invocation, writer) = try await runtimeClient.nextInvocation()
+                let traceId = invocation.metadata.traceID
 
                 // Create a per-request logger with request-specific metadata
                 let requestLogger = loggingConfiguration.makeLogger(
                     label: "Lambda",
                     requestID: invocation.metadata.requestID,
-                    traceID: invocation.metadata.traceID
+                    traceID: traceId
                 )
 
                 // when log level is trace or lower, print the first 6 Mb of the payload
@@ -99,9 +100,9 @@ public enum Lambda {
                 // in the handler's async task tree (e.g. OpenTelemetry instrumentation).
                 // In single-concurrency mode, also set the _X_AMZN_TRACE_ID env var
                 // for backward compatibility with legacy tooling.
-                try await LambdaContext.$currentTraceID.withValue(invocation.metadata.traceID) {
+                try await LambdaContext.$currentTraceID.withValue(traceId) {
                     if isSingleConcurrencyMode {
-                        setenv("_X_AMZN_TRACE_ID", invocation.metadata.traceID, 1)
+                        setenv("_X_AMZN_TRACE_ID", traceId, 1)
                     }
                     defer {
                         if isSingleConcurrencyMode {
@@ -115,7 +116,7 @@ public enum Lambda {
                             responseWriter: writer,
                             context: LambdaContext(
                                 requestID: invocation.metadata.requestID,
-                                traceID: invocation.metadata.traceID,
+                                traceID: traceId,
                                 tenantID: invocation.metadata.tenantID,
                                 invokedFunctionARN: invocation.metadata.invokedFunctionARN,
                                 deadline: LambdaClock.Instant(
--- a/Sources/AWSLambdaRuntime/Lambda.swift
+++ b/Sources/AWSLambdaRuntime/Lambda.swift
@@ -53,28 +53,6 @@ public enum Lambda {
         )
     }
 
-    @available(
-        *,
-        deprecated,
-        message:
-            "This method will be removed in a future major version update. Use runLoop(runtimeClient:handler:loggingConfiguration:logger:isSingleConcurrencyMode:) instead."
-    )
-    @inlinable
-    package static func runLoop<RuntimeClient: LambdaRuntimeClientProtocol, Handler>(
-        runtimeClient: RuntimeClient,
-        handler: Handler,
-        loggingConfiguration: LoggingConfiguration,
-        logger: Logger
-    ) async throws where Handler: StreamingLambdaHandler {
-        try await self.runLoop(
-            runtimeClient: runtimeClient,
-            handler: handler,
-            loggingConfiguration: LoggingConfiguration(logger: logger),
-            logger: logger,
-            isSingleConcurrencyMode: true
-        )
-    }
-
     @inlinable
     package static func runLoop<RuntimeClient: LambdaRuntimeClientProtocol, Handler>(
         runtimeClient: RuntimeClient,
@@ -116,26 +94,44 @@ public enum Lambda {
                     metadata: metadata
                 )
 
-                do {
-                    try await handler.handle(
-                        invocation.event,
-                        responseWriter: writer,
-                        context: LambdaContext(
-                            requestID: invocation.metadata.requestID,
-                            traceID: invocation.metadata.traceID,
-                            tenantID: invocation.metadata.tenantID,
-                            invokedFunctionARN: invocation.metadata.invokedFunctionARN,
-                            deadline: LambdaClock.Instant(
-                                millisecondsSinceEpoch: invocation.metadata.deadlineInMillisSinceEpoch
-                            ),
-                            logger: requestLogger
+                // Wrap handler invocation in a TaskLocal scope so that
+                // LambdaContext.currentTraceID is available to all code
+                // in the handler's async task tree (e.g. OpenTelemetry instrumentation).
+                // In single-concurrency mode, also set the _X_AMZN_TRACE_ID env var
+                // for backward compatibility with legacy tooling.
+                try await LambdaContext.$currentTraceID.withValue(invocation.metadata.traceID) {
+                    if isSingleConcurrencyMode {
+                        setenv("_X_AMZN_TRACE_ID", invocation.metadata.traceID, 1)
+                    }
+                    defer {
+                        if isSingleConcurrencyMode {
+                            unsetenv("_X_AMZN_TRACE_ID")
+                        }
+                    }
+
+                    do {
+                        try await handler.handle(
+                            invocation.event,
+                            responseWriter: writer,
+                            context: LambdaContext(
+                                requestID: invocation.metadata.requestID,
+                                traceID: invocation.metadata.traceID,
+                                tenantID: invocation.metadata.tenantID,
+                                invokedFunctionARN: invocation.metadata.invokedFunctionARN,
+                                deadline: LambdaClock.Instant(
+                                    millisecondsSinceEpoch: invocation.metadata.deadlineInMillisSinceEpoch
+                                ),
+                                logger: requestLogger
+                            )
                         )
-                    )
-                    requestLogger.trace("Handler finished processing invocation")
-                } catch {
-                    requestLogger.trace("Handler failed processing invocation", metadata: ["Handler error": "\(error)"])
-                    try await writer.reportError(error)
-                    continue
+                        requestLogger.trace("Handler finished processing invocation")
+                    } catch {
+                        requestLogger.trace(
+                            "Handler failed processing invocation",
+                            metadata: ["Handler error": "\(error)"]
+                        )
+                        try await writer.reportError(error)
+                    }
                 }
             }
         } catch is CancellationError {
--- a/Sources/AWSLambdaRuntime/Lambda.swift
+++ b/Sources/AWSLambdaRuntime/Lambda.swift
@@ -17,6 +17,7 @@ import Dispatch
 import Logging
 import NIOCore
 import NIOPosix
+import ServiceContextModule
 
 #if os(macOS)
 import Darwin.C
@@ -117,12 +118,14 @@ public enum Lambda {
                     metadata: metadata
                 )
 
-                // Wrap handler invocation in a TaskLocal scope so that
-                // LambdaContext.currentTraceID is available to all code
-                // in the handler's async task tree (e.g. OpenTelemetry instrumentation).
+                // Wrap handler invocation in a ServiceContext scope so that
+                // downstream libraries can access the trace ID via
+                // ServiceContext.current?.traceID without depending on AWSLambdaRuntime.
                 // In single-concurrency mode, also set the _X_AMZN_TRACE_ID env var
                 // for backward compatibility with legacy tooling.
-                try await LambdaContext.$currentTraceID.withValue(traceId) {
+                var serviceContext = ServiceContext.current ?? ServiceContext.topLevel
+                serviceContext.traceID = traceId
+                try await ServiceContext.withValue(serviceContext) {
                     if isSingleConcurrencyMode {
                         setenv("_X_AMZN_TRACE_ID", traceId, 1)
                     }
--- a/Sources/AWSLambdaRuntime/Lambda.swift
+++ b/Sources/AWSLambdaRuntime/Lambda.swift
@@ -53,6 +53,28 @@ public enum Lambda {
         )
     }
 
+    @available(
+        *,
+        deprecated,
+        message:
+            "This method will be removed in a future major version update. Use runLoop(runtimeClient:handler:loggingConfiguration:logger:isSingleConcurrencyMode:) instead."
+    )
+    @inlinable
+    package static func runLoop<RuntimeClient: LambdaRuntimeClientProtocol, Handler>(
+        runtimeClient: RuntimeClient,
+        handler: Handler,
+        loggingConfiguration: LoggingConfiguration,
+        logger: Logger
+    ) async throws where Handler: StreamingLambdaHandler {
+        try await self.runLoop(
+            runtimeClient: runtimeClient,
+            handler: handler,
+            loggingConfiguration: LoggingConfiguration(logger: logger),
+            logger: logger,
+            isSingleConcurrencyMode: true
+        )
+    }
+
     @inlinable
     package static func runLoop<RuntimeClient: LambdaRuntimeClientProtocol, Handler>(
         runtimeClient: RuntimeClient,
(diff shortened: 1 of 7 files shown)


sebsto:
I think API Breakage test is a false positive as the function has `package` access. It's not part of the public API. Nevertheless, I added a `@deprecated` notation to silence this error.

Tests/AWSLambdaRuntimeTests/LambdaTraceIDPropagationTests.swift
```diff
@@ -0,0 +1,195 @@
...
+
+#if os(macOS)
+import Darwin.C
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#endif
+
+@Suite("Trace ID Propagation Tests")
```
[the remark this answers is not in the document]
sebsto: Done https://github.com/awslabs/swift-aws-lambda-runtime/pull/652/commits/5a23719c3891f24bfd887a92e633cc896366a220

Tests/AWSLambdaRuntimeTests/LambdaTraceIDPropagationTests.swift
```diff
@@ -0,0 +1,195 @@
...
+import Testing
+
+@testable import AWSLambdaRuntime
+
+#if os(macOS)
+import Darwin.C
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
```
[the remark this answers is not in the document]
sebsto: We don't support Windows

Sources/AWSLambdaRuntime/LambdaContext.swift
```diff
@@ -222,6 +222,26 @@ public struct LambdaContext: CustomDebugStringConvertible, Sendable {
...
+    /// Returns `nil` when accessed outside of a Lambda invocation scope.
+    ///
+    /// ```swift
+    /// // Inside a Lambda handler or any code called from it:
+    /// if let traceID = LambdaContext.currentTraceID {
+    ///     // Use traceID for downstream propagation
+    /// }
+    /// ```
+    @TaskLocal
+    public static var currentTraceID: String?
```
manojmahapatra: Question: so this is effectively writable by consumers, is this intended? Or `currentTraceID` need to be public read only (getter)?

sebsto: `@TaskLocal` properties don't have a traditional setter. The backing `$currentTraceID` is a let `TaskLocal<String?>` — it's immutable. The only way to change the value is via `$currentTraceID.withValue(_:operation:)`, which scopes the value to the caller's own async task tree for the duration of a closure. It doesn't mutate any shared state and has no effect on other tasks.

So while a consumer could technically call `LambdaContext.$currentTraceID.withValue("x") { ... }`, this only shadows the value within their own task scope — it cannot affect the runtime-set value seen by other code. This is safe by design and consistent with how `TaskLocal` works throughout the Swift ecosystem — for example, swift-service-context declares `@TaskLocal public static var current: ServiceContext?` with the same pattern.

The property must be public because it's referenced from @inlinable code in the runtime loop. The compiler requires that any symbol used inside an @inlinable function be either public or @usableFromInline, and the @TaskLocal macro doesn't support @usableFromInline on its synthesized backing property. Making it package or internal would cause a compilation error.

sebsto: But you make think about something - this solution is scoped to packaged that know about lambda.
I should have use the swift service context to allow downstream dependencies, such as the HTTP Async client or database drivers, to access this value.

readme.md
```diff
@@ -397,6 +397,46 @@ try await runtime.run()
...
+
+    return "OK"
+}
+
+try await runtime.run()
+```
+
+#### Accessing the trace ID from a downstream library
+
+Libraries that don't depend on `AWSLambdaRuntime` can read the trace ID through `ServiceContext`:
```
manojmahapatra, line 423: Suggestion: I think this is not correct. Downstream code still need to import `AWSLambdaRuntime` to read trace ID.

Sources/AWSLambdaRuntime/LambdaContext.swift
manojmahapatra: The new changes are nice and clean IMO. I like it. Also thanks for introducing me to ServiceContextModule, I will read on it.

readme.md
sebsto: Not of they are using Swift Service Context.
I have a doubt about the naming conventions and how we are suppose to inject that X RAY TRACE ID into the context. I asked @ktoso to review and comment

Sources/AWSLambdaRuntime/Lambda.swift
```diff
@@ -116,26 +118,46 @@ public enum Lambda {
...
+                // for backward compatibility with legacy tooling.
+                var serviceContext = ServiceContext.current ?? ServiceContext.topLevel
+                serviceContext.traceID = traceId
+                try await ServiceContext.withValue(serviceContext) {
+                    if isSingleConcurrencyMode {
+                        setenv("_X_AMZN_TRACE_ID", traceId, 1)
+                    }
+                    defer {
+                        if isSingleConcurrencyMode {
+                            unsetenv("_X_AMZN_TRACE_ID")
```
ktoso, line 134: Maybe you want to have a constant for this string somewhere?

Sources/AWSLambdaRuntime/Lambda.swift
```diff
@@ -116,26 +118,46 @@ public enum Lambda {
...
+                        }
+                    }
+
+                    do {
+                        try await handler.handle(
+                            invocation.event,
+                            responseWriter: writer,
+                            context: LambdaContext(
+                                requestID: invocation.metadata.requestID,
+                                traceID: traceId,
```
ktoso, line 144: nitpick: `traceID` for consistency

Sources/AWSLambdaRuntime/ServiceContext+TraceId.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// This value is automatically set by the Lambda runtime before calling the handler
+    /// and is available to all code running within the handler's async task tree.
+    ///
+    /// Downstream libraries can read this without depending on `AWSLambdaRuntime`:
+    /// ```swift
+    /// if let traceID = ServiceContext.current?.traceID {
+    ///     // propagate traceID to outgoing HTTP requests, etc.
+    /// }
+    /// ```
+    public var traceID: String? {
```
ktoso, line 42: It is a bit problematic to declare the key in the lambda runtime, it'd be better if it was declared inside some other library that is "the xray library" and aws lambda runtime would depend on it for the key.

This way other libs which want to use xray specifically could do so without conflicting here...

I think what we may need to do in the short term -- unless we spin out a lib -- would be to call this `awsLambdaXRayTraceID` as long as it is, it won't cause conflicts in the future if there were some "xray library" which would be the right place to declare the proper key and aws lambda runtime would then use it as well 🤔 

WDYT?

Also cc @slashmo for opinions

sebsto: @maxday How runtimes are supporting this in other languages ?

sebsto: @jbelkins should we create a `swift-xray-library` that extends `ServiceConfig` with that key ? Can the AWS SDK use it ?

PR #663, opened 2026-05-09
Add logGroupName and logStreamName to LambdaContext

### LINKED ISSUE #662: Include logStreamName/logGroupName in LambdaContext
### Describe the feature

Other runtimes support getting the log stream & log group names from the context object. As far as I can tell this isn't possible in the Swift runtime at all.

Python has the `log_stream_name` and `log_group_name` properties for example: https://docs.aws.amazon.com/lambda/latest/dg/python-context.html

### Use Case

I'm using lambda functions to handle some jobs & it would be really handy if my admin dashboard could link from the "job status" I'm recording in DynamoDB directly to the CloudWatch log stream for that particular execution. It would be pretty easy to build that if I could get access to the log stream name for the current execution from inside the closure I pass to LambdaRuntime.

### Proposed Solution

Add `logStreamName` and `logGroupName` to LambdaContext, retrieved however it is that these are retrieved in other runtimes.

### Other Information

Workaround: none that I'm aware of.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

sebsto:
## Summary
- Adds `logGroupName` and `logStreamName` properties to `LambdaContext`, read from `AWS_LAMBDA_LOG_GROUP_NAME` and `AWS_LAMBDA_LOG_STREAM_NAME` environment variables
- Values are read once at runtime initialization (not on every access) and stored as properties, consistent with how other static config is handled
- Matches behavior of Python, Node.js, Java, Go, and other AWS Lambda runtimes

Closes #662

## Test plan
- [x] `swift build` compiles cleanly
- [x] `swift test --filter LambdaContextTests` — all 8 tests pass
- [x] Full test suite (140 tests) passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/AWSLambdaRuntime/Docs.docc/Proposals/0001-v2-api.md
+++ b/Sources/AWSLambdaRuntime/Docs.docc/Proposals/0001-v2-api.md
@@ -247,12 +247,6 @@ public struct LambdaContext: Sendable {
     ///
     /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
     public var logger: Logger { get }
-
-    /// The name of the Amazon CloudWatch Logs group for the function.
-    public var logGroupName: String { get }
-
-    /// The name of the Amazon CloudWatch Logs stream for the current invocation of the function.
-    public var logStreamName: String { get }
 }
 ```
 
--- a/Sources/AWSLambdaRuntime/Docs.docc/Proposals/0001-v2-api.md
+++ b/Sources/AWSLambdaRuntime/Docs.docc/Proposals/0001-v2-api.md
@@ -247,6 +247,12 @@ public struct LambdaContext: Sendable {
     ///
     /// - note: The `LogLevel` can be configured using the `LOG_LEVEL` environment variable.
     public var logger: Logger { get }
+
+    /// The name of the Amazon CloudWatch Logs group for the function.
+    public var logGroupName: String { get }
+
+    /// The name of the Amazon CloudWatch Logs stream for the current invocation of the function.
+    public var logStreamName: String { get }
 }
 ```
 
--- a/Sources/AWSLambdaRuntime/Lambda.swift
+++ b/Sources/AWSLambdaRuntime/Lambda.swift
@@ -85,6 +85,9 @@ public enum Lambda {
     ) async throws where Handler: StreamingLambdaHandler {
         var handler = handler
 
+        let logGroupName = Lambda.env("AWS_LAMBDA_LOG_GROUP_NAME") ?? ""
+        let logStreamName = Lambda.env("AWS_LAMBDA_LOG_STREAM_NAME") ?? ""
+
         do {
             while !Task.isCancelled {
 
@@ -128,7 +131,9 @@ public enum Lambda {
                             deadline: LambdaClock.Instant(
                                 millisecondsSinceEpoch: invocation.metadata.deadlineInMillisSinceEpoch
                             ),
-                            logger: requestLogger
+                            logger: requestLogger,
+                            logGroupName: logGroupName,
+                            logStreamName: logStreamName
                         )
                     )
                     requestLogger.trace("Handler finished processing invocation")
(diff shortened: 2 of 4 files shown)


sebsto:
The API breakage is expected because this PR changes the public initializer of `LambdaContext`. This initializer was already modified when we introduced [support for Lambda tenants](https://github.com/awslabs/swift-aws-lambda-runtime/commit/2abe7eb7de5038dfded58d766d62d815d1d39cc2) and the original (non API breaking) initializer was kept around and marked as deprecated.

WFT:
Are these env vars ever actually expected to be missing when running on AWS? If no, I kind of think they should be non-optional, defaulting to the empty string. That way people don’t have to deal with the optional when it doesn’t really matter.

Otherwise looks good

WFT:
And thank you for taking this up so quickly!

0xTim (approved):
_Technically_ this is a breaking change with the new init parameters, but I don't know if this project has an opinion that. NIO definitely considers it a break and adds a new initialiser, Vapor does not consider it a break. Just something to be aware of

sebsto:
@0xTim Thank you for the review.

We broke the 2.0 API when we started adding parameters to the `LambdaContext` initializer (see PR https://github.com/awslabs/swift-aws-lambda-runtime/pull/608) This is where we marked the existing initialize deprecated and introduced a new one. 

This PR adds more arguments to this initializer, but the 2.0 one is still available, marked as deprecated.

I know this is grey area, but I'll go for the non breaking change because the 2.0 initializer is still around.

sebsto:
@WFT my initial implementation used empty strings as you suggested. Then I changed to optional for two reasons:

1. the other optional values in this struct are already using optionals. I don't want to have some properties with a default value and others being optionals
2. I think optional better captures the semantic of "doesn't exist" vs "exist and has an empty value". It's also more idiomatic to write 

```swift 
if let groupName = context.logGroupName {
   ...
}
```

than testing for an empty value:

```swift 
if !context.logGroupName.isEmpty {
 ...
}
```

You're right: the environment variable should always be present. But as Amazon's CTO Werner Vogels always say "Everything fails all the time", the optional seems more robust to handle all possibilities :-)

WFT:
@sebsto  It is more semantically correct to use an optional IF it has a chance of not existing at runtime. But I disagree with your analysis about that a bit:

All the other optional properties in this struct (`clientContext`, `cognitoIdentity`, and `tenantID`) are things that might really not exist in a well-formed lambda invocation - not everyone uses these features. But IMO the log group/stream properties are more akin to the non-optionals (`requestID`, `traceID`, `invokedFunctionARN`, and `deadline`).

Sure, the lambda environment could break & not set the log env vars, but it could also not send `requestID` and that would also be broken.

By comparison, Rust's equivalent does not have these as Options: https://github.com/aws/aws-lambda-rust-runtime/blob/01237499db5ffc365c6e77072009356a4970dfdc/lambda-runtime/src/lib.rs#L54-L64 (included in the context here: https://github.com/aws/aws-lambda-rust-runtime/blob/01237499db5ffc365c6e77072009356a4970dfdc/lambda-runtime/src/types.rs#L87)

(I think I'm understanding this correctly — I don't read a lot of Rust)

WFT:
I should say this is a fairly minor concern. This feature will be incredibly helpful even if shipped as optional! I'm just trying to help push it in what I believe to be an even more helpful direction.

sebsto:
Thank your for sharing your concerns and having this discussion. I think it's important to correctly capture the semantic and to propose a Swift idiomatic way to manage this. Let try a sample code and re-think about this.

sebsto:
I followed up with https://github.com/awslabs/swift-aws-lambda-runtime/pull/664

sebsto:
@WFT you sold it to me by referring to the behaviour on Rust :-) 
Here it is on release 2.9.0
https://github.com/awslabs/swift-aws-lambda-runtime/releases/tag/2.9.0

WFT:
Awesome! Thanks for the release!

PR #668, opened 2026-05-14
[plugin] Display a warning when compiling on Amazon Linux 2 and optin documentation and examples for Amazon Linux 2023

### LINKED ISSUE #666: Use AmazonLinux2023 in the `archive` plugin
### Describe the feature

Official Swift Amazon Linux 2023 are out.

### Use Case

Amazon Linux 2 is very outdated.

### Proposed Solution

Use Amazon Linux 2023 instead of Amazon Linux 2.

### Other Information

Even in a `swift:6.3-amazonlinux2023`, [Penny's deployment CI](https://github.com/vapor/penny-bot/actions/runs/25826283269/workflow) is for some reason building Amazon Linux 2 images in Docker.
Not sure why, but this feature request will resolve it anyway.
I did check `system-release` and it was what it should have been + correct file permissions (world-readable), so not sure how the plugin logic decided to use Amazon Linux 2.

Sample [CI](https://github.com/vapor/penny-bot/actions/runs/25826283269/job/75880128404#logs) output:

```bash
configuration
{
  outputDirectory: ./zips
  products: ["UsersLambda"]
  buildConfiguration: release
  baseDockerImage: swift:amazonlinux2
  disableDockerImageUpdate: false
  containerCLI: docker
}
building "penny-bot"
building "UsersLambda"
archiving "UsersLambda"
/usr/bin/zip --recurse-paths --symlinks UsersLambda.zip bootstrap
Working directory: ./zips/UsersLambda
    adding: bootstrap
   (deflated 67%)
1 archive created
  * UsersLambda at ./zips/UsersLambda/UsersLambda.zip
```

Notice `baseDockerImage: swift:amazonlinux2`

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### LINKED ISSUE #670: Update deployment docs and SAM template to use provided.al2023
Now that the default toolchain is 6.3 and the packager builds against amazonlinux2023 by default, the deployment documentation and SAM templates still reference `provided.al2`.

We should update:
- `Sources/AWSLambdaRuntime/Docs.docc/Deployment.md`
- `readme.md`
- SAM template examples

to use `provided.al2023` as the Lambda runtime.

Follow-up from #668.

sebsto:
This PR has been reworked. Instead of silently switching the default base image based on Swift version, we now:

1. **Keep Amazon Linux 2 as the default** base Docker image for the packager plugin
2. **Add a prominent deprecation warning** when AL2 is used (either via Docker or natively), informing developers that AL2 reaches End of Life on June 30, 2026
3. **Migrate all examples** (READMEs, SAM templates, scripts) to build and deploy on Amazon Linux 2023 (`provided.al2023` runtime + `--base-docker-image swift:amazonlinux2023`)
4. **Update documentation** (readme, quick-setup) with migration notes

The warning includes the `--base-docker-image swift:6.3-amazonlinux2023` flag and reminds developers to use the `provided.al2023` runtime when deploying.

After June 30, 2026, the default will switch to AL2023.

---

<details>
<summary>Original PR description (superseded)</summary>

~~Now that Docker Hub has official Swift images based on Amazon Linux 2023 (starting with 6.3), the packager plugin picks the right base image automatically depending on the Swift version:~~
~~- Swift 6.3 and later: `swift:<version>-amazonlinux2023`~~
~~- Earlier versions: `swift:<version>-amazonlinux2` (unchanged behavior)~~
~~- No version specified (latest): defaults to `amazonlinux2023`~~

~~When only a major version is provided (e.g. `--swift-version 6` without a minor), we conservatively treat it as 6.0 and use Amazon Linux 2, since we can't be sure it's 6.3+.~~
~~Also added a verbose log line showing the resolved Swift version, Amazon Linux version, and final base image to help with debugging.~~
~~The `--base-docker-image` flag still overrides everything as before.~~

</details>

### DIFF
--- a/.github/workflows/scripts/check-archive-plugin.sh
+++ b/.github/workflows/scripts/check-archive-plugin.sh
@@ -29,7 +29,7 @@ ZIP_FILE=${OUTPUT_DIR}/MyLambda/MyLambda.zip
 pushd "Examples/${EXAMPLE}" || exit 1
 
 # package the example (docker and swift toolchain are installed on the GH runner)
-LAMBDA_USE_LOCAL_DEPS=../.. swift package archive --allow-network-connections docker || exit 1
+LAMBDA_USE_LOCAL_DEPS=../.. swift package archive --allow-network-connections docker --base-docker-image swift:6.3-amazonlinux2023 || exit 1
 
 # did the plugin generated a Linux binary?
 [ -f "${OUTPUT_FILE}" ]
--- a/.github/workflows/scripts/check-archive-plugin.sh
+++ b/.github/workflows/scripts/check-archive-plugin.sh
@@ -29,7 +29,7 @@ ZIP_FILE=${OUTPUT_DIR}/MyLambda/MyLambda.zip
 pushd "Examples/${EXAMPLE}" || exit 1
 
 # package the example (docker and swift toolchain are installed on the GH runner)
-LAMBDA_USE_LOCAL_DEPS=../.. swift package archive --allow-network-connections docker --base-docker-image swift:6.3-amazonlinux2023 || exit 1
+LAMBDA_USE_LOCAL_DEPS=../.. swift package archive --allow-network-connections docker --base-docker-image swift:amazonlinux2023 || exit 1
 
 # did the plugin generated a Linux binary?
 [ -f "${OUTPUT_FILE}" ]
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -509,9 +509,13 @@ private struct Configuration: CustomStringConvertible {
             baseDockerImageArgument.first ?? "swift:\(swiftVersion.map { $0 + "-" } ?? "")\(amazonLinuxVersion)"
 
         if verboseArgument {
-            print(
-                "swift version: \(swiftVersion ?? "latest"), amazon linux version: \(amazonLinuxVersion), base docker image: \(self.baseDockerImage)"
-            )
+            if baseDockerImageArgument.isEmpty {
+                print(
+                    "swift version: \(swiftVersion ?? "latest"), amazon linux version: \(amazonLinuxVersion), base docker image: \(self.baseDockerImage)"
+                )
+            } else {
+                print("base docker image (user provided): \(self.baseDockerImage)")
+            }
         }
 
         self.disableDockerImageUpdate = disableDockerImageUpdateArgument
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -509,7 +509,9 @@ private struct Configuration: CustomStringConvertible {
             baseDockerImageArgument.first ?? "swift:\(swiftVersion.map { $0 + "-" } ?? "")\(amazonLinuxVersion)"
 
         if verboseArgument {
-            print("swift version: \(swiftVersion ?? "latest"), amazon linux version: \(amazonLinuxVersion), base docker image: \(self.baseDockerImage)")
+            print(
+                "swift version: \(swiftVersion ?? "latest"), amazon linux version: \(amazonLinuxVersion), base docker image: \(self.baseDockerImage)"
+            )
         }
 
         self.disableDockerImageUpdate = disableDockerImageUpdateArgument
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -339,7 +339,8 @@ struct AWSLambdaPackager: CommandPlugin {
                                           (default is latest)
                                           This parameter cannot be used when --base-docker-image  is specified.
             --base-docker-image <name>    The name of the base docker image to use for the build.
-                                          (default : swift-<version>:amazonlinux2)
+                                          (default: swift:<version>-amazonlinux2023 for Swift >= 6.3,
+                                           swift:<version>-amazonlinux2 for earlier versions)
                                           This parameter cannot be used when --swift-version is specified.
             --disable-docker-image-update Do not attempt to update the docker image
             --container-cli <name>        The container CLI to use (docker or container)
@@ -484,8 +485,32 @@ private struct Configuration: CustomStringConvertible {
         }
 
         let swiftVersion = swiftVersionArgument.first ?? .none  // undefined version will yield the latest docker image
+
+        // Swift 6.3+ uses amazonlinux2023, earlier versions use amazonlinux2
+        let amazonLinuxVersion: String
+        if let version = swiftVersion {
+            let components = version.split(separator: ".").compactMap { Int($0) }
+            if components.count >= 2 {
+                let major = components[0]
+                let minor = components[1]
+                amazonLinuxVersion = (major > 6 || (major == 6 && minor >= 3)) ? "amazonlinux2023" : "amazonlinux2"
+            } else if let major = components.first {
+                // treat "6" (no minor) as possibly 6.0, hence amazonlinux2
+                amazonLinuxVersion = major > 6 ? "amazonlinux2023" : "amazonlinux2"
+            } else {
+                amazonLinuxVersion = "amazonlinux2023"
+            }
+        } else {
+            // no version specified means "latest", which is >= 6.3
+            amazonLinuxVersion = "amazonlinux2023"
+        }
+
         self.baseDockerImage =
-            baseDockerImageArgument.first ?? "swift:\(swiftVersion.map { $0 + "-" } ?? "")amazonlinux2"
+            baseDockerImageArgument.first ?? "swift:\(swiftVersion.map { $0 + "-" } ?? "")\(amazonLinuxVersion)"
+
+        if verboseArgument {
+            print("swift version: \(swiftVersion ?? "latest"), amazon linux version: \(amazonLinuxVersion), base docker image: \(self.baseDockerImage)")
+        }
 
         self.disableDockerImageUpdate = disableDockerImageUpdateArgument
         self.containerCLI = try ContainerCLI.parse(
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -334,7 +334,7 @@ struct AWSLambdaPackager: CommandPlugin {
         print("Amazon Linux 2023 will become the default after June 30, 2026.")
         print("")
         print("To switch now, re-run with:")
-        print("  --base-docker-image swift:6.3-amazonlinux2023")
+        print("  --base-docker-image swift:amazonlinux2023")
         print("")
         print("When using Amazon Linux 2023, you must also update your Lambda")
         print("deployment to use the provided.al2023 runtime.")
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -39,7 +39,10 @@ struct AWSLambdaPackager: CommandPlugin {
         }
 
         // display deprecation warning when building on or for Amazon Linux 2
-        if self.isAmazonLinux(.al2) || configuration.baseDockerImage.hasSuffix("amazonlinux2") {
+        if self.isAmazonLinux(.al2)
+            || (configuration.baseDockerImage.contains("amazonlinux2")
+                && !configuration.baseDockerImage.contains("amazonlinux2023"))
+        {
             self.displayDeprecationWarning()
         }
 
@@ -326,8 +329,10 @@ struct AWSLambdaPackager: CommandPlugin {
 
     private func displayDeprecationWarning() {
         let separator = String(repeating: "=", count: 68)
+        let red = "\u{001b}[38;2;255;66;69m"
+        let reset = "\u{001b}[0m"
         print("")
-        print(separator)
+        print("\(red)\(separator)")
         print("WARNING: Amazon Linux 2 reaches End of Life on June 30, 2026.")
         print("")
         print("You must migrate to Amazon Linux 2023.")
@@ -340,7 +345,8 @@ struct AWSLambdaPackager: CommandPlugin {
         print("deployment to use the provided.al2023 runtime.")
         print("")
         print("For more information: https://aws.amazon.com/amazon-linux-2")
-        print(separator)
+        print("Available images: https://hub.docker.com/_/swift/tags?name=amazonlinux")
+        print("\(separator)\(reset)")
         print("")
     }
 
@@ -376,6 +382,8 @@ struct AWSLambdaPackager: CommandPlugin {
             --base-docker-image <name>    The name of the base docker image to use for the build.
                                           (default: swift:<version>-amazonlinux2)
                                           Note: Amazon Linux 2023 will become the default after June 30, 2026.
+                                          Visit Docker Hub for all available swift tags:
+                                          https://hub.docker.com/_/swift/tags?name=amazonlinux
                                           This parameter cannot be used when --swift-version is specified.
             --disable-docker-image-update Do not attempt to update the docker image
             --container-cli <name>        The container CLI to use (docker or container)
--- a/Plugins/AWSLambdaPackager/Plugin.swift
+++ b/Plugins/AWSLambdaPackager/Plugin.swift
@@ -38,9 +38,14 @@ struct AWSLambdaPackager: CommandPlugin {
             )
         }
 
+        // display deprecation warning when building on or for Amazon Linux 2
+        if self.isAmazonLinux(.al2) || configuration.baseDockerImage.hasSuffix("amazonlinux2") {
+            self.displayDeprecationWarning()
+        }
+
         let builtProducts: [LambdaProduct: URL]
-        if self.isAmazonLinux() {
-            // build directly on the machine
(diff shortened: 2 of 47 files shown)


sebsto:
The swift-nightly-main failure is unrelated to this PR. It is caused by stricter region-isolation diagnostics in the nightly compiler hitting the current pinned version of swift-async-algorithms (MultiProducerSingleConsumerAsyncChannel+Internal.swift). See https://github.com/apple/swift-async-algorithms/issues/418.

sebsto:
Hey @MahdiBM Can you have a look and comment on this change ?

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -484,8 +485,34 @@ private struct Configuration: CustomStringConvertible {
...
+                amazonLinuxVersion = (major > 6 || (major == 6 && minor >= 3)) ? "amazonlinux2023" : "amazonlinux2"
+            } else if let major = components.first {
+                // treat "6" (no minor) as possibly 6.0, hence amazonlinux2
+                amazonLinuxVersion = major > 6 ? "amazonlinux2023" : "amazonlinux2"
+            } else {
+                amazonLinuxVersion = "amazonlinux2023"
+            }
+        } else {
+            // no version specified means "latest", which is >= 6.3
+            amazonLinuxVersion = "amazonlinux2023"
```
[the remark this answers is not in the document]
sebsto: Good point. The deployment docs already need updating for AL2023 support in general (Lambda runtime name, etc.), but that's a separate concern from this PR which only changes the build-time default. I'll address the docs update in a follow-up: #670.

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -484,8 +485,34 @@ private struct Configuration: CustomStringConvertible {
...
+        }
+
         self.baseDockerImage =
-            baseDockerImageArgument.first ?? "swift:\(swiftVersion.map { $0 + "-" } ?? "")amazonlinux2"
+            baseDockerImageArgument.first ?? "swift:\(swiftVersion.map { $0 + "-" } ?? "")\(amazonLinuxVersion)"
+
+        if verboseArgument {
+            print(
+                "swift version: \(swiftVersion ?? "latest"), amazon linux version: \(amazonLinuxVersion), base docker image: \(self.baseDockerImage)"
+            )
```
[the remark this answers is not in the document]
sebsto: Fair point. Fixed in 68446ce: the verbose log now shows the computed AL version only when the default image is used. When `--base-docker-image` is provided, it just prints the user-supplied image without the misleading AL version.

0xTim (approved):

MahdiBM:
Technically speaking, any app that runs on Swift 6.2 and below should also be able to run on Swift 6.3. So checking the Swift version is probably not very helpful. A swift version upgrade is supposed to always work out, although sometimes it can go wrong when the newer compiler comes with some issues or bugs.

I'm only worried maybe some people are relying on the plugin using amazonlinux2 images. If that is a concern (e.g. if something breaks when you upload a amazonlinux2023 executable to a amazonlinux2 aws lambda) then the safest option would be to print a big scary warning by default and mention that they should pass some kind of flag to ensure they'll use amazon linux 2023 since amazon linux 2 is very outdated 🤔.

So yeah. I'm not sure how gracefully AWS handles that. If things will be just fine, then we can take a bit of a risk and just get rid of amazon Linux 2. Otherwise the safer option is a big warning to ask users to manually choose amazonlinux2023 (--base-docker-image swift:amazonlinux2023 ?), IMO.

sebsto:
Good points. Let me address them:

**On the Swift version check:** You're right that a package targeting swift-tools-version 6.1 will compile fine with the Swift 6.3 compiler (source compatibility is forward). The `--swift-version` flag in the plugin is really just about choosing which Docker image to pull from Docker Hub, not about language compatibility. Since we dropped support for Swift 6.0 in this release and the minimum is now 6.1, we could simplify this further and always default to `swift:amazonlinux2023` (latest). The version-based logic is only there as a safety net for people who explicitly pass `--swift-version 6.1` (which pulls `swift:6.1-amazonlinux2`... because there's no `swift:6.1-amazonlinux2023` image on Docker Hub).

**On the real risk (build/deploy mismatch):** The actual concern is not Swift version compatibility but OS-level library compatibility. `--static-swift-stdlib` only statically links the Swift runtime, not system C libraries. If someone uses `aws-sdk-swift` (which dynamically links `libcrypto` via `aws-crt-swift`), a binary built on AL2023 (OpenSSL 3.x / `libcrypto.so.3`) will crash on a `provided.al2` Lambda runtime (which only has `libcrypto.so.1.1`). The reverse is also true. glibc differences (2.26 on AL2 vs 2.34 on AL2023) add another layer.

So the build environment and the Lambda runtime **must** match. This was already the case before this PR (build on AL2 → deploy to `provided.al2`). The change here just updates the default to match the current state of the world: AL2 is EOL (June 2025), `provided.al2023` is the recommended Lambda runtime, and Swift 6.3 images are available for it.

**On the approach:** I think we can simplify: always default to `amazonlinux2023` regardless of `--swift-version`, and print a note like:

```
note: building on amazonlinux2023. Deploy with --runtime provided.al2023.
      Use --base-docker-image swift:amazonlinux2 if you still target provided.al2.
```

This makes the migration explicit without being scary. Users who still need AL2 have a clear escape hatch.

What do you think?

MahdiBM:
So ... I think we should take the safer approach unless we can confirm just updating to al2023 will not cause issues for like more than ~1% of the users.

The actual issue is that users will need to manually update to provided.al2023.
We shouldn't deploy al2023-built images to al2 because there are no ABI gurantees on Linux and shared libraries like glibc. If we have a binary linked against a previous glibc version (or any other system libraries), it might not work properly with a newer glibc version. In theory at least.

So unless someone with deep knowledge/experience can come through and say the risk is very low, we should take the safer route IMO.
Which is something like a very big and eye catching warning about the need to ensure provided.al2023 is used in the lambda definition, and that a flag is passed to the plugin to make it build for al2023. Perhaps +  a link to aws docs. Essentially a big warning that is also helpful.

In the future we can still change the default to 2023 if we ever think it's a safe default.

MahdiBM:
Which then means we'll also have to update the docs and mention that they default al2 is only kept for legacy reasons and users should theoretically always pass that argument to make the plugin use al2023, if the machine they're building on is not already al2023.



Also ... a question. Penny CI builds stuff in an amazonlinux2023 container. Shouldn't the plugin simply just use the environment instead of using the al2 default?
Basically I think there should be a detection mechanism that if the plugin is triggered already in al2023, then just proceed with al2023 and don't default to al2.

sebsto:
Hello @MahdiBM 

The ABI compatibility is not in the discussion because the binaries are statically linked with all Swift libraries.  The risk is about system libraries, such as libgc and the crypto ones.

If I summarize your message, you're suggesting to have an Amazon Linux 2023 compilation as an opt-in flag to avoid breaking builds or deployments.

I kind of agree with that suggestion. 

Let's bring in the wider community for discussion before taking a decision.

sebsto:
@MahdiBM I apologize for the pause working on this. I pushed a set of changes to address your feedback. Instead of silently switching the default base image, the plugin now keeps AL2 as the default and displays a prominent deprecation warning when AL2 is used. All examples have been migrated to build and deploy on AL2023. Please take a look when you get a chance.

sebsto:
Here is the output when compiling on Amazon Linux 2 

```bash
➜  HelloJSON git:(plugin/amazonlinux2023) LAMBDA_USE_LOCAL_DEPS=../..  swift package archive --allow-network-connections docker
====================================================================
WARNING: Amazon Linux 2 reaches End of Life on June 30, 2026.

You must migrate to Amazon Linux 2023.
Amazon Linux 2023 will become the default after June 30, 2026.

To switch now, re-run with:
  --base-docker-image swift:amazonlinux2023

When using Amazon Linux 2023, you must also update your Lambda
deployment to use the provided.al2023 runtime.

For more information: https://aws.amazon.com/amazon-linux-2
====================================================================

building "hellojson" in docker
updating "swift:amazonlinux2" image
  amazonlinux2: Pulling from library/swift
  Digest: sha256:49ad698b70e08c773f378b3616b272dc070cdd5e04935208714064ceab060dc2
Status: Image is up to date for swift:amazonlinux2
  docker.io/library/swift:amazonlinux2
building "HelloJSON"
  [0/1] Planning build
  Building for production...
  [0/3] Write sources
  [1/3] Write swift-version-24593BA9C3E375BF.txt
  [3/4] Compiling AWSLambdaRuntime Context+Foundation.swift
  [4/6] Compiling HelloJSON main.swift
  [4/6] Write Objects.LinkFileList
  /usr/lib/swift_static/linux/libFoundationEssentials.a(Data+Writing.swift.o):_ThreadLocal.swift.o:function $s20FoundationEssentials19createTemporaryFile33_FC9EC52B075D2ACCFF86F1C9F84293BELL2at6inPath6prefix7options7variants5Int32V_SStSS_AA0Q5OrURLOSSAA4DataV14WritingOptionsVSSSgtKFAJ_SStSgSpys4Int8VGSgKXEfU_: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
[5/6] Linking HelloJSON
  Build of product 'HelloJSON' complete! (20.66s)
archiving "HelloJSON"
1 archive created
  * HelloJSON at /Users/stormacq/Documents/code-amzn/swift/lambda/swift-aws-lambda-runtime/Examples/HelloJSON/.build/plugins/AWSLambdaPackager/outputs/AWSLambdaPackager/HelloJSON/HelloJSON.zip
```

sebsto:
and here is the output when compiling on Amazon Linux 2023

```bash
➜  HelloJSON git:(plugin/amazonlinux2023) ✗ LAMBDA_USE_LOCAL_DEPS=../..  swift package archive --allow-network-connections docker --base-docker-image swift:amazonlinux2023
building "hellojson" in docker
updating "swift:amazonlinux2023" image
  amazonlinux2023: Pulling from library/swift
  Digest: sha256:dab6dd1e149263977ca6e833a748c12d8e777c9da6772180413a6db0c694143c
Status: Image is up to date for swift:amazonlinux2023
docker.io/library/swift:amazonlinux2023
building "HelloJSON"
  [0/1] Planning build
  Building for production...
  [0/2] Write swift-version-24593BA9C3E375BF.txt
...
  [29/31] Compiling HelloJSON main.swift
  [29/31] Write Objects.LinkFileList
  /usr/bin/ld: /usr/lib/swift_static/linux/libFoundationEssentials.a(Data+Writing.swift.o): in function `$s20FoundationEssentials19createTemporaryFile33_FC9EC52B075D2ACCFF86F1C9F84293BELL2at6inPath6prefix7options7variants5Int32V_SStSS_AA0Q5OrURLOSSAA4DataV14WritingOptionsVSSSgtKFAJ_SStSgSpys4Int8VGSgKXEfU_':
_ThreadLocal.swift.o:(.text+0x804): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
[30/31] Linking HelloJSON
  Build of product 'HelloJSON' complete! (75.38s)
archiving "HelloJSON"
1 archive created
  * HelloJSON at /Users/stormacq/Documents/code-amzn/swift/lambda/swift-aws-lambda-runtime/Examples/HelloJSON/.build/plugins/AWSLambdaPackager/outputs/AWSLambdaPackager/HelloJSON/HelloJSON.zip
```

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -339,8 +374,8 @@ struct AWSLambdaPackager: CommandPlugin {
                                           (default is latest)
                                           This parameter cannot be used when --base-docker-image  is specified.
             --base-docker-image <name>    The name of the base docker image to use for the build.
-                                          (default: swift:<version>-amazonlinux2023 for Swift >= 6.3,
-                                           swift:<version>-amazonlinux2 for earlier versions)
+                                          (default: swift:<version>-amazonlinux2)
+                                          Note: Amazon Linux 2023 will become the default after June 30, 2026.
```
MahdiBM, line 384: Minor addition so the newer users can find their way better:
```
Visit Docker Hub for an exhaustive list of all possible swift tags: https://hub.docker.com/_/swift/tags?name=amazonlinux
```

.github/workflows/scripts/check-archive-plugin.sh
```diff
@@ -29,7 +29,7 @@ ZIP_FILE=${OUTPUT_DIR}/MyLambda/MyLambda.zip
 pushd "Examples/${EXAMPLE}" || exit 1
 
 # package the example (docker and swift toolchain are installed on the GH runner)
-LAMBDA_USE_LOCAL_DEPS=../.. swift package archive --allow-network-connections docker || exit 1
+LAMBDA_USE_LOCAL_DEPS=../.. swift package archive --allow-network-connections docker --base-docker-image swift:6.3-amazonlinux2023 || exit 1
```
MahdiBM: Outside of this PR's scope and might require further verifications, but I almost always use `set -eu` at the minimum at the start of all bash scripts, and the `e` part of that means that it'll automatically exit whenever any commands fail, which is usually what you'd expect. Then `exit 1` wouldn't be needed.

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -38,9 +38,14 @@ struct AWSLambdaPackager: CommandPlugin {
             )
         }
 
+        // display deprecation warning when building on or for Amazon Linux 2
+        if self.isAmazonLinux(.al2) || configuration.baseDockerImage.hasSuffix("amazonlinux2") {
```
MahdiBM: We have `amazonlinux2-slim` images as well.

```suggestion
        if self.isAmazonLinux(.al2) || configuration.baseDockerImage.contains("amazonlinux2") {
```

Plugins/AWSLambdaPackager/Plugin.swift
```diff
@@ -299,14 +304,44 @@ struct AWSLambdaPackager: CommandPlugin {
...
+        print("To switch now, re-run with:")
+        print("  --base-docker-image swift:6.3-amazonlinux2023")
+        print("")
+        print("When using Amazon Linux 2023, you must also update your Lambda")
+        print("deployment to use the provided.al2023 runtime.")
+        print("")
+        print("For more information: https://aws.amazon.com/amazon-linux-2")
+        print(separator)
+        print("")
     }
```
MahdiBM, line 351: I'd make this colorful, even though there is a chance this is run in CI or ANSI colors are not supported, in which case we don't really lose anything.
If you want that, simply use `print("\u{001b}[38;2;255;66;69m")` at the beginning (corresponds to Apple's light red in HIG) and use `print("\u{001b}[0m")` at the end.

Sources/AWSLambdaRuntime/Docs.docc/quick-setup.md
```diff
@@ -136,6 +136,8 @@ archiving "MyFirstLambdaFunction"
 cp .build/plugins/AWSLambdaPackager/outputs/AWSLambdaPackager/MyFirstLambdaFunction/MyFirstLambdaFunction.zip ~/Desktop
 ```
 
+> Note: The archive command currently defaults to Amazon Linux 2 (`swift:amazonlinux2`) as the build environment. Amazon Linux 2 reaches End of Life on June 30, 2026 and the default will change to Amazon Linux 2023 after that date. To migrate early, re-run the archive command with `--base-docker-image swift:6.3-amazonlinux2023`. When deploying a function built on Amazon Linux 2023, you must use the `provided.al2023` Lambda runtime instead of `provided.al2`.
```
MahdiBM: I'm fine with it but as the maintainer are you also fine with mentioning `6.3`? You need to keep these updated when 6.4 etc... come out. There are a few of these in some files.
If you do always do a one-off update for each swift version anyway, then this isn't really an issue.

MahdiBM (approved):
One minor issue + 3 subjective comments. Feel free to reject / accept / discuss.

Also thank you, no worries, I wasn't blocked 🙂. I noticed I should use the `--base-docker-image` flag and just did that.

Plugins/AWSLambdaPackager/Plugin.swift
sebsto: added https://github.com/awslabs/swift-aws-lambda-runtime/pull/668/commits/de5c13e0c0ff5ffa004b67bc754dc6a80d3cb92f

Plugins/AWSLambdaPackager/Plugin.swift
sebsto: We can't just check for `contains` as it would match `amazonlinux2023` as well :-) Fixed in https://github.com/awslabs/swift-aws-lambda-runtime/pull/668/commits/de5c13e0c0ff5ffa004b67bc754dc6a80d3cb92f

Plugins/AWSLambdaPackager/Plugin.swift
sebsto: done https://github.com/awslabs/swift-aws-lambda-runtime/pull/668/commits/de5c13e0c0ff5ffa004b67bc754dc6a80d3cb92f

Sources/AWSLambdaRuntime/Docs.docc/quick-setup.md
sebsto: done https://github.com/awslabs/swift-aws-lambda-runtime/pull/668/commits/de5c13e0c0ff5ffa004b67bc754dc6a80d3cb92f

PR #693, opened 2026-07-29
[Examples] Update CDK dependency

sebsto:
Update CDK dependencies


sebsto:
CI : failed Format Check is expected - fixed in PR #692

PR #698, opened 2026-09-21
add a public initializer to Lambda clock

### LINKED ISSUE #697: Can we add a public initialiser for `LambdaClock`

sebsto:
fix #697


sebsto:
The two CI errors are due to building with Swift 6.4 in the plugin, not to this change, I will fix them in a separate PR

adam-fowler (approved):

